PR #124, opened 2020-12-30
Implement stopping strategy for Publisher SDK

KacperKluka:
The new stopping strategy makes the `stop()` method work in an asynchronous way. Now it accepts a callback that returns either a success or failure. Whatever the result is,after calling `stop()` no further action should be called on the Publisher. Additionally, we're handling multiple calls to `stop()`, each of the callers will receive an appropriate callback.

In order to stop the publisher we're performing 3 major actions:
- leaving all trackable channels (asynchronous, but we're blocking queue until it's done)
- stopping Mapbox (synchronous)
- stopping Ably (asynchronous)

Calling `stop()` sets a flag `isStopping` which signalizes that the stopping process is ongoing and no events that aren't connected to the stopping procedure should be processed. In order to not lose track of those events we're logging each of the ignored event.
When stopping is finished then the `isStopped` flag is set.
Those flags help us react to `stop()` calls when the stopping procedure is ongoing or finished.

I've spotted a problem (which is fixed in this PR) with Mapbox not stopping the location updates after stopping our SDK. There was always a notification on the device that was saying that the trip is ongoing. This was because we were only removing location listeners, but we weren't stopping the navigation trip. Currently, the whole Mapbox is being cleaned and this fixes that issue.

FYI: I've started this branch on top of the new interface PR in order to be able to use that interface. The new interface is almost done and needs just some minor adjustments that shouldn't affect that PR.

### DIFF
--- a/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
+++ b/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
@@ -184,10 +184,15 @@ class MainActivity : AppCompatActivity() {
     }
 
     private fun stopTracking() {
-        publisher?.stop()
-        publisher = null
-
-        changeNavigationButtonState(false)
+        publisher?.stop {
+            when (it) {
+                is SuccessResult -> {
+                    publisher = null
+                    changeNavigationButtonState(false)
+                }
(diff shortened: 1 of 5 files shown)


KacperKluka:
I'm closing this PR since it's obsolete and irrelevant after landing of https://github.com/ably/ably-asset-tracking-android/pull/203

PR #131, opened 2021-01-11
Add proguard configuration

KacperKluka:
I've prepared a Proguard configuration so the SDKs users will not have to worry about configuring Proguard for our SDK.
I've had to add `@SerializedName` annotations so fields that are parsed to JSON won't get obfuscated. Because of that I also had to add that annotations to the `Resolution` class which I don't like TBH. Maybe we should introduce some other class that will be a container for Resolution that will be then parsed to JSON? :thinking:


QuintinWillison:
This work is likely obsolete.

PR #139, opened 2021-01-14
Stop sending raw location updates from Publisher to Subscriber

KacperKluka:
I've removed all code connected to sending raw location updates from the Publisher to the Subscriber.
I've left the "raw location changed" event in the Publisher, since we're using the raw location as the current location when setting a destination.


kavalerov:
@paddybyers @KacperKluka by looking at this, and also considering that we had a beta release right on the 14th, I think the reason we removed the raw locations was that we didn't have time to think through the API for this. I think now we have this time, and there is a potential use case in mind (debugging), so I am wondering if we can get back to this

PR #150, opened 2021-01-20
Integration test for Publisher and Subscriber interaction

KacperKluka:
I've made a basic test that exercises Subscriber and Publisher SDKs. In the current state the test is randomly failing. It happens either on wrong amount of events received by Subscriber or the `track()` returning failure instead of success.


core-sdk/src/main/java/com/ably/tracking/LocationUpdateModels.kt
```diff
@@ -2,14 +2,27 @@ package com.ably.tracking
 
 import android.location.Location
 
-open class LocationUpdate(val location: Location, val batteryLevel: Float?)
+open class LocationUpdate(val location: Location, val batteryLevel: Float?) {
+    override fun equals(other: Any?): Boolean =
```
QuintinWillison, line 6: Have we documented anywhere that the battery level is ignored when it comes to equality comparison of location updates?

integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
```diff
@@ -0,0 +1,157 @@
...
+        val scope = CoroutineScope(Dispatchers.Default)
+
+        // when
+        val subscriber = createAndStartSubscriber(trackableId)
+
+        subscriber.locations
+            .onEach { receivedLocations.add(it) }
+            .launchIn(scope)
+
+        // TODO replace this sleep with an await on an expectation that the subscriber is ready (connected and subscribed)
```
QuintinWillison, line 63: If there isn't one already, please create an issue for this.

integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
```diff
@@ -0,0 +1,157 @@
...
+                trackResultExpectation.fulfill(Result.success(Unit))
+            } catch (e: Exception) {
+                testLogD("track failed")
+            }
+        }
+
+        // await
+        dataEndedExpectation.await()
+
+        // TODO replace this sleep with an await on an expectation
```
QuintinWillison, line 91: This needs an issue. Plus it needs more detail as to what it will be awaiting.
(or is this code now obsolete, given the previous statement)

QuintinWillison (approved):
A few things to mop up please, @KacperKluka - either by way of issue creation or with a subsequent PR. Thanks.

core-sdk/src/main/java/com/ably/tracking/LocationUpdateModels.kt
KacperKluka: I don't think so and also I think it might have not been intended as this change was present after the most recent rebase of this branch :thinking: I created https://github.com/ably/ably-asset-tracking-android/issues/210 so we can discuss and choose whether we want to compare the battery level too :wink:

integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
KacperKluka: I wasn't the one who added this TODO but I've tried to investigate what it does and decide if it's needed. It turns out that this sleep is necessary for this test to work. My guess is that it waits for the publisher to send the location events and for the subscriber to receive them. That's because the `dataEndedExpectation.await()` is called when the last part of the predefined data is received, which may be before that data is actually sent. I'll create an issue for that :wink:

integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/issues/211

integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/issues/212

PR #190, opened 2021-02-24
Redesign publisher example app to support adding multiple assets

KacperKluka:
According to the requirements I've added two new screens and changed the one that was already present in the app.

The first new screen is the main screen with trackables list. You're able to change settings from it or add a new trackable. When there already are some trackables then you can see them in a list. Clicking on a trackable from a list will display that trackable's details.

The second new screen is the add trackable screen. It's pretty simple, you can only enter the trackable ID that should be tracked. If there are no trackables currently added, then when adding the first trackable the publisher is also started.

The last screen is the trackable details screen. It's almost the same screen that was previously used in the app. The main difference is that you can use the button to remove a trackable that's currently being tracked. If the last trackable is removed then the publisher is stopped.

In order to implement this I've had to merge changes from https://github.com/ably/ably-asset-tracking-android/pull/188 to this PR. The changes were added in commit https://github.com/ably/ably-asset-tracking-android/pull/190/commits/2e7dbf810db4850481f44fde0d028e208ee112a7.


QuintinWillison:
By way of a visual "diff", I've run this up on a Nokia 2.3. All screen shots grabbed via Android Studio's LogCat view and then resized to 50% via the macOS Preview app. I've redacted GPS coordinates (orange rectangles).

On launching the app I see:

device-2021-03-11-163924

When I click the "+" (bottom right):

device-2021-03-11-164052

Entering an identifier for the trackable brings up the keyboard:

device-2021-03-11-164138

I had to dismiss the keyboard which then gave me access to the "add" button, which I pressed to then see:

device-2021-03-11-164218

It took me a moment to realise I was now in a new activity child of the main screen, but when I touched the Android back button I saw:

device-2021-03-11-164248

If I went through the process of adding a second trackable then I saw:

device-2021-03-11-164339

And, indeed, touching either trackable in that list would go to its specific view. I didn't try stopping either of them.

QuintinWillison (approved):
I don't want to block this pull request on this, but a couple of things we could improve:

1. When entering the identifier for a new trackable the Enter key actually puts a new line into the text box (not really sensical), where I would expect it to perform like a done/add button
2. Once add has been touched for a new trackable it might actually be more logical to simply return to the root screen, rather than present the trackable's specific screen - the flow is a bit confusing

PR #199, opened 2021-03-05
Decide which builder fields are required

KacperKluka:
I removed the `LogConfiguration` class as it was never used anywhere.
Then I marked the required and optional fields of both builders.


publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
```diff
@@ -131,47 +130,40 @@ interface Publisher {
      */
     interface Builder {
         /**
-         * Sets the Ably connection configuration.
+         * **REQUIRED** Sets the Ably connection configuration.
```
QuintinWillison, line 133: Just a little bit SHOUTY, but fair enough. 😝

QuintinWillison (approved):
I always love a PR where there's more code removed than added. 😁

PR #204, opened 2021-03-08
Move Ably wrapper to the common module

KacperKluka:
Before we've had separate Ably wrappers for Subscriber and Publisher but their code was mostly the same. Therefore I've merged them into a single Ably wrapper and moved it to the common module.


QuintinWillison (approved):
Slightly difficult diff to navigate but looks sensible on a skim read. Thanks.

PR #208, opened 2021-03-11
Unify the way of parsing enums sent over Ably channels

KacperKluka:
I've created message models for all domain classes that are being sent through Ably channels. After that I was able to specify the format of enums to be in the `SCREAMING_CAMEL_CASE`

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/MessageModels.kt
+++ b/common/src/main/java/com/ably/tracking/common/MessageModels.kt
@@ -46,7 +46,7 @@ enum class AccuracyMessage {
     HIGH,
 
     @SerializedName("MAXIMUM")
-    MAXIMUM
+    MAXIMUM,
 }
 
 data class EnhancedLocationUpdateMessage(
@@ -61,5 +61,5 @@ enum class LocationUpdateTypeMessage {
     PREDICTED,
 
     @SerializedName("ACTUAL")
-    ACTUAL
+    ACTUAL,
 }
--- a/common/src/main/java/com/ably/tracking/common/MessageModels.kt
+++ b/common/src/main/java/com/ably/tracking/common/MessageModels.kt
@@ -1,7 +1,6 @@
 package com.ably.tracking.common
 
 import com.ably.tracking.GeoJsonMessage
-import com.ably.tracking.LocationUpdateType
 import com.ably.tracking.Resolution
 
 object GeoJsonTypes {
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/ably/tracking/common/MessageModels.kt
```diff
@@ -24,9 +24,42 @@ enum class PresenceAction {
...
+    LOW,
+
+    @SerializedName("BALANCED")
+    BALANCED,
+
+    @SerializedName("HIGH")
+    HIGH,
+
+    @SerializedName("MAXIMUM")
+    MAXIMUM
```
QuintinWillison, line 49: Missing comma and end of line.

common/src/main/java/com/ably/tracking/common/MessageModels.kt
```diff
@@ -24,9 +24,42 @@ enum class PresenceAction {
...
-    val type: LocationUpdateType
+    val type: LocationUpdateTypeMessage
 )
+
+enum class LocationUpdateTypeMessage {
+    @SerializedName("PREDICTED")
+    PREDICTED,
+
+    @SerializedName("ACTUAL")
+    ACTUAL
```
QuintinWillison, line 64: Missing comma and end of line.

common/src/main/java/com/ably/tracking/common/MessageModels.kt
```diff
@@ -24,9 +24,42 @@ enum class PresenceAction {
...
+data class PresenceDataMessage(val type: String, val resolution: ResolutionMessage? = null)
+
+data class ResolutionMessage(
+    val accuracy: AccuracyMessage,
+    val desiredInterval: Long,
+    val minimumDisplacement: Double
+)
+
+enum class AccuracyMessage {
+    @SerializedName("MINIMUM")
```
QuintinWillison, line 36: Surprising that these can't be generated from the enum values somehow.

QuintinWillison (changes requested):

common/src/main/java/com/ably/tracking/common/MessageModels.kt
KacperKluka: They can be generated but as it's pointed out in [this article](https://medium.com/key-lessons-from-books/ditching-gsons-field-naming-strategy-4dd217b51f82) we should be careful with using automated field naming policies. 
Firstly, there's no default naming policy that supports SCREAMING SNAKE CASE, so we would need to implement this by ourselves. 
Secondly, we will need to be extra careful where we're using that naming policy because it will change the names of all fields it will parse, so we will need to introduce some hacks so it will only change enums names. 
Overall it seems that it's far simpler and safer to just use that `@SerializedName` annotation which also makes thing very explicit :wink:

common/src/main/java/com/ably/tracking/common/MessageModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/208/commits/9242554c5bdbc6ba7c6e0c23a60809f470debb5a

common/src/main/java/com/ably/tracking/common/MessageModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/208/commits/9242554c5bdbc6ba7c6e0c23a60809f470debb5a

common/src/main/java/com/ably/tracking/common/MessageModels.kt
QuintinWillison: Fair point and thanks for explaining your reasoning. I guess there will always be a tension between "being explicit" and our general desire and tidy programmers to keep things [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). As usual, neither is a panacea. 😁 Let's leave it as is.

QuintinWillison (approved):

PR #220, opened 2021-03-16
Fix emitting trackable state in the publisher SDK

KacperKluka:
Currently, we are attaching the channel state listener after a channel is created. Because of that we don't receive a "Offline->Online" state change because the channel is already online. In order to fix it I'm emitting the current channel state when we're attaching a channel state listener.

Additionally, I've changed the `Online` state to be an `object` instead of a `class`. This fixes states comparing and there are no unnecessary state updates (previously when the current state was `Online` and the new state was `Online` the comparision `newOnline == oldOnline` returned `false`).

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -35,6 +35,7 @@ interface Ably {
 
     /**
      * Adds a listener for the channel state updates.
+     * After adding a listener it will emit the current state of the channel.
      * Should be called only when there's an existing channel for the [trackableId].
      * If a channel for the [trackableId] doesn't exist then nothing happens.
      *
@@ -143,7 +144,12 @@ class DefaultAbly(
     }
 
     override fun subscribeForChannelStateChange(trackableId: String, listener: (ConnectionStateChange) -> Unit) {
-        channels[trackableId]?.on { listener(it.toTracking()) }
+        channels[trackableId]?.let { channel ->
+            channel.state.toTracking().let { currentChannelState ->
+                listener(ConnectionStateChange(currentChannelState, currentChannelState, null))
+            }
+            channel.on { listener(it.toTracking()) }
+        }
     }
 
     override fun connect(
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -15,6 +15,9 @@ import io.ably.lib.realtime.ConnectionState
 import io.ably.lib.types.AblyException
 import io.ably.lib.types.ChannelOptions
 import io.ably.lib.types.ErrorInfo
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -143,7 +144,12 @@ class DefaultAbly(
     }
 
     override fun subscribeForChannelStateChange(trackableId: String, listener: (ConnectionStateChange) -> Unit) {
-        channels[trackableId]?.on { listener(it.toTracking()) }
+        channels[trackableId]?.let { channel ->
+            channel.state.toTracking().let { currentChannelState ->
+                listener(ConnectionStateChange(currentChannelState, currentChannelState, null))
```
QuintinWillison, line 149: It looks like this is calling `listener` synchronously, blocking completion of this method until that listener has returned. Is that intended? As a caller of any method that accepts a callback or other form of listener, I would pretty much always assume that it's going to be called asynchronously at some point after the method returns.

QuintinWillison (changes requested):
Given the amount of commentary in your opening comment on this pull request, it feels to me like some (perhaps most) of that commentary should be included within the code to help with maintainability.

I can see there's a need for:

- commentary inline to the implementation to explain the state synthesise and why you're doing it
- commentary on the interfaces to define the behaviours - in particular if callbacks are also called synchronously, though as I've said already that feels like a bad design pattern

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: Right, haven't thought about that :thinking: I think that we shouldn't call that listener directly as we're doing right now, but at the same time we need to make sure that it's called before any other state change events. The same applies to the `subscribeForPresenceMessages` there we're also calling the listener in a blocking manner. I'll think about some good way in which we could fix it :wink:

QuintinWillison: I'm a little troubled by this statement:

> we need to make sure that it's called before any other state change events

Why do you believe that should be the case?

The event is simply dispatched off for consumption "at some point in the future" on the abstraction resting on the main / UI thread. At the point of execution of this method we may not be in the main context as the user can call our methods from any thread.

KacperKluka: It will be easier to explain using this method as an example:
Here, we have two methods that use the `listener`. One is the `channel.state` which returns the current state of the channel, the other is `channel.on` which keeps returning the state changes of the channel. We want to emit the current channel state as the first call of our `listener`. Therefore, we need to make sure that the `channel.state` result is the first "event" that's being send to the `listener`. Otherwise, in a really bad scenario, we could get a `Online -> Offline` change from the `on` method and then receive `Online` from the `state` and in result end with a bad state.

KacperKluka: I've checked and I might have been prematurely optimizing this stuff :sweat_smile: It looks like we should be safe with only wrapping the actual trouble making `listener()` calls :wink:
https://github.com/ably/ably-asset-tracking-android/pull/220/commits/e2898fd865344fc74dd33209afdf01566ca28d0f

QuintinWillison (approved):

PR #222, opened 2021-03-16
Add token based authentication

KacperKluka:
I've added the possibility to authenticate in Ably with a token. The old way of authentication with secret key is still available.
In order to adjust the code for Java users I've used the same approach that I took with `LocationSource` classes.


kavalerov:
@paddybyers do you mind having a look at this?

publishing-example-java-app/src/test/java/com/ably/tracking/example/javapublisher/PublisherInterfaceUsageExamples.java
```diff
@@ -79,7 +80,7 @@ public void publisherBuilderUsageExample() {
 
     @Test
     public void publisherTokenAuthUsageExample() {
-        ConnectionConfiguration configuration = ConnectionConfigurationToken.create((params) -> "TOKEN", "CLIENT_ID");
+        ConnectionConfiguration configuration = ConnectionConfigurationToken.create((params) -> new TokenRequest(0, "", "", 0, "", "", ""), "CLIENT_ID");
```
paddybyers, line 83: I don't think this is a very useful example, and the code won't work because it's not a valid `TokenRequest`. If we're including an example, shouldn't the token callback generate a valid token?

paddybyers (commented):
Overall this looks right, but:
- I think there need to be tests and examples that actually generate valid token requests;
- I've not reviewed the naming of everything. I'm not sure calling things `ToAbly()` is right, because the asset tracking SDK is still an Ably SDK, it's just not an Ably realtime messaging SDK.

publishing-example-java-app/src/test/java/com/ably/tracking/example/javapublisher/PublisherInterfaceUsageExamples.java
KacperKluka: I agree but in reality that `TokenRequest` will be generated by the user's backend server :thinking: How should we showcase it here? Will it be enough to create a method called `getTokenRequestFromServer()` or something like that and make that method will return this "dummy" token request object?

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
```diff
@@ -74,10 +75,35 @@ fun io.ably.lib.realtime.ConnectionStateListener.ConnectionStateChange.toTrackin
...
+ * Extension converting Ably Realtime auth token params to the equivalent [TokenRequestParameters] API
+ * presented to users of the Ably Asset Tracking SDKs.
+ */
+fun Auth.TokenParams.toTracking(): TokenRequestParameters =
+    TokenRequestParameters(ttl, capability, clientId, timestamp)
+
+/**
+ * Extension converting Asset Tracking SDK [TokenRequest] to the equivalent [Auth.TokenRequest] from Ably.
+ */
+fun TokenRequest.toAbly(): Auth.TokenRequest =
```
QuintinWillison, line 98: In response to [this comment](https://github.com/ably/ably-asset-tracking-android/pull/222#pullrequestreview-629883418), I would suggest renaming `toAbly` to `toAuth`.

publishing-example-java-app/src/test/java/com/ably/tracking/example/javapublisher/PublisherInterfaceUsageExamples.java
QuintinWillison: Perhaps what is not clear here is that this code is only really in place to demonstrate that these APIs are callable from a Java-based codebase... all we were aiming to see from tests in this suite is that they pass the compiler.

The test Paddy speaks of would be elsewhere, and probably pure Kotlin in the first instance.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Represents a set of parameters that are used when requesting an auth token.
+ */
+open class TokenRequestParameters(
+    /**
+     * Time to live of the token.
+     * If set to 0 then the default value will be used.
+     */
+    val ttl: Long,
+    /**
```
QuintinWillison, line 75: There should be a blank line above these comments. It's all too bunched up for both this class and the one below.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,119 @@
...
+     * This method will be called each time a [TokenRequest] needs to be obtained.
+     *
+     * @param requestParameters Parameters of the token request
+     * @return [TokenRequest] used to obtain the token
+     */
+    fun onRequestToken(requestParameters: TokenRequestParameters): TokenRequest
+}
+
+/**
+ * Represents a set of parameters that are used when requesting an auth token.
```
QuintinWillison, line 67: I think this comment could be much clearer.

I'm reading it and finding myself immediately asking, for example, _who_ is requesting an auth token? i.e. who defines these params and who modifies their behaviour based upon them.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,119 @@
...
+     * Client ID associated with the token.
+     */
+    val clientId: String,
+    /**
+     * Timestamp in milliseconds of this request.
+     */
+    val timestamp: Long
+)
+
+class TokenRequest(
```
QuintinWillison, line 89: This class needs a comment. Who creates this thing? What do they pass it to? How will it be used? Is there any additional documentation we can point to?

QuintinWillison (changes requested):

publishing-example-java-app/src/test/java/com/ably/tracking/example/javapublisher/PublisherInterfaceUsageExamples.java
KacperKluka: We've decided that we will show token auth directly in the example apps 
https://github.com/ably/ably-asset-tracking-android/issues/285

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/3ff96860584e53990c8d710c1a27894d197b8053

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/973f248a5aa9caef59a14f5f97d49f4e013a007f

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/8d38156fc51d3cab161d811a6dfb2e582b6da7cb

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: I'm not sure if I've used the correct documentation link :thinking: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/8d38156fc51d3cab161d811a6dfb2e582b6da7cb

QuintinWillison: The link location is ok, but your markup is not recognised by dokka. See staging deployment [here](https://sdk.ably.com/builds/ably/ably-asset-tracking-android/pull/222/dokka/core-sdk/core-sdk/com.ably.tracking/-token-request/index.html). I would just make the word "here" your hyperlink.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    private constructor() : this({ TokenRequest(0, "", "", 0, "", "", "") }, "")
+    private constructor(callback: TokenRequestCallback, clientId: String) : this(
+        { callback.onRequestToken(it) },
+        clientId
+    )
+}
+
+/**
+ * Interface provided for those using the [ConnectionConfigurationToken] from Java code.
```
QuintinWillison, line 54: Sorry, I should have spotted this in my first round of review but should this interface not be defined in [core-sdk-java](https://github.com/ably/ably-asset-tracking-android/tree/main/core-sdk-java)?

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,132 @@
...
+        /**
+         * @param callback [TokenRequestCallback] that will be called each time a [TokenRequest] needs to be obtained.
+         * @param clientId ID of the client
+         */
+        @JvmStatic
+        fun create(callback: TokenRequestCallback, clientId: String) =
+            ConnectionConfigurationToken(callback, clientId)
+    }
+
+    private constructor() : this({ TokenRequest(0, "", "", 0, "", "", "") }, "")
```
QuintinWillison, line 46: I don't understand why this exists. These dummy values are clearly not viable.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,132 @@
...
+    companion object {
+        /**
+         * @param apiKey Ably key string as obtained from the dashboard.
+         * @param clientId ID of the client
+         */
+        @JvmStatic
+        fun create(apiKey: String, clientId: String) = ConnectionConfigurationKey(apiKey, clientId)
+    }
+
+    private constructor() : this("", "")
```
QuintinWillison, line 19: What is this doing? These values are clearly not viable.

QuintinWillison (changes requested):
In this morning's standup we discussed the likely need for a call on this, after I expressed my concern at the number of properties simply cloned 1:1 from Ably by `TokenRequest` given I had the impression it should be more opaque.

It seems this pull request is currently following [approach 2.1 "Ably TokenRequest is created by your servers and passed to clients"](https://ably.com/documentation/core-features/authentication#token-request-process), necessitating quite a lot of API surface area.

Could we be much more opinionated and only offer a JWT based flow? That would mean refactoring to just follow approach [2.3 "Ably JWT is created by your servers and passed to clients"](https://ably.com/documentation/core-features/authentication#ably-jwt-process), with much less API surface area required in the AAT SDKs.

@paddybyers and @kavalerov - are there any reasons we can't do that? The JWT flow looks architecturally simpler to me. Do we know enough about likely use cases and/or Ably's preferred best practice in this area to insist on that flow?

lmars:
@QuintinWillison I saw this in passing:

> Could we be much more opinionated and only offer a JWT based flow?

If you do chose this, you can include a link to this example I recently wrote of doing so using AWS Lambda 🙂 

https://github.com/ably-labs/ably-serverless-auth

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: Unfortunately, if it was defined in the `core-sdk-java` module then we wouldn't be able to access it here in the `ConnectionConfigurationToken` class.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: For some reason, I've thought that I need to add an empty private constructor to forbid creating that object in the Java code. But it turns out that it's not needed at all, thanks for pointing it out :wink: 
https://github.com/ably/ably-asset-tracking-android/pull/222/commits/20bc91e0d59d1edb2ec9cb1e2c8e3e64006e3215

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/20bc91e0d59d1edb2ec9cb1e2c8e3e64006e3215

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/07f971ad8bc8d754c75a7e2790825ec443fb6e3e

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: After our discussion in Slack I've changed the `ConnectionConfiguration` to be an `interface` instead of a `sealed class` in order to move that Java specific code to the `core-sdk-java`. 
Please take a look and let me know if the naming and that whole factory thing is looking alright :wink: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/124b05c26952f3c9a9b01af612c71567d048f8e8

KacperKluka:
As discussed in a call today, I'll add an integration test that will create and use the token auth flow as a part of this PR.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,126 @@
...
+ *  Base interface for connection configurations.
+ */
+interface ConnectionConfiguration {
+    val clientId: String
+}
+
+/**
+ *  Represents a [ConnectionConfiguration] that uses the basic authentication and requires to provide the API key for Ably.
+ */
+interface ConnectionConfigurationKey : ConnectionConfiguration {
```
QuintinWillison, line 13: I would prefer that the base interface is a suffix, not a prefix. I also think that `Key` may not be the right naming at this level - given the presentation in our [Auth and Security docs](https://ably.com/documentation/core-features/authentication).

On the assumption that we move these interfaces to a new package `com.ably.tracking.connection` and rename the super-interface to `AuthenticationConfiguration` then I propose that this interface be renamed to `BasicAuthenticationConfiguration`.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,126 @@
...
+    private class DefaultConfiguration(
+        override val apiKey: String,
+        override val clientId: String
+    ) : ConnectionConfigurationKey
+}
+
+/**
+ *  Represents a [ConnectionConfiguration] that uses the token authentication and requires to provide a callback that will be called each time a new [TokenRequest] is required.
+ */
+interface ConnectionConfigurationToken : ConnectionConfiguration {
```
QuintinWillison, line 38: I would rename this interface to `TokenAuthenticationConfiguration` and make it a generic type, where that type specifies the return type from the callback. When we fully fledge this API out then this will be one of:

- Token Request (see [2.1](https://ably.com/documentation/core-features/authentication#token-request-process))
- Token (see [2.2](https://ably.com/documentation/core-features/authentication#token-process))
- JWT (see [2.3](https://ably.com/documentation/core-features/authentication#ably-jwt-process) and [2.3(a)](https://ably.com/documentation/core-features/authentication#jwt-embed-process))

For now we're only going to offer interfaces for the token request route, as discussed.

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,126 @@
...
+    private class DefaultConfiguration(
+        override val callback: (TokenRequestParameters) -> TokenRequest,
+        override val clientId: String
+    ) : ConnectionConfigurationToken
+}
+
+/**
+ * Represents a set of parameters that are passed to [TokenRequestCallback.onRequestToken] when Ably SDK is requesting an auth token.
+ */
+open class TokenRequestParameters(
```
QuintinWillison, line 63: Looking at the common themes across all diagrams in our [Token Authentication docs](https://ably.com/documentation/core-features/authentication#token-authentication), I wonder whether we could rename this to `LoginCredentials` given all four ways of doing token authentication all have this entity labelled as "custom login credentials". 🤔

QuintinWillison (changes requested):

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,126 @@
...
+        /**
+         * @param apiKey Ably key string as obtained from the dashboard.
+         * @param clientId ID of the client
+         */
+        @JvmSynthetic
+        fun create(apiKey: String, clientId: String): ConnectionConfigurationKey =
+            DefaultConfiguration(apiKey, clientId)
+    }
+
+    private class DefaultConfiguration(
```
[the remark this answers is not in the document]
KacperKluka: After giving it a second thought I think that it shouldn't be an interface and it doesn't have to be one. I've changed the connection configurations classes to interfaces because I thought it would be necessary in order to move the `TokenRequestCallback` to the `core-java-sdk` module. 
But since I've had to introduce that static factory of configurations for Java users we can safely return to the sealed classes which are more appropriate here. That's because we have a defined set of configurations available and we don't want the users to have the ability to implement their own ones :wink: 
https://github.com/ably/ably-asset-tracking-android/pull/222/commits/bb3ddceec1f734f9c45d031ffecaead84a36edb8

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
```diff
@@ -0,0 +1,126 @@
+package com.ably.tracking
+
+/**
+ *  Base interface for connection configurations.
+ */
+interface ConnectionConfiguration {
```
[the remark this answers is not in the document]
KacperKluka: Moved all connection configuration classes to the `com.ably.tracking.connection` package https://github.com/ably/ably-asset-tracking-android/pull/222/commits/dce98de6d285677034320d946541335b73399c8f
Renamed `ConnectionConfiguration` to `AuthenticationConfiguration` https://github.com/ably/ably-asset-tracking-android/pull/222/commits/46ce3469f0417c5f77d7de7693634cc7b3c7942e

I'm wondering if after that renaming we should also modify our API and in both Publisher and Subscriber builder interfaces rename the
```
        fun connection(configuration: AuthenticationConfiguration): Builder
```
to
```
        fun authentication(configuration: AuthenticationConfiguration): Builder
```
WDYT?

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/224026ae0b8e069788c6f2d6dacbdc76590476be

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/251f620ae45167a1e9c039e69728a2e95726eada

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
paddybyers: Login? No, I don't think so

kavalerov: I think we need to keep it the same across Ably SDKs, so if we use one terminology in our platform SDKs, we should use the same here. My understanding is that it means we should keep it as `TokenRequest`

QuintinWillison: I'm specifically talking about the nomenclature for the entity passed to the app developer's auth callback, where there are broadly three "ways" of doing this (see [this comment](https://github.com/ably/ably-asset-tracking-android/pull/222#discussion_r613533504)). I was just keen to challenge our terminology as there's an opportunity to make it clearer with these new SDKs.

What this means, as I understand it, is that we're keeping with:

| "way" | docs | entity passed to the app's auth callback | entity received back from app's auth callback |
| - | - | - | - |
| Token Request | [2.1](https://ably.com/documentation/core-features/authentication#token-request-process) | TokenRequestParameters | Token Request |
| Token | [2.2](https://ably.com/documentation/core-features/authentication#token-process) | Token Request Parameters | Token |
| JWT | [2.3](https://ably.com/documentation/core-features/authentication#ably-jwt-process), [2.3(a)](https://ably.com/documentation/core-features/authentication#jwt-embed-process) | Token Request Parameters | JSON Web Token |

It just feels to me that "token request parameters" isn't necessarily the right naming for all those "ways".

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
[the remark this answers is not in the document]
QuintinWillison: I guess this boils down to the reasoning behind [TO3j](https://docs.ably.io/client-lib-development-guide/features/#TO3j) (being that `ClientOptions` "embeds" `AuthOptions`), meaning that we're naturally leaning towards blending these two as is the case in the underlying client library.

I like the method name `connection`, but perhaps we need an intermediate layer for `ConnectionConfiguration` that has a single property `authentication: AuthenticationConfiguration`?

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
tbedford: Couple of questions. In your table you call the entity passed "Token Request Parameters" but aren't they "token parameters", as for example: `createTokenRequest(TokenParams tokenParams, AuthOptions authOptions, callback(ErrorInfo err, TokenRequest tokenRequest))` or `requestToken(TokenParams tokenParams, AuthOptions authOptions, callback(ErrorInfo err, TokenDetails tokenDetails))`? Also, when you say passed to app's auth callback, could that also be an auth URL?

KacperKluka:
@QuintinWillison I added the test that exercises the token authentication https://github.com/ably/ably-asset-tracking-android/pull/222/commits/effda4361aeba0dca4df7c40be3c3df6bbb17773

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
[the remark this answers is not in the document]
KacperKluka: We could but then the users would have to use it like this:
```
// before
connection(BasicAuthenticationConfiguration(ABLY_KEY, CLIENT_ID))

// after
connection(ConnectionConfiguration(BasicAuthenticationConfiguration(ABLY_KEY, CLIENT_ID)))
```
I mean it's not a big deal but it adds some boilerplate code for the users. On the advantages though if we would ever consider adding some additional options for the Ably SDK then they could just be put in that `ConnectionConfiguration` and won't make any breaking change in the API. :thinking: I'd like to hear what others think about that idea :wink: @paddybyers @kavalerov WDYT?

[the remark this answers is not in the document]
KacperKluka: @QuintinWillison btw I've forgotten to rename the code for java users - I've fixed it in this commit https://github.com/ably/ably-asset-tracking-android/pull/222/commits/418bba3b96e90f6dfc8983b13defc401493cb5aa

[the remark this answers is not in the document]
kavalerov: I don't mind adding a bit of boilerplate in exchange for more powerful extensibility in the future.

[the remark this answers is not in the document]
KacperKluka: `ConnectionConfiguration` added :wink:  https://github.com/ably/ably-asset-tracking-android/pull/222/commits/1b9747eff2f3ed4a2410d6c48ae4ecdb796708ce

core-sdk/src/main/java/com/ably/tracking/ConnectionConfigurationModels.kt
QuintinWillison: I was taking that from the class name I originally commented on and will admit I didn't look back to the Ably client library to double check the naming there. I'll take a look. FYI, @KacperKluka

QuintinWillison: We had a call on this yesterday followed by some refactoring commits by me and some more to come this morning from @KacperKluka, so this has evolved to the point that I am happy to close this conversation.

publishing-example-java-app/src/test/java/com/ably/tracking/example/javapublisher/PublisherInterfaceUsageExamples.java
```diff
@@ -124,7 +124,13 @@ public long getTimestamp() {
...
     @Test
     public void publisherTokenAuthUsageExample() {
-        ConnectionConfiguration configuration = ConnectionConfigurationFactory.createToken((params) -> createDummyTokenRequest(), "CLIENT_ID");
+        // TODO fix: Cannot access kotlin.jvm.functions.Function1
+
+        // encountered when this line is uncommented:
+        // ConnectionConfiguration configuration = new ConnectionConfiguration(Authentication.tokenRequest(tokenParams -> createDummyTokenRequest(), "CLIENT_ID"));
+
+        // possibly solved with:
+        // https://stackoverflow.com/questions/35161913/kotlin-cannot-access-kotlin-jvm-functions-function1-when-calling-kotlin-function/55742956
```
KacperKluka: @QuintinWillison do you intend to leave this commented code for now or was it committed by mistake? :wink:

subscribing-example-java-app/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -107,7 +107,7 @@ public long getTimestamp() {
 
     @Test
     public void subscriberTokenAuthUsageExample() {
-        ConnectionConfiguration configuration = ConnectionConfigurationFactory.createToken((params) -> createDummyTokenRequest(), "CLIENT_ID");
+        // TODO reinstate once Java interop figured out
```
KacperKluka, line 110: We should create an issue for that probably :wink:

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
```diff
@@ -18,7 +18,7 @@ sealed class Authentication(val clientId: String) {
          * @param callback Callback that will be called with [TokenParams] each time a [TokenRequest] needs to be obtained.
          * @param clientId ID of the client
          */
-        @JvmStatic
+        @JvmSynthetic
```
KacperKluka, line 39: We should create an issue for adding token authentication support to Java users

publishing-example-java-app/src/test/java/com/ably/tracking/example/javapublisher/PublisherInterfaceUsageExamples.java
QuintinWillison: Since our conversation this morning you have reminded me that this would not be solved by anything mentioned in that Stackoverflow post.

subscribing-example-java-app/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
QuintinWillison: Yes, I'll do that and will link the issue into this thread shortly.

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
QuintinWillison: Yes, I'll do that. Relates to [your other comment](https://github.com/ably/ably-asset-tracking-android/pull/222#discussion_r616492523).

QuintinWillison (approved):

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
QuintinWillison: Issue created: https://github.com/ably/ably-asset-tracking-android/issues/297

publishing-example-java-app/src/test/java/com/ably/tracking/example/javapublisher/PublisherInterfaceUsageExamples.java
QuintinWillison: I've removed these empty tests now: https://github.com/ably/ably-asset-tracking-android/pull/222/commits/f9c7c91b73543dc281fef9c68ed84044c62751ea

subscribing-example-java-app/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
QuintinWillison: Added [as a comment](https://github.com/ably/ably-asset-tracking-android/issues/297#issuecomment-823553952) to existing issue.

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
QuintinWillison: https://github.com/ably/ably-asset-tracking-android/issues/297

PR #224, opened 2021-03-17
Fix modifying Publisher state from the Ably's disconnect callback

KacperKluka:
We shouldn't modify Publisher state from outside of the events queue "thread". Therefore I've added a new event that's used when removing a trackable from the publisher.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -296,18 +296,29 @@ constructor(
                             }
 
                             // Leave Ably channel.
-                            ably.disconnect(event.trackable.id, state.presenceData) {
-                                state.lastChannelConnectionStateChanges.remove(event.trackable.id)
-                                if (it.isSuccess) {
-                                    event.handler(Result.success(true))
+                            ably.disconnect(event.trackable.id, state.presenceData) { result ->
+                                if (result.isSuccess) {
+                                    request(
+                                        DisconnectSuccessEvent(event.trackable) {
+                                            if (it.isSuccess) {
+                                                event.handler(Result.success(true))
+                                            } else {
+                                                event.handler(Result.failure(it.exceptionOrNull()!!))
+                                            }
+                                        }
+                                    )
                                 } else {
-                                    event.handler(Result.failure(it.exceptionOrNull()!!))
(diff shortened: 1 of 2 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -296,19 +296,30 @@ constructor(
...
 
                             // Leave Ably channel.
-                            ably.disconnect(event.trackable.id, state.presenceData) {
-                                state.lastChannelConnectionStateChanges.remove(event.trackable.id)
-                                if (it.isSuccess) {
-                                    event.handler(Result.success(true))
+                            ably.disconnect(event.trackable.id, state.presenceData) { result ->
+                                if (result.isSuccess) {
+                                    request(
+                                        DisconnectSuccessEvent(event.trackable) {
```
QuintinWillison, line 302: This event name alongside the enclosing verb `request` does not feel right to me. Looking at `PublisherEvents.kt` brought me some clarity but I think there's more work could be done here to make the codebase more understandable for maintainers:

1. Consider whether `request` is the right method name, or perhaps whether it's being overused in this codebase - is it meaningful?
2. Fully document the APIs for all `Event` classes, explaining where they're used for and what their arguments convey

QuintinWillison (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
KacperKluka: ad 1)
I think that the `request` is the correct method name as we've designed it to be used from the `DefaultPublisher` to `request` some action from the `CorePublisher`. I think the real issue is that we mixed events that are sent from `DefaultPublisher` with events that are used internally in the `CorePublisher` :thinking: 
Currently, we have two types of events:
- `Request` - an event with a handler (callback)
- `AdhocEvent` - an event without a handler (callback)

Maybe we should add one more type (or two new types if we'd like to separate internal `AdhocEvent`s too) that will represent an event with a handler that is used only internally in the `CorePublisher` (and `CoreSubscriber` ofc)? I can't think of a good name for it though.
Then we'd have to replace the `request` calls with a new method that will be a private method of `CorePublisher` and enqueue that new internal event.
WDYT @QuintinWillison ?

ad 2)
I've created an issue for that https://github.com/ably/ably-asset-tracking-android/issues/260

Issue #214, opened 2021-03-12
Handle exception thrown when subscribing for presence messages fails in publisher


KacperKluka:
As mentioned [in this comment](https://github.com/ably/ably-asset-tracking-android/pull/193/files#r592935578) we need to decide how to handle the exception thrown when we subscribe for presence messages. This happens after successfully joining the presence of a channel.

Maybe a good solution would be to try to subscribe for the presence messages once more? Maybe we shouldn't call it directly after trying it the first time (so in the catch block) but try to do it after some time delay or after sending the first location update? I'm not quite sure what can be the cause of failure when subscribing for presence messages, but if it's the Ably connection status, then we can try to re-subscribe when the Ably connection changes to the connected state?

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ATS-277) by [Unito](https://www.unito.io/learn-more)

PR #247, opened 2021-03-22
Unify approach of GPS location updates

KacperKluka:
I've changed our logic of starting and stopping location updates to work in the same way it works on iOS.
- The location updates are started when the first trackable is added to the Publisher
- The location updates are stopped when the last trackable is removed from the Publisher

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -163,12 +163,6 @@ constructor(
                     }
                 }
                 when (event) {
-                    is StartEvent -> {
-                        if (!state.isTracking) {
-                            state.isTracking = true
-                            mapbox.startTrip()
-                        }
-                    }
                     is SetDestinationSuccessEvent -> {
                         state.estimatedArrivalTimeInMilliseconds =
                             System.currentTimeMillis() + event.routeDurationInMilliseconds
@@ -272,6 +266,10 @@ constructor(
                         }
                     }
                     is ConnectionForTrackableCreatedEvent -> {
+                        if (!state.isTracking) {
+                            state.isTracking = true
+                            mapbox.startTrip()
+                        }
                         state.trackables.add(event.trackable)
                         scope.launch { _trackables.emit(state.trackables) }
                         resolveResolution(event.trackable, state)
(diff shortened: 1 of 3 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -226,7 +217,24 @@ constructor(
...
-                        createChannelForTrackableIfNotExisits(event.trackable, event.handler, state)
+                        ably.connect(event.trackable.id, state.presenceData) { result ->
+                            try {
+                                result.getOrThrow()
+                                try {
+                                    ably.subscribeForPresenceMessages(event.trackable.id) {
+                                        enqueue(PresenceMessageEvent(event.trackable, it))
+                                    }
+                                } catch (exception: ConnectionException) {
+                                    // TODO - what to do here? should we fail the whole process when subscribing for presence fails? or should it continue?
```
QuintinWillison, line 228: Do we have an explicit issue for this yet?

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -226,7 +217,24 @@ constructor(
...
+                                }
+                                ably.subscribeForChannelStateChange(event.trackable.id) {
+                                    enqueue(ChannelConnectionStateChangeEvent(it, event.trackable.id))
+                                }
+                                request(ConnectionForTrackableCreatedEvent(event.trackable, event.handler))
```
QuintinWillison, line 233: I've mentioned this on a previous PR, but I'm not sure that the `request` verb feels correct given the event name.

QuintinWillison (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
KacperKluka: Yes we do :wink: https://github.com/ably/ably-asset-tracking-android/issues/214

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
KacperKluka: Right, let's discuss it under that [PR comment](https://github.com/ably/ably-asset-tracking-android/pull/224#discussion_r602286727) :wink:

Issue #226, opened 2021-03-18
Fix calling suspend functions from Subscriber events queue


KacperKluka:
On Android, we're calling suspending functions in the events queue when we're emitting the current asset state and it can lead to some problems. We should be calling them in a fire-and-forget manner.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ATS-292) by [Unito](https://www.unito.io/learn-more)

PR #252, opened 2021-03-23
Use Ably and channel state when determining the trackable state in Subscriber SDK

KacperKluka:
Like in the Publisher SDK, in order to decide if a trackable is online, offline or failed we're now using the Ably and channel state together with the presence of the Publisher in the channel.

### DIFF
--- a/subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
+++ b/subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
@@ -1,6 +1,8 @@
 package com.ably.tracking.subscriber
 
 import com.ably.tracking.ConnectionException
+import com.ably.tracking.ConnectionState
+import com.ably.tracking.ConnectionStateChange
 import com.ably.tracking.LocationUpdate
 import com.ably.tracking.Resolution
 import com.ably.tracking.TrackableState
@@ -63,6 +65,7 @@ private class DefaultCoreSubscriber(
                 sequenceEventsQueue(channel)
             }
         }
+        ably.subscribeForAblyStateChange { enqueue(AblyConnectionStateChangeEvent(it)) }
     }
 
     override fun enqueue(event: AdhocEvent) {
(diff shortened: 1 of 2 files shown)


KacperKluka:
Merging this should also solve the issue https://github.com/ably/ably-asset-tracking-android/issues/226

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -153,16 +191,20 @@ private class DefaultCoreSubscriber(
...
+    private fun notifyAssetIsOffline() {
+        scope.launch { _trackableStates.emit(TrackableState.Offline()) }
     }
 
     private inner class State(
         var isStopped: Boolean = false,
+        var isPublisherOnline: Boolean = false,
+        var trackableState: TrackableState = TrackableState.Offline(),
+        var lastConnectionStateChange: ConnectionStateChange = ConnectionStateChange(
+            ConnectionState.OFFLINE, ConnectionState.OFFLINE, null
```
QuintinWillison, line 203: I'm a little wary of this synthesis as it infers that it provides value to the runtime during operation, but the reality is that you only ever access `state` and `errorInformation` on these state change events.

Also, the `last` prefix is pretty much meaningless given the context of a `State` encapsulation... by definition this is a snapshot of state so it's implicitly "current" / "currently known".

QuintinWillison (approved):
Functionally this looks correct so I'll get this PR landed now, but @KacperKluka could you please create an issue to improve the naming and consider scope of state capture.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/issues/267
https://github.com/ably/ably-asset-tracking-android/issues/268

Issue #230, opened 2021-03-18
Decide what should happen with Publisher state after stopping


KacperKluka:
I'm wondering if we should remove the trackables from the Publisher when we stop it. And what about the overall state of the publisher after stopping, should we clear it or leave it as it is?

_Originally posted by @KacperKluka in https://github.com/ably/ably-asset-tracking-android/pull/203#r594965613_

QuintinWillison:
It feels to me that given we've gone to quite a lot of effort to think about a proper stopping strategy, with an asynchronous presentation, that we should follow through on that and formally/properly/fully remove all trackables - awaiting confirmation from the server that this is the case - before we close the Ably connection.

@paddybyers @kavalerov would you both agree?

In terms of the "overall state of the publisher after stopping", we know that a stopped publisher is in a final state. In which case, there's not a high amount of work to do here assuming dynamic runtime resources have been released (i.e. Ably and MapBox).

QuintinWillison:
A corresponding issue will also need creating in the [Cocoa repository](https://github.com/ably/ably-asset-tracking-cocoa) for this.

kavalerov:
Agree, we probably should clear it, yes.

PR #255, opened 2021-03-25
Clear the Publisher state after it is stopped

KacperKluka:
After the successful stopping of the Publisher we're clearing its state.


publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -591,5 +592,21 @@ constructor(
                 this@DefaultCorePublisher.routingProfile = value
                 field = value
             }
+
+        fun clear() {
```
QuintinWillison, line 596: I can see that this serves the purpose we were aiming for here, however this leaves me wondering:

- Would `dispose` not be a more common name for this method?
- Could we enhance the design of this class so that, once disposed of, it will throw a runtime "invalid state" exception if any properties are read?

QuintinWillison (changes requested):
I would have liked to have landed this pull request, as it does offer some improvement, but I don't feel it answers the proposal I put forward in [this comment](https://github.com/ably/ably-asset-tracking-android/issues/230#issuecomment-802586333).

In particular, I was expecting to see network operations started to disconnect the objects in `state.trackable` on the Ably server-side. Or am I missing something and this is already being done elsewhere?

QuintinWillison:
Perhaps removing trackables with consideration to their server-side state will be assisted by the work done for https://github.com/ably/ably-asset-tracking-android/pull/262.

KacperKluka:
In regards to disconnecting the trackables from the server-side, this is already done by calling `ably.close()` a line before we're clearing the state. That `ably.close()` closes all the channels and the whole Ably instance and only after that is successfully done it progresses further with clearing the publisher state.

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
KacperKluka: Renamed that method https://github.com/ably/ably-asset-tracking-android/pull/255/commits/58c0aa71c3dfb543dc0e869b3fa9d26305248922
Prepared a separate PR for that exception throwing because I wasn't sure if you're going to like how it turned out :sweat_smile: https://github.com/ably/ably-asset-tracking-android/pull/264

PR #261, opened 2021-03-26
Change actions order when disconnecting from Ably channel

KacperKluka:
I changed the order of actions in the `disconnect()` method of the Ably wrapper. Now it will clear the channel listeners and remove it from the channels collection only after it successfully leaves its presence.


QuintinWillison:
I've changed the base for this pull request but am yet to push a merge commit to make that work.

QuintinWillison (approved):

PR #263, opened 2021-03-29
Add docs workflow

QuintinWillison:


QuintinWillison:
Of more interest to reviewers than the commits added to the underlying branch is what you see when you press the [View deployment](https://sdk.ably.com/builds/ably/ably-asset-tracking-android/pull/263/dokka) button.

I've not tweaked the Dokka configuration, nor otherwise tried to improve what it generates, as this pull request is solely focussed on creating the workflow that surfaces what we've got at the moment.

kavalerov (approved):
I haven't looked at any files changed, but the docs that were generated look ok. I would prefer to have more "higher level overview" in there, and we can review how we render some parts of it in the future - but I think this is a great first version. 

I have to say though that surprisingly many items have very little docs (i.e. no docs for https://sdk.ably.com/builds/ably/ably-asset-tracking-android/pull/263/dokka/publishing-sdk/publishing-sdk/com.ably.tracking.publisher/-destination/index.html) - but I think this shouldn't stop this PR, as the PR is not about the content but rather about adding the workflow itself. 

On the side note - is the documentation going to be open to search engines for crawling?

QuintinWillison:
Thanks for taking a look, @kavalerov. I would also like to see some higher level information there too. Still early days with Dokka so there may be quick wins to be had there for all I know (e.g. source from readme). 😁 

In respect of:
> surprisingly many items have very little docs

Yes, you're absolutely right. This was one of the reasons I wanted to get this content surfaced so we could start assessing how much more work needs to be done on it to bring it up to a high enough level of quality, especially given our wider docs context.

In terms of search engine crawling you raise a really good point. As this is our staging site I was not intending for these `sdk.ably.com` locations, even the `main` branch environment, to become permanent canonical homes for API reference docs. I'll upload a `robots.txt` file to the root of the S3 bucket to prevent crawling.

As a final note, I'm seeing that the Dokka configuration needs work just to surface everything that's there (e.g. `ConnectionConfiguration` on the [`Publisher.Builder.connection` method's page](https://sdk.ably.com/builds/ably/ably-asset-tracking-android/pull/263/dokka/publishing-sdk/publishing-sdk/com.ably.tracking.publisher/-publisher/-builder/connection.html) is not clickable). I'm going to add a commit to fix that as part of this pull request.

KacperKluka (approved):
Docs look good :+1:

tbedford (approved):
Looks good. Only one minor comment - copyright notice needs to be correct format, and mention Ably.

PR #270, opened 2021-03-31
Update code present in the README

KacperKluka:
I updated the code in the `README.md` file so it matches our current API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The Asset Publishing SDK is used to get the location of the assets that need to 
 Here is an example of how the Asset Publishing SDK can be used:
 
 ```kotlin
-// Prepare Resolution Constraints for the Resolution Policy
+// Prepare Resolution Constraints for an asset that will be used in the Resolution Policy
 val exampleConstraints = DefaultResolutionConstraints(
     DefaultResolutionSet( // this constructor provides one Resolution for all states
         Resolution(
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,17 +82,21 @@ val subscriber = Subscriber.subscribers() // Get an AssetSubscriber
...
     .start() // start listening for updates
 
 // Listen for location updates
 locations
     .onEach { } // provide a function to be called when enhanced location updates are received
     .launchIn(scope) // coroutines scope on which the locations are received
 
+// Listen for asset state changes
+trackableStates
+    .onEach { } // provide a function to be called when the asset changes its state
```
QuintinWillison, line 94: I know this applies to other dummy `onEach` implementations in this file, but I wonder if it might be more illustrative if we made it more explicit what has been emitted by the shared flow - perhaps by implementing this function using a `print` statement, or even being more Android focussed and illustrating with something like `textView.setText(`... ?

QuintinWillison (approved):
Happy to land this as is, but I do think there is room for improvement.

README.md
KacperKluka: We could add named arguments to those lambdas and `print` them as you've suggested :wink: 
```
// Listen for location updates
locations
    .onEach { locationUpdate -> print(locationUpdate) } // provide a function to be called when enhanced location updates are received
    .launchIn(scope) // coroutines scope on which the locations are received

// Listen for asset state changes
trackableStates
    .onEach { trackableState -> print(trackableState) } // provide a function to be called when the asset changes its state
    .launchIn(scope) // coroutines scope on which the statuses are received
```
@kavalerov WDYT?

kavalerov: I like the option with `print` statements. I think textViews will be an overkill.

QuintinWillison: @KacperKluka can you please create an issue so we don't forget to do this, per @kavalerov's suggestion. Thanks.

KacperKluka: Issue created https://github.com/ably/ably-asset-tracking-android/issues/290

PR #273, opened 2021-04-01
Run MapboxNavigation calls in the main thread

KacperKluka:
As it is required by Mapbox we're calling all the methods of the MapboxNavigation on the main thread.
I've had to use the `Dispatchers.Main.immediate` to avoid deadlocks that were caused by using `runBlocking {}` with the regular `Dispatchers.Main`.


owenpearson (approved):
Unfortunately I don't know enough Kotlin to understand much of this but from what I can tell it looks good 👍

QuintinWillison (approved):
@KacperKluka please could you create an issue to add some commentary (probably against the `DefaultMapbox` class) around our methodology of calling the MapBox SDK? It will need a particular focus on [runBlocking](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html), as they say:

> designed to bridge regular blocking code to libraries that are written in suspending style, to be used in `main` functions and in tests

PR #274, opened 2021-04-06
Update Mapbox to version 2.0

KacperKluka:
Luckily, the update didn't cause a lot of troubles. I only had to change the `NavigationOptions.Builder` creation code.

Please note that the v2.0 of the Mapbox is still in the `beta` phase so we should consider if we want to update it right now or wait for it to become stable.


owenpearson (approved):
LGTM

QuintinWillison (approved):

PR #276, opened 2021-04-06
Add a button for centering the camera on the marker on the map

KacperKluka:
I added a button that centers the camera on the marker on the map. The button is displayed only when a trackable is actively tracked.
Screenshot_20210406-150816_Subscriber

### DIFF
--- a/subscribing-example-app/src/main/java/com/ably/tracking/example/subscriber/MainActivity.kt
+++ b/subscribing-example-app/src/main/java/com/ably/tracking/example/subscriber/MainActivity.kt
@@ -2,6 +2,7 @@ package com.ably.tracking.example.subscriber
 
 import android.location.Location
 import android.os.Bundle
+import android.view.View
 import android.widget.Toast
 import androidx.appcompat.app.AppCompatActivity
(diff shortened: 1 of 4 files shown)


owenpearson (approved):
Code changes look fine to me. I'm unsure about the icon used here though. If I saw that icon on a map I would generally assume that it would centre on _my_ location, rather than the location of the thing being tracked. I wonder if we should use a different icon here? @kavalerov do you have any thoughts on this?

kavalerov:
I just added a comment on the issue - lets just centre on the marker all the time, and remove button - this should showcase locating the asset, so user doesn't really need to be able to scroll around the map.

kavalerov:
@KacperKluka this PR will later be updated with autocentering, right?

KacperKluka:
@kavalerov yes, that's my plan but I have other priorities right now so don't know when I'll work on this. So we can either land it as it is or remove this PR for now and I'll add that autocentering in the future :wink:

KacperKluka:
I'm closing that because we've implemented a better solution with https://github.com/ably/ably-asset-tracking-android/pull/289

Issue #209, opened 2021-03-11
Improve adding trackable flow in the publisher example app


KacperKluka:
As [commented on here](https://github.com/ably/ably-asset-tracking-android/pull/190#pullrequestreview-609994820):

1. When entering the identifier for a new trackable the Enter key actually puts a new line into the text box (not really sensical), where I would expect it to perform like a done/add button
2. Once add has been touched for a new trackable it might actually be more logical to simply return to the root screen, rather than present the trackable's specific screen - the flow is a bit confusing

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ATS-272) by [Unito](https://www.unito.io/learn-more)

KacperKluka:
I agree with point 1. :+1: 
@kavalerov you've designed the flow so I'd like to hear your opinion on point 2. :wink:

PR #278, opened 2021-04-06
Allow to add trackable from the trackable id input keyboard

KacperKluka:
As suggested in the https://github.com/ably/ably-asset-tracking-android/issues/209 I replaced the "Enter" key with a "done" key. If a user presses that "done" key then it acts just as if he has clicked on the "Add trackable" button.


kavalerov:
@KacperKluka can you please add the same? Feel free to do it in separate PR if thats easier.

KacperKluka:
@kavalerov I've created https://github.com/ably/ably-asset-tracking-android/issues/288 for the subscriber example app :wink:

QuintinWillison (approved):

PR #279, opened 2021-04-08
Setting resolution for publisher and trackables in the publisher example app

KacperKluka:
I added fields for setting the default publisher resolution in the settings screen.
Screenshot from 2021-04-08 15-24-23

I added fields for setting the trackable resolution in the add trackable screen. The default values for those fields are set from the default resolution from the settings screen.
Screenshot from 2021-04-08 15-24-36

For some reason I kept getting a lint issue about not specifying the locale. I wasn't able to resolve it even if I was specifying a locale like `Locale.ENGLISH`. Therefore, I've added that lint rule to the `informational` so its still logged but won't fail the build.


kavalerov (approved):
So, I have only tested on 3 devices but worked well on them. 

I think this is good to go. These screens will change once we have new designs from the design team, but this PR is good to merge

PR #281, opened 2021-04-09
Add optional Firebase Crashlytics integration to the publisher example app

KacperKluka:
I integrated the Firebase Crashlytics into the project. In order to make it work we'll need a `google-services.json` file placed in the `publishing-example-app` directory. 

It is optional to provide this file, if the file is provided then Crashlytics are enabled. If the file is not present then we're not using the Crashlytics.

@QuintinWillison the `google-services.json` isn't required to build the project so I'm wondering if it's worth putting that file in the secrets of this repo, WDYT?


kavalerov (approved):
Looks good to me!

QuintinWillison (approved):

PR #282, opened 2021-04-12
Create a test-common module for shared unit tests code

KacperKluka:
Added a module that will keep shared code (like stubs, mocks, utils, etc.) used in unit tests.


build.gradle
```diff
@@ -206,6 +206,7 @@ subprojects {
             }
 
             testImplementation 'junit:junit:4.13.1'
+            testImplementation project(':test-common')
```
QuintinWillison, line 209: I would need to dig deeper locally but my instance reaction to this is to wonder whether it would create a circular dependency with `core-sdk` and `common`, given they're both included in `test-common/build.gradle`. 🤔

test-common/src/main/java/com/ably/tracking/test/common/Mocks.kt
```diff
@@ -0,0 +1,5 @@
+package com.ably.tracking.test.common
+
+import android.location.Location
+
+fun anyLocation() = Location("fused")
```
QuintinWillison, line 5: Given this is common functionality which could end up being used by many tests, it definitely deserves some API commentary. The use of static `"fused"` immediately has me scratching my head, for example, so that may need explaining.

QuintinWillison (changes requested):
I'm not entirely sure where this pull request is going, given that the function it exposes is not used anywhere. It's purely academic, so in its own right I'm not really convinced it moves the codebase forward in any meaningful way. If there were some test support methods moved from elsewhere into this new module as part of this pull request then I think it would make more sense.

test-common/src/main/java/com/ably/tracking/test/common/Mocks.kt
KacperKluka: `"fused"` location type is the type that we're using across our app when we're creating a location :wink: Actually, this is a bit wrong as the `Location` is an Android type that can't be that easily mocked in tests. I've changed it to use the `mockk` library and create a proper mock. https://github.com/ably/ably-asset-tracking-android/commit/47d8f0379e63ac1ce0fd85b5150ad1e7a6339a28#diff-e962a3b4c6e646736850513a9ed73eb4fbdd627102982924570ab8e0a9b109f4

KacperKluka:
> I'm not entirely sure where this pull request is going, given that the function it exposes is not used anywhere.

Right, this PR was meant to prepare the ground for adding more unit tests (i.e. resolution tests) that's why it's quite simple and plain. However, I've moved all test code that creates `Location` objects to that module in order to make it a bit more useful :wink: https://github.com/ably/ably-asset-tracking-android/pull/282/commits/47d8f0379e63ac1ce0fd85b5150ad1e7a6339a28

build.gradle
KacperKluka: I've quickly looked and found [this comment](https://discuss.gradle.org/t/gradle-support-for-cyclic-dependencies/14355/2) and I think that we have a similar scenario and we should be fine :wink: I don't think that gradle would build a project if it had some circular dependencies, I'd expect it to throw an error in that case :thinking:

KacperKluka:
@QuintinWillison the emulate workflow is failing because it turns out that module `android-test-common` has the same package name as the new `test-common` :sweat_smile: The package name is `com.ably.tracking.test.common`. I think that we should change it in the `android-test-common` to be something like `com.ably.tracking.test.android.common`, WDYT? Would it be OK to do it as a part of this PR?

QuintinWillison:
@KacperKluka Given the name clash has been introduced by this pull request then it would definitely be the right thing to do to fix it in this pull request. Thanks. Re [this comment](https://github.com/ably/ably-asset-tracking-android/pull/282#issuecomment-825488767).

QuintinWillison (approved):

PR #284, opened 2021-04-13
Only run deployment fail step on docs workflow if needed

QuintinWillison:
Something I noticed in [the equivalent Cocoa workflow](https://github.com/ably/ably-asset-tracking-cocoa/pull/130).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -126,7 +126,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Set Deployment Status to Failure
-        if: ${{ steps.create_deployment.outcome == 'success' && steps.succeed_deployment.outcome == 'skipped' && failure() }}
+        if: ${{ steps.create_deployment.outcome == 'success' && steps.succeed_deployment.outcome != 'failure' && failure() }}
         uses: octokit/request-action@v2.x
         with:
           route: POST /repos/:repository/deployments/:deployment/statuses
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -111,6 +111,7 @@ jobs:
           aws s3 sync --acl public-read build/dokka/htmlMultiModule "s3://sdk.ably.com/builds/$GITHUB_REPOSITORY/$DEPLOYMENT_REMOTE_PATH"
 
       - name: Set Deployment Status to Success
+        id: succeed_deployment
         uses: octokit/request-action@v2.x
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -124,9 +125,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      # If the previous step (set deployment status to success) FAILS then this step will ALSO get run.
       - name: Set Deployment Status to Failure
-        if: ${{ failure() }}
+        if: ${{ steps.create_deployment.outcome == 'success' && steps.succeed_deployment.outcome == 'skipped' && failure() }}
```
KacperKluka, line 129: I'm not that familiar with GitHub actions so those questions may be stupidly obvious :sweat_smile: but:
1. Why are you checking if `create_deployment` was successful? Shouldn't we always set the status to failure if something goes wrong anywhere?
2. Why is the `skipped` for `succeed_deployment` treated as its failed state?

KacperKluka (commented):
In terms od the code it look good :+1: But I don't understand a few things :sweat_smile:

.github/workflows/docs.yml
QuintinWillison: They're not stupid questions at all. Thanks for asking!

I added more explanation in [this commit](https://github.com/ably/ably-asset-tracking-cocoa/pull/130/commits/3956f67f3b60d3c59dbfa709a1e38993227abf25), so this repository is just reaping benefit from that improvement.

I am checking if `create_deployment` was successful because it is possible for one of the workflow steps prior to `create_deployment` to fail. When that happens then `failure()` is `true` which means that we then try to set deployment status to failed for a deployment that never even started/created because its step was skipped.

In terms of checking for `skipped` on `succeed_deployment`, you may have spotted a mistake... I think I probably meant to check that it wasn't `failure` (see [`steps` context](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#steps-context)) as what I wanted to guard against is the deployment being marked as a failure if we had failed the step to mark it as a success!

I'll push a commit to address that.

QuintinWillison: Fixed in https://github.com/ably/ably-asset-tracking-android/pull/284/commits/9a5d7b9a83adae4366f9f8b83a8a8e6c725543d2.

KacperKluka: Assuming that each job without a custom `if` will just be skipped if some error occurs in the actions chain, I think it will work as desired :grin: Thank you for your explanation :bow:

KacperKluka (approved):
LGTM :+1:

PR #292, opened 2021-04-16
Add skipped location updates to the LocationUpdate

KacperKluka:
I added the skipped location updates to the `LocationUpdate` class. Those locations are the ones that aren't sent because of the trackable's resolution. After successfully sending a location update the skipped location updates for the trackable are cleared.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/MessageMappers.kt
+++ b/common/src/main/java/com/ably/tracking/common/MessageMappers.kt
@@ -44,6 +44,7 @@ fun EnhancedLocationUpdate.toJson(gson: Gson): String =
         EnhancedLocationUpdateMessage(
             location.toGeoJson(),
             batteryLevel,
+            skippedLocations.map { it.toGeoJson() },
             intermediateLocations.map { it.toGeoJson() },
             type.toMessage()
         )
@@ -55,6 +56,7 @@ fun Message.getEnhancedLocationUpdate(gson: Gson): EnhancedLocationUpdate =
             EnhancedLocationUpdate(
                 message.location.toLocation(),
                 message.batteryLevel,
+                message.skippedLocations.map { it.toLocation() },
                 message.intermediateLocations.map { it.toLocation() },
                 message.type.toTracking()
             )
--- a/common/src/main/java/com/ably/tracking/common/MessageModels.kt
+++ b/common/src/main/java/com/ably/tracking/common/MessageModels.kt
@@ -52,6 +52,7 @@ enum class AccuracyMessage {
 data class EnhancedLocationUpdateMessage(
     val location: GeoJsonMessage,
     val batteryLevel: Float?,
+    val skippedLocations: List<GeoJsonMessage>,
     val intermediateLocations: List<GeoJsonMessage>,
     val type: LocationUpdateTypeMessage
 )
--- a/core-sdk/src/main/java/com/ably/tracking/LocationUpdateModels.kt
+++ b/core-sdk/src/main/java/com/ably/tracking/LocationUpdateModels.kt
@@ -2,10 +2,10 @@ package com.ably.tracking
 
 import android.location.Location
 
-open class LocationUpdate(val location: Location, val batteryLevel: Float?) {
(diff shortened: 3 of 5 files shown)


kavalerov:
I think one thing is missing from this PR at the moment - it is the ability to configure wether these "skipped" location updates are sent. I think depending on the use case customers may want or not want to include these skipped events in the updates (i.e. when subscriber doesn't visualise them on the map). 

I think it is worth adding this as a configuration option, and I can think about 3 places to put it in:

1. As a parameter of Publisher Builder itself (as it is applied across all the trackables)
2. As a parameter of the ResolutionPolicy, as that's the part of the system this configuration most closely operates with, or
3. As a parameter of the ConnectionConfiguration, because that's where most of the impact of this option will manifest. 

I think the correct option is 2, as by its nature ResolutionPolicy's responsibility is "to identify the frequency of updates sent over network", and I think adjusting this responsibility to be "to identify frequency _and content_ of updates sent over network" is a small and logical change.  

@QuintinWillison @KacperKluka @paddybyers  WDYT?

KacperKluka:
@kavalerov I'd prefer to use option 1.

The `ResolutionPolicy` purpose is to calculate resolutions based on a few external factors. I think it shouldn't care what then happens with that resolutions and if they cause any locations to be skipped. I mean that probably it shouldn't tell the `CorePublisher` what should it do with location updates as it's not its responsibility :wink:

I think we also have option 4:
Add the `sendSkippedLocations` flag to the `Trackable` class and allow the users to choose which trackables should have skipped locations sent. We could make this an optional field with the default value set to `true`.
I'm not sure if that granular control over skipped locations is useful at all, though. :wink:

kavalerov:
> Add the sendSkippedLocations flag to the Trackable class and allow the users to choose which trackables should have skipped locations sent. We could make this an optional field with the default value set to true.
I'm not sure if that granular control over skipped locations is useful at all, though. 😉

I definitely can see where granular control may be useful - i.e. in the case when user wants to track driver and actual deliveries differently. However, I am also thinking about ways to not build too much fine grain detail until we can confirm the requirement. 


> The ResolutionPolicy purpose is to calculate resolutions based on a few external factors. I think it shouldn't care what then happens with that resolutions and if they cause any locations to be skipped. I mean that probably it shouldn't tell the CorePublisher what should it do with location updates as it's not its responsibility 😉

I see your point. However, I am still thinking about semantic point of view, that resolution of networking, and actual resolution of updates can mean different things. I.e. when you have resolution of 10 minutes for sending updates, but also you send all intermediate locations with 10s precision - you both have realtime resolution of 10 minutes, and historical location resolution of 10s. 

Also, the other part of the argument is the following: shouldn't subscriber be able to request wether it wants to receive intermediate locations? The same logic applies as with resolution - if no-one is looking at the map, there is no need to get fine-grain intermediate locations, and this allows to save on bandwidth by reducing the message size. And vice versa - subscriber may request to get intermediate locations to be able to visualise the asset better on the map.

KacperKluka:
I think we have two options here:

1. The simple one
**Goal:**
We just want to enable or disable sending skipped location updates
**What should be done:**
We could simply add a flag to either the publisher builder or the `ResolutionPolicy` interface and implement this in the `DefaultResolutionPolicy`. This flag will enable or disable collecting and sending the skipped location updates

2. The complex one
**Goal:**
Add much complex logic for choosing if we want to collect and send the skipped locations (i.e. if no subscribers are watching). Allow the subscribers to request sending those skipped locations. 
**What should be done:**
I'd say that in this case, we need to keep track of this preference for each trackable separately. Then since it's so closely related to the `Resolution` perhaps this is the correct place to put it? We could add a flag `sendSkippedLocations` to the `Resolution` that would just enable or disable skipped locations. In the future, we might extend this by allowing to provide minimumDisplacement and desiredInterval for those skipped locations (not sure if that's to any use though).

@kavalerov @QuintinWillison @paddybyers  WDYT?

kavalerov:
Notes from the call we just had: 
- for now we are going to send all intermediate updates by default => this PR is good to go as is (granted it passes review :) )
- I will create a follow up issue shortly to implement the option 2 from the above, to add flag on the resolution level.

QuintinWillison (approved):

PR #298, opened 2021-04-21
Second Beta Release - 1.0.0-beta.2

QuintinWillison:
As per usual, I will push the tag after this lands on `main`.

Then, in terms of [releases](https://github.com/ably/ably-asset-tracking-android/releases), when I come to create it this will continue to be tagged as a "Pre-release" (as were the previous two previews and the previous beta also).


paddybyers (approved):
LGTM, thanks

KacperKluka (approved):
LGTM :+1:

PR #302, opened 2021-04-26
Annotate DTO classes

KacperKluka:
Annotated DTO classes with a `@Dto` annotation.

There's a problem with the `GeoJsonMessage` class because it is technically a DTO class but it's exposed in our API in `LocationHistoryData`. The `@Dto` annotation is placed in the `common` module and is not accessible from the `core-sdk` module in which that `GeoJsonMessage` is placed. The easiest solution would be to just move that `@Dto` annotation to the `core-sdk` module but it will make it visible for SDK users. @QuintinWillison WDYT? What should we do with it?


QuintinWillison (changes requested):

common/src/main/java/com/ably/tracking/common/Dto.kt
```diff
@@ -0,0 +1,4 @@
+package com.ably.tracking.common
+
+@Target(AnnotationTarget.CLASS)
+annotation class Dto
```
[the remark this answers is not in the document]
KacperKluka: Given your commentary (i.e. the part about exported files), I've moved that annotation to the core module https://github.com/ably/ably-asset-tracking-android/pull/302/commits/917f70c563140555616b04bd819c4906c179b073
Because of above, I was able to annotate GeoJSON classes with that annotation https://github.com/ably/ably-asset-tracking-android/pull/302/commits/8b3b1f3467541d9bc08ec70990a999bedc758bbe
Then I've renamed the annotation and added commentary https://github.com/ably/ably-asset-tracking-android/pull/302/commits/aa82010d1ba705e838f8c3155ee6f475700d5129

core-sdk/src/main/java/com/ably/tracking/Shared.kt
```diff
@@ -0,0 +1,10 @@
+package com.ably.tracking
```
QuintinWillison, line 1: Now that you've moved it, the name `Shared` is now in the "global" namespace (as it were) which looks odd.

I think a more appropriate package name would be `com.ably.tracking.annotations`.

QuintinWillison (changes requested):
Thanks. Just one final tweak, please. 😁

core-sdk/src/main/java/com/ably/tracking/Shared.kt
KacperKluka: Good point :smile: https://github.com/ably/ably-asset-tracking-android/pull/302/commits/34c20a6e18b9f4ceb376881a6ee1a47eec96d253

QuintinWillison (approved):

Issue #201, opened 2021-03-05
Exceptions are not routed meaningfully


QuintinWillison:
Prompted by [this comment](https://github.com/ably/ably-asset-tracking-android/pull/179#pullrequestreview-599844949) (deferment).

A useful starting point:

```
ably-asset-tracking-android % git grep 'Timber\.e([a-z]'
publishing-example-app/src/main/java/com/ably/tracking/example/publisher/S3Helper.kt:                { error -> Timber.e(error, "Error when listing S3 files") }
publishing-example-app/src/main/java/com/ably/tracking/example/publisher/S3Helper.kt:                { error -> Timber.e(error, "Error when downloading S3 file") }
publishing-example-app/src/main/java/com/ably/tracking/example/publisher/S3Helper.kt:                    { error -> Timber.e(error, "Error when uploading S3 file") }
publishing-sdk/src/main/java/com/ably/tracking/publisher/Ably.kt:            Timber.e(exception, "sendEnhancedLocationMessage: publishing: $locationUpdateJson")
publishing-sdk/src/main/java/com/ably/tracking/publisher/Ably.kt:                Timber.e(exception, "subscribeForPresenceMessages: Cannot subscribe to presence messages")
publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt:                    Timber.e(throwable, "Failed call to requestRoutes.")
publishing-sdk/src/main/java/com/ably/tracking/publisher/locationengine/FusedAndroidLocationEngine.kt:            Timber.e(exception)
publishing-sdk/src/main/java/com/ably/tracking/publisher/locationengine/FusedAndroidLocationEngine.kt:                Timber.e(exception)
publishing-sdk/src/main/java/com/ably/tracking/publisher/locationengine/FusedAndroidLocationEngine.kt:                    Timber.e(exception)
subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Ably.kt:            Timber.e(exception, "subscribeForEnhancedEvents: Cannot subscribe to enhanced locations messages")
subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Ably.kt:            Timber.e(ablyException)
subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Ably.kt:            Timber.e(exception, "subscribeForPresenceMessages: Cannot subscribe to presence messages")
subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Ably.kt:            Timber.e(ablyException)

[...]

PR #311, opened 2021-04-29
Add missing error routing

KacperKluka:
From the list from the issue https://github.com/ably/ably-asset-tracking-android/issues/201 I was only able to take care of the one about setting route in `Mapbox`. All Ably errors were handled in some previous issues. The errors from S3 are related to publisher example app and the optional S3 feature so I would leave them as logs. The errors from location engine are based on the original location engine implementation and they're handled by Mapbox SDK so we don't have any control over them.
Regarding setting route error I have only escalated it from `Mapbox` to `CorePublisher` because I don't have any useful idea on how we could inform a user about that error. That's because setting route can be triggered from multiple places and we don't always have a callback that we can call with the error.


publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -494,7 +494,11 @@ constructor(
                 removeCurrentDestination(state)
                 state.currentDestination = destination
                 mapbox.setRoute(currentLocation, destination, state.routingProfile) {
-                    enqueue(SetDestinationSuccessEvent(it))
+                    try {
+                        enqueue(SetDestinationSuccessEvent(it.getOrThrow()))
+                    } catch (exception: MapException) {
+                        // TODO - what to do here if setting destination fails
```
QuintinWillison, line 500: This can't be left like this. What you now have is worse than what we had before. This `exception` must be routed somewhere.

QuintinWillison (changes requested):

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
QuintinWillison: Deferred: https://github.com/ably/ably-asset-tracking-android/issues/338

QuintinWillison (approved):

PR #314, opened 2021-04-29
Replace interpolated string definitions with plain literals

QuintinWillison:
Small tidy up thing as I was looking at the build scripts.


KacperKluka (approved):
LGTM :+1:

Issue #313, opened 2021-04-29
Migrate from Kotlin synthetics to Jetpack view binding


QuintinWillison:
We're getting a warning when Gradle is configuring the `:publishing-example-app` and `:subscribing-example-app` projects.

See [here](https://github.com/ably/ably-asset-tracking-android/runs/2465521536?check_suite_focus=true#step:3:24) and [here](https://github.com/ably/ably-asset-tracking-android/runs/2465521536?check_suite_focus=true#step:3:27) for examples. It states:

> Warning: The 'kotlin-android-extensions' Gradle plugin is deprecated. Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.

deanna-lad:
The corresponding Jira issue is https://ably.atlassian.net/browse/SDK-805

PR #316, opened 2021-04-29
Replace jCenter repository with maven Central

QuintinWillison:
This PR is in draft for now because it's generating new lint errors which I can't explain.

I'm seeing locally:

```
> Task :subscribing-example-app:lint FAILED
Ran lint on variant debug: 13 issues found
Ran lint on variant release: 13 issues found
/Users/quintinwillison/code/ably/ably-asset-tracking-android/subscribing-example-app/src/main/res/layout/activity_main.xml:137: Error: Replace the <fragment> tag with FragmentContainerView. [FragmentTagUsage]
  <fragment
   ~~~~~~~~
```

I suspect it may relate to #313.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ buildscript {
 
 plugins {
     id 'org.jlleitschuh.gradle.ktlint' version '9.4.1'
-    id 'org.jetbrains.dokka' version '1.4.30'
+    id 'org.jetbrains.dokka' version '1.4.32'
 }
 
 allprojects {
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
     ext.kotlin_version = '1.4.32'
     repositories {
         google()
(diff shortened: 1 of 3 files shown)


KacperKluka:
@QuintinWillison the lint issue was about using the plain `fragment` element in the layout XML file instead of the `FragmentContainerView`. The latter is a new thing that should be used for fragments, however, it isn't very useful in our case. We need to access the fragment that we add in the layout file and this is something I couldn't easily do with that new container view. This change was forced by updating the `easypermissions` library but since it's only needed in the publishing example app I've moved it there and now the lint isn't complaining about the `fragment` tag :wink:

QuintinWillison:
Thanks for adding those commits, @KacperKluka ... if you're happy with the state of the PR now then can you please approve it?

KacperKluka (approved):
LGTM :+1:

PR #328, opened 2021-05-13
Add slf4j logging support

KacperKluka:
Replaced `Timber` logging library with `slf4j`. The SDKs use only the `slf4j-api` dependency (more precisely they use a Kotlin wrapper of that library) which allows them to use a logger, but the specific implementation must be provided by the SDKs users. If no implementation is provided then a built-in no-op logger is used.
Added a `LogHandler` for the `Ably` that forwards all the log messages to the `slf4j` logger.
Used `logback-android` which is an implementation of `slf4j` in the example apps. Set it so the logs go to the Logcat. Additionally created a sample custom logging handler for the publishing example app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,21 @@ This repository also contains example apps that showcase how the Ably Asset Trac
 
 To build these apps you will need to specify [credentials](#api-keys-and-access-tokens) in Gradle properties.
 
+## Logging
+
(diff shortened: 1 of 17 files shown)


AndyNicks (approved):

QuintinWillison (approved):

QuintinWillison:
We're parking this work for the moment as we work to understand customer needs in this area better.

PR #330, opened 2021-05-14
Bring back section about secrets to the readme.

kavalerov:
SDK users who visit it's Github page should see all information that is needed to get started in the README. This includes the instruction on how to configure required secrets.

This was moved to the CONTRIBUTING previously, but this should also be mentioned in the core README as it is needed by SDK users as well, not just contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,16 @@ This repository also contains example apps that showcase how the Ably Asset Trac
 
 To build these apps you will need to specify [credentials](#api-keys-and-access-tokens) in Gradle properties.
 
+There are, of course, [many other ways](https://docs.gradle.org/current/userguide/build_environment.html) to inject project properties into Gradle builds
+
+## API Keys and Access Tokens
+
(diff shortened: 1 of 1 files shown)


tbedford (approved):

KacperKluka (approved):
LGTM :+1:

QuintinWillison (changes requested):
In respect of:

> needed by SDK users as well, not just contributors

That is not entirely true, so I would like to iterate this PR a little further to make that clear. @kavalerov, I'm happy to take it from here.

kavalerov:
Right now to run example apps one needs to be able to provide secrets to Gradle. This is not related to contributing to the repo or to the SDK - customer may be running these example apps just to see how SDK works.

However, I agree that this is only related to running example apps from source, and not applicable to "production" use of the SDKs. Let me quickly suggest another change if that's ok :)

QuintinWillison (approved):
Thanks, @kavalerov - that's much better. 😁

PR #339, opened 2021-05-20
Add simple log handler

KacperKluka:
I've added a very simple `LogHandler` interface that can be used to receive log messages from Asset Tracking SDKs. It is optional to provide an implementation of this interface. If no implementation is provided then no log messages are emitted.

### DIFF
--- a/core-sdk/src/main/java/com/ably/tracking/logging/LoggingConfigurationModels.kt
+++ b/core-sdk/src/main/java/com/ably/tracking/logging/LoggingConfigurationModels.kt
@@ -1,0 +1,9 @@
+package com.ably.tracking.logging
+
+interface LogHandler {
+    fun logMessage(level: LogLevel, message: String, throwable: Throwable? = null)
+}
+
+enum class LogLevel(val levelInt: Int) {
+    VERBOSE(1), INFO(2), DEBUG(3), WARN(4), ERROR(5)
+}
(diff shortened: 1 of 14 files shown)


AndyNicks (approved):

QuintinWillison (approved):
Elegant. Thanks, @KacperKluka - nice work! 🚀

core-sdk/src/main/java/com/ably/tracking/logging/LoggingConfigurationModels.kt
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Gets called when a log message is sent from the SDK.
+     * This method is called synchronously and can be called from any thread.
+     * Don't run blocking code in this method as this can cause the SDK to malfunction.
+     *
+     * @param level The importance level of the message.
+     * @param message The message text.
+     * @param throwable Optional error object.
+     */
+    fun logMessage(level: LogLevel, message: String, throwable: Throwable? = null)
```
paddybyers, line 16: Why is there explicit support for `Throwable` in the interface but not `ErrorInformation` ?

QuintinWillison: It's exposed via [ConnectionException](https://sdk.ably.com/builds/ably/ably-asset-tracking-android/pull/339/dokka/core-sdk/core-sdk/com.ably.tracking/-connection-exception/index.html), which is of course a subclass of `Throwable`... so, to me, this is a pretty standard pattern / design.

PR #344, opened 2021-05-24
Use ScrollView in publisher kotlin app to fix entering tracking ID

ben-xD:
On small screens, the Publisher Example App written in Kotlin doesn't allow scrolling, and so entering tracking ID is impossible. The fix is to use scroll view

Left is the original, right has the scroll view.
Frame 101


KacperKluka (changes requested):
That's a good spot, thanks :smile: However, you've replaced the `ConstraintLayout` with a `LinearLayout` which is a completely different type of `ViewGroup` thus the layout will probably malfunction (that's why the progress bar on your second screenshot is displayed below the button :wink: ). I'm wondering what would be the best approach with the scrolling :thinking: Maybe let's try to make the button and the screen title fixed in place and only allow to scroll what's between them? In order words let's make only the form fields scroll, WDYT?

BTW If you have some more important things to do I'll be happy to take over this PR from you, just let me know :wink:

KacperKluka (approved):
LGTM :+1:

PR #354, opened 2021-06-01
Fix crash when presence data is missing or in wrong format

KacperKluka:
Hotfix for crashes that occur when the `type` from `PresenceData` is null or the presence data is missing.
In both cases we will just log such presence message and ignore it so the Publisher and Subscriber SDKs won't receive it.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -268,12 +268,26 @@ constructor(
             try {
                 // Emit the current presence messages of the channel
                 channel.presence.get(true).let { messages ->
-                    messages.forEach {
+                    messages.forEach { presenceMessage ->
                         // Each message is launched in a fire-and-forget manner to not block this method on the listener() call
-                        scope.launch { listener(it.toTracking(gson)) }
+                        scope.launch {
+                            val parsedMessage = presenceMessage.toTracking(gson)
+                            if (parsedMessage != null) {
+                                listener(parsedMessage)
+                            } else {
+                                Timber.w("Presence message in wrong format: $presenceMessage")
+                            }
+                        }
                     }
                 }
-                channel.presence.subscribe { listener(it.toTracking(gson)) }
+                channel.presence.subscribe {
+                    val parsedMessage = it.toTracking(gson)
+                    if (parsedMessage != null) {
+                        listener(parsedMessage)
+                    } else {
+                        Timber.w("Presence message in wrong format: $it")
+                    }
+                }
             } catch (exception: AblyException) {
                 throw exception.errorInfo.toTrackingException()
             }
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -275,7 +275,7 @@ constructor(
                             if (parsedMessage != null) {
                                 listener(parsedMessage)
                             } else {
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
```diff
@@ -120,13 +120,17 @@ fun TokenRequest.toAuth(): Auth.TokenRequest =
...
  */
-fun io.ably.lib.types.PresenceMessage.toTracking(gson: Gson) =
-    PresenceMessage(
-        this.action.toTracking(),
-        this.getPresenceData(gson),
-        this.clientId
-    )
+fun io.ably.lib.types.PresenceMessage.toTracking(gson: Gson): PresenceMessage? =
+    this.getPresenceData(gson)?.let { presenceData ->
+
```
owenpearson, line 127: Is there a reason for the blank line here?

owenpearson (approved):
LGTM, thanks

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
KacperKluka: No, I've overlooked it, thanks :wink: https://github.com/ably/ably-asset-tracking-android/pull/354/commits/0bbd53beacbf0466bf89a685ee2c876cea92d802

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -268,12 +268,26 @@ constructor(
...
-                    messages.forEach {
+                    messages.forEach { presenceMessage ->
                         // Each message is launched in a fire-and-forget manner to not block this method on the listener() call
-                        scope.launch { listener(it.toTracking(gson)) }
+                        scope.launch {
+                            val parsedMessage = presenceMessage.toTracking(gson)
+                            if (parsedMessage != null) {
+                                listener(parsedMessage)
+                            } else {
+                                Timber.w("Presence message in wrong format: $presenceMessage")
```
kavalerov, line 278: Very minor, but I would write "unexpected" vs "wrong".

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -268,12 +268,26 @@ constructor(
...
+                            }
+                        }
+                    }
+                }
+                channel.presence.subscribe {
+                    val parsedMessage = it.toTracking(gson)
+                    if (parsedMessage != null) {
+                        listener(parsedMessage)
+                    } else {
+                        Timber.w("Presence message in wrong format: $it")
```
kavalerov, line 288: Same as above - I would prefer "unexpected" over "wrong"

kavalerov (approved):
Looks good to me except for a minor wording issue. 

I also would want to add tests for situations with multiple publishers, but this is probably outside of the scope of this PR, so I will raise this separately.

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/354/commits/f73876e1a64407fd88480721cad44786f1e42a66

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/354/commits/f73876e1a64407fd88480721cad44786f1e42a66

PR #356, opened 2021-06-01
Apply presence data hotfix to the main branch

KacperKluka:
We've worked on a quick hotfix that lead to a new release of the SDKs and now we should apply those changes to the `main` branch.


.github/workflows/publish.yml
```diff
@@ -19,6 +19,10 @@ jobs:
 
     - name: Publish to Repositories
       env:
+        ORG_GRADLE_PROJECT_MAPBOX_DOWNLOADS_TOKEN: ${{ secrets.MAPBOX_DOWNLOADS_TOKEN }}
```
owenpearson, line 22: Didn't you cherry pick this change from a commit that was already on main? Seems weird that this file would have any changes 🤔

owenpearson (approved):
LGTM, thanks

.github/workflows/publish.yml
KacperKluka: Yeah it's quite strange because those changes are already [there](https://github.com/ably/ably-asset-tracking-android/blob/main/.github/workflows/publish.yml) :thinking:

PR #357, opened 2021-06-02
Send trip start and trip end events on a meta channel from Publisher SDK

KacperKluka:
When a trackable is added we're sending trip start event. When a trackable is removed we're sending trip end event. When the publisher is stopped we're sending end trip events for all trackables in the publisher.

I've changed the approach of postponing operations until a raw location is present. Before we had a static field with a `Destination` that was used to postpone setting the destination until any raw location is received. Now we have a list of commands that are run when a raw location is received. This way it's much more flexible and allows to queue multiple operations that depend on raw location.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Constants.kt
+++ b/common/src/main/java/com/ably/tracking/common/Constants.kt
@@ -6,8 +6,14 @@ const val MILLISECONDS_PER_MINUTE = MILLISECONDS_PER_SECOND * 60
 
 const val METERS_PER_KILOMETER = 1000
 
+object ChannelNames {
+    const val METADATA_TRIP = "[meta]asset-tracking:trip-lifecycle"
+}
+
 object EventNames {
     const val ENHANCED = "enhanced"
+    const val TRIP_START = "trip.start"
+    const val TRIP_END = "trip.end"
 }
 
 object ClientTypes {
--- a/common/src/main/java/com/ably/tracking/common/Constants.kt
+++ b/common/src/main/java/com/ably/tracking/common/Constants.kt
@@ -17,6 +17,13 @@ object ChannelNames {
     const val METADATA_TRIP = "[meta]asset-tracking:trip-lifecycle"
 }
 
+/**
+ * The trip event naming approach (e.g. `trip.start`) is similar to what we use in
+ * lifecycle events on Ably metachannels (see
+ * [Lifecycle Events](https://ably.com/documentation/realtime/metachannels#lifecycle-events)).
+ *
+ * This way we align the metachannel message format.
+ */
 object EventNames {
     const val ENHANCED = "enhanced"
     const val TRIP_START = "trip.start"
--- a/common/src/main/java/com/ably/tracking/common/Constants.kt
+++ b/common/src/main/java/com/ably/tracking/common/Constants.kt
@@ -7,6 +7,13 @@ const val MILLISECONDS_PER_MINUTE = MILLISECONDS_PER_SECOND * 60
 const val METERS_PER_KILOMETER = 1000
 
 object ChannelNames {
+    /**
+     * This is a metachannel, which will be defined across the Ably infrastructure.
+     *
+     * **Note**: Simply using `tracking` as the namespace prefix wouldn't be enough as,
+     * in the wider Ably context, `tracking` can mean many things. Whereas,
+     * `asset-tracking` specifies a product.
+     */
     const val METADATA_TRIP = "[meta]asset-tracking:trip-lifecycle"
 }
 
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -132,6 +132,24 @@ interface Ably {
      * @throws ConnectionException if something goes wrong.
      */
     suspend fun close(presenceData: PresenceData)
+
+    /**
+     * Sends trip metadata when the trip starts.
+     *
+     * @param metadata The metadata of a tracked trackable.
+     *
+     * @throws ConnectionException if something goes wrong.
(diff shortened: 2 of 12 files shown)


AndyNicks (approved):

common/src/main/java/com/ably/tracking/common/Constants.kt
```diff
@@ -6,8 +6,14 @@ const val MILLISECONDS_PER_MINUTE = MILLISECONDS_PER_SECOND * 60
 
 const val METERS_PER_KILOMETER = 1000
 
+object ChannelNames {
+    const val METADATA_TRIP = "[meta]asset-tracking:trip-lifecycle"
```
QuintinWillison, line 10: If this is a channel name prefix, then shouldn't it be `tracking` (not `asset-tracking`) to match [what's defined elsewhere](https://github.com/ably/ably-asset-tracking-android/blob/774c02cc52038e91e5c12a3a8a44645e9961ff78/common/src/main/java/com/ably/tracking/common/Ably.kt#L137)?

common/src/main/java/com/ably/tracking/common/Constants.kt
```diff
@@ -6,8 +6,14 @@ const val MILLISECONDS_PER_MINUTE = MILLISECONDS_PER_SECOND * 60
...
+}
+
 object EventNames {
     const val ENHANCED = "enhanced"
+    const val TRIP_START = "trip.start"
```
QuintinWillison, line 15: Is there a foundation for this naming convention?

It seems incongruous with the concrete value of `ENHANCED`.

QuintinWillison (changes requested):
Some immediate observations.

However, I'm also wondering if there's a corresponding issue defining the scope of this work?

common/src/main/java/com/ably/tracking/common/Constants.kt
KacperKluka: As far as I understand this channel shouldn't be treated in the same way as other trackable channels. This is a unique channel that will be used across all AAT users. Is it right @AndyNicks @kavalerov ? :wink:

common/src/main/java/com/ably/tracking/common/Constants.kt
KacperKluka: @AndyNicks @kavalerov I think that you've come up with those names, was there any particular reason behind them? :wink:

AndyNicks: If this naming convention conflicts with the agreed  scheme, let's change it.

QuintinWillison: As far as I'm aware there's no agreed scheme, it just seems like we may be forming one. 😁

AndyNicks: So let it be :-)

common/src/main/java/com/ably/tracking/common/Constants.kt
kavalerov: This is a metachannel, which will be defined across Ably infrastructure, therefore `tracking` wouldn't be enough (in wider Ably context tracking can mean many things, whereas `asset-tracking` specifies a product). So it should be `asset-tracking`.

common/src/main/java/com/ably/tracking/common/Constants.kt
kavalerov: So this approach (`trip.start`) is similar to what we use in lifecycle events on Ably metachannels (see more: https://ably.com/documentation/realtime/metachannels#lifecycle-events).

This way we align the metachannel message format. 

Now, I can see potential reason for us changing the `ENHANCED` in the future, but I think it would be outside of the scope of this PR for sure.

common/src/main/java/com/ably/tracking/common/Constants.kt
QuintinWillison: Great, thanks @kavalerov ... @KacperKluka could you please add commentary to that effect and add the commit reference to this conversation when done so I can resolve. Thanks.

common/src/main/java/com/ably/tracking/common/Constants.kt
QuintinWillison: Great, thanks. @KacperKluka, could you please add commentary including the essence of what Ivan states above. Thanks.

common/src/main/java/com/ably/tracking/common/Constants.kt
QuintinWillison: https://github.com/ably/ably-asset-tracking-android/pull/357/commits/a72c54c1cac051079205596db03562a24abdacac

common/src/main/java/com/ably/tracking/common/Constants.kt
QuintinWillison: https://github.com/ably/ably-asset-tracking-android/pull/357/commits/b9a00dcda319422ff9612a317a386fed44d61bbd

kavalerov:
Just to confirm - we have tested that this does produce all required events on the metachannel on Ably (as in - one can see them by subscribing to the metachannel, i.e. in the dashboard)?

QuintinWillison:
I've not tested it myself. I don't know whether @KacperKluka did while developing this.

kavalerov:
Ok, then I don't think this should've been merged

QuintinWillison:
Thanks, @kavalerov - I agree. 😁 

As discussed in yesterday afternoon's retro and planning meeting, we failed to validate against the acceptance criteria specified in https://github.com/ably/ably-asset-tracking-android/issues/374.

We are going to address that under https://github.com/ably/ably-asset-tracking-android/issues/376.

PR #379, opened 2021-06-21
Send failed location updates

KacperKluka:
When sending a location update fails then try to resend it. If it fails again then add the location to the `skippedLocations` list.

In order to implement that we've changed the `channel.publish()` method from a sync version to an async one. Thanks to that we now know when a location is send and if the sending procedure ends with success or failure. 
This has introduced some complexity to the way of handling state connected to published locations. After considering a few different approaches we've decided that we won't send any other messages for a given trackable if it already has a pending message that's being published. We're collecting all location updates that appear during publishing a message in the waiting list. After a pending message is published we're taking the next location update from that waiting list and process it.

This approach has one potential problem: when we're processing next location updates from the waiting list we're not sending them as separate events (as this introduces ordering issues) but rather we're iterating them until we find an location update that can be sent. If such thing is found then the queue is "unblocked" and starts processing other events. In worst case scenario the queue can be "blocked" for as long as it would take to process all the location updates in the waiting list. On the other hand to process them it only needs to compare a few things so this probably won't be a real problem I think :thinking:

When resending the message we're using the idempotent publishing feature of Ably to avoid sending duplicate location updates.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/LocationsPublishingState.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/LocationsPublishingState.kt
@@ -1,23 +1,63 @@
 package com.ably.tracking.publisher
 
+/**
+ * Class responsible for managing state connected to location updates that are going to or are being published.
+ */
 internal class LocationsPublishingState {
+    /**
+     * The maximum number of retries after a location update is considered to be failed.
+     */
     private val MAX_RETRY_COUNT = 1
+
+    /**
+     * Stores information about trackables that have currently pending messages.
+     */
     private val pendingMessages: MutableSet<String> = mutableSetOf()
+
+    /**
+     * Stores location updates that are waiting to be processed, for each trackable independently.
+     */
     private val waitingLocationUpdates: MutableMap<String, MutableList<EnhancedLocationChangedEvent>> = mutableMapOf()
+
+    /**
+     * Stores the number of retries of the current location update, for each trackable independently.
+     */
     private val retryCounter: MutableMap<String, Int> = mutableMapOf()
 
+    /**
+     * Marks that the specified trackable has a pending message.
+     *
+     * @param trackableId The ID of the trackable.
+     */
     fun markMessageAsPending(trackableId: String) {
         pendingMessages.add(trackableId)
     }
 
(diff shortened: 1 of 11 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/LocationsPublishingState.kt
```diff
@@ -0,0 +1,56 @@
+package com.ably.tracking.publisher
+
+internal class LocationsPublishingState {
```
QuintinWillison, line 3: It would be nice to see some unit tests for this class, in isolation.

publishing-sdk/src/main/java/com/ably/tracking/publisher/SkippedLocations.kt
```diff
@@ -0,0 +1,30 @@
+package com.ably.tracking.publisher
+
+import com.ably.tracking.Location
+
+internal class SkippedLocations {
```
QuintinWillison, line 5: Also very unit testable.

QuintinWillison (approved):
Looks sensible to me, especially given the public API for SDK users remains the same.

My only concerns relate to missed opportunities to unit test plus lack of documentation on internal classes (though lack of docs can be countered with unit tests that proactively demonstrate what a class is meant to do 😉).

publishing-sdk/src/main/java/com/ably/tracking/publisher/SkippedLocations.kt
KacperKluka: Good idea, added both tests and docs :wink: https://github.com/ably/ably-asset-tracking-android/pull/379/commits/02baa9c189aa3b3bf22d6865d723fb58f02f5312

publishing-sdk/src/main/java/com/ably/tracking/publisher/LocationsPublishingState.kt
KacperKluka: Good idea, added both tests and docs :wink: https://github.com/ably/ably-asset-tracking-android/pull/379/commits/c044fbdff9f1932f8674957041c35e886e9d0033

PR #380, opened 2021-06-22
Increase minSdk to 24 for example apps

KacperKluka:
Changed minSdk for example apps from 21 to 24.


QuintinWillison (changes requested):
This would be less repetitive were you to introduce a second ternary operator (multiple instances of `minSdkVersion` gives me future maintenance risk wobbles).

QuintinWillison (approved):

PR #381, opened 2021-06-22
Fix overlapping text in subscriber example app

KacperKluka:
The resolution values were overlapping their labels. In order to fix it We've aligned those value texts to the one with the longest label ("Min displacement").
device-2021-06-22-094017


AndyNicks (approved):
LGTM

ben-xD (approved):

PR #382, opened 2021-06-22
Add a map to the Android Publisher example

KacperKluka:
I added the Mapbox dependency to the publishing example app and created a simple activity that displays a map from Mapbox. On the map the current asset location is displayed and the camera follows the asset location.

I've used an older version of Mapbox dependency (v8) because we aim to check compatibility with older versions.

Since there are no design requirements at the time of implementing this feature, I've taken the simplest approach possible to deliver it as fast as possible :wink: 

Also to keep things simple I've used a "Marker" approach of displaying asset location icon instead of using the recommended (and more complicated) approach of using the Annotation Plugin which probably is causing the marker to be displayed in black color instead of red sometimes :thinking: Let me know if that's something we need to fix or if it's okay for our purpose?
device-2021-06-22-114751


AndyNicks (approved):
LGTM

PR #385, opened 2021-06-24
Remove previousState from ConnectionStateChange

KacperKluka:
The `previousState` field was removed from `ConnectionStateChange` as it wasn't used anywhere.


AndyNicks (approved):

QuintinWillison (approved):
Technically a breaking change so we should ideally mention that in the change log when we do the next release. We're still in beta (pre-release) and it's unlikely users will be relying on this API to track previous state.

PR #386, opened 2021-06-24
Remove obsolete TODO comments from Publisher and Subscriber builders

KacperKluka:
After looking at the `TODO`s in the builder functions from both Subscriber and Publisher SDKs I've removed them because they're obsolete now.

The first one about using `@JvmStatic` was more of an information than a TODO and it was introduced at the very start of the project. Now, when we're more familiar with Kotlin and we're using `@JvmStatic` elsewhere I think it doesn't bring any value.

The second one was about providing a predefined default builder instead of an empty builder. This also seemed outdated since some time ago we've discussed which builder fields are required and which are optional. When the user doesn't set those optional fields then they'll be set to some default values, so it basically works as a predefined builder.


AndyNicks (approved):

QuintinWillison (approved):

QuintinWillison:
I can't see how the test failure could possibly relate to the change that's been made so I'm going to land this pull request.

PR #389, opened 2021-06-29
Show custom notification for Mapbox trip service

KacperKluka:
Replaced the default Mapbox trip notification with a custom one and added a method to the builder which allows users to specify their own notification. I had to exclude the original notification module and specify which of our classes will be responsible for handling the notification. I've been following [this](https://docs.mapbox.com/android/navigation/guides/modularization) and [this](https://github.com/mapbox/mapbox-base-android/blob/master/MODULARIZATION.md) guide.

Thanks to that we're able to run 2 services under one notification. All we need to do is just pass the publisher service notification to the publisher SDK and it will be used for the Mapbox trip service.

Additionally, the original notification had a button that allowed the user to stop Mapbox service while Publisher was working which led to malfunction.


AndyNicks (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/MapboxTripNotification.kt
```diff
@@ -0,0 +1,60 @@
...
+    override fun getNotification(): Notification = notification.getNotification()
+    override fun getNotificationId(): Int = notification.getNotificationId()
+    override fun onTripSessionStarted() = Unit
+    override fun onTripSessionStopped() = Unit
+    override fun updateNotification(routeProgress: RouteProgress?) = Unit
+}
+
+internal class DefaultMapboxTripNotification(context: Context) : TripNotification {
+    private val NOTIFICATION_CHANNEL_ID = "DefaultTripNotificationChannelId"
+    private val NOTIFICATION_ID = 819203
```
QuintinWillison, line 27: I'm curious as to where this number comes from - it seems a bit "magic". Could you perhaps add a comment to explain?

QuintinWillison (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/MapboxTripNotification.kt
KacperKluka: Yes, so this is one of those things that are not very clear in Android :sweat_smile: In [the docs](https://developer.android.com/training/notify-user/build-notification#notify) they say that `notificationId is a unique int for each notification that you must define` and that it's used to `update or remove the notification`. 
However, they don't let us know if this should be unique device-wise, app-wise or only SDK-wise :unamused: 
So as my rule of thumb, I try to come up with some random number that hopefully won't collide with any other notification ID. But this may be just my superstition :sweat_smile:

QuintinWillison: @KacperKluka Looking at [the API reference for the `notify` method](https://developer.android.com/reference/android/app/NotificationManager#notify(int,%20android.app.Notification)), it states:

> An identifier for this notification unique within your application

Which, I feel, clears up the confusion (i.e. it is "app-wise").

I'm not happy with this being baked into the SDK as there is no reason why the app developer who has embedded our SDK might not pluck this number out of the air themselves also.

It's also made me take a closer look at what we're doing anyway with these notifications which is, in turn, making me challenge whether this is something that's appropriate for the SDK to be doing at all! Not to mention the lack of [localisation](https://developer.android.com/guide/topics/resources/localization) opportunity for the textual strings used.

QuintinWillison (changes requested):
On [further reflection](https://github.com/ably/ably-asset-tracking-android/pull/389#discussion_r665669644) this needs more work. We'll discuss internally and then move this pull request forwards.

KacperKluka:
After today's call we've decided to not provide a default notification but to make the `notification()` builder method (which will be renamed to `backgroundTrackingNotificationProvider()` as a part of those changes) a required one.
Moreover, we'll make some changes to the `AssetTrackingNotification` so it will be responsible of providing the notification only and the notification ID will become a param of the builder method.

publishing-sdk/src/main/java/com/ably/tracking/publisher/MapboxTripNotification.kt
KacperKluka: I've made the changes we've discussed on our today's call :wink: Here are they https://github.com/ably/ably-asset-tracking-android/pull/389/files/f05c3610d4303b202b0f274abd2072ebd9fc145e..5d7df052bb647abdd1e62084c4fada47a4235b80

AndyNicks (approved):

QuintinWillison (approved):

ben-xD (approved):

PR #390, opened 2021-07-02
Adopt SDK upload action

QuintinWillison:
I was convinced this had been done already for this repository, but it seems not.

Following in the path trodden by compatriots [ably-asset-tracking-swift](https://github.com/ably/ably-asset-tracking-swift/blob/2db19a54624562bb45314dde50e13718d1a6364c/.github/workflows/docs.yml#L31) and [ably-flutter](https://github.com/ably/ably-flutter/blob/0e2be4e8cee30925eaa6ce4a2190deab07391136/.github/workflows/docs.yml#L28).


KacperKluka (approved):
LGTM :+1:

PR #398, opened 2021-08-02
Explicitly specify DTO field names during serialization

KacperKluka:
Specified the serialized names of DTO classes.
Before I could do that I had to create a separate model for the `LocationMessage` https://github.com/ably/ably-asset-tracking-android/pull/398/commits/1d7a43820171292971b7006d4c83d25f79a34629 because previously we used the `GeoJsonMessage` which is also exposed in our public API. Thus I wasn't able to use the Gson library annotations with `GeoJsonMessage` to not expose that dependency in our API.
Then I've moved some mappers to a more appropriate location https://github.com/ably/ably-asset-tracking-android/pull/398/commits/eb21bb0c114bed3311c650b5ba30e54c509530e9 (and fixed tests https://github.com/ably/ably-asset-tracking-android/pull/398/commits/eae265aada2aa73bf73c514b8ea798a8783067c3).
In the end I finally annotated the DTO classes with explicit field names https://github.com/ably/ably-asset-tracking-android/pull/398/commits/8c63e89ab26b689bece95cd764c7b13427814d83.


QuintinWillison (changes requested):
As [discussed in Slack](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1627917316034500), I think it would be helpful for this pull request to set `minifyEnabled true` to the integration tests.

QuintinWillison (approved):

KacperKluka:
I've prepared an issue for running integration tests with `minifyEnabled true` https://github.com/ably/ably-asset-tracking-android/issues/400

PR #419, opened 2021-08-31
Run integration testing workflow on multiple API levels

KacperKluka:
Used the Github Actions matrix strategy to run the `emulate` workflow on multiple API levels.

### DIFF
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/NotificationHelpers.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/NotificationHelpers.kt
@@ -8,6 +8,9 @@ import android.os.Build
 const val NOTIFICATION_CHANNEL_ID = "test-notification-channel"
 
 fun createNotificationChannel(context: Context) {
+    // This guard is required by the official Android documentation as the notification
+    // channels API is not available in the support library.
+    // https://developer.android.com/training/notify-user/channels#CreateChannel
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
         context.getSystemService(NotificationManager::class.java).apply {
             createNotificationChannel(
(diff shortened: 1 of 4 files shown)


android-test-common/src/main/java/com/ably/tracking/test/android/common/NotificationHelpers.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.content.Context
+import android.os.Build
+
+const val NOTIFICATION_CHANNEL_ID = "test-notification-channel"
+
+fun createNotificationChannel(context: Context) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
QuintinWillison, line 11: Looks curious (zero!?). What does this achieve?

QuintinWillison (approved):

android-test-common/src/main/java/com/ably/tracking/test/android/common/NotificationHelpers.kt
KacperKluka: It's not zero but O as in Oreo :smile: This is needed as the channel creation method isn't backward compatible and calling it on devices with OS lower than Oreo will result in a runtime crash :wink:

QuintinWillison: That is hilarious, though perhaps worthy of a comment to clarify?

KacperKluka: Yup, that's quite funny, but that's how we do things in the Android world :smile: Added the comment https://github.com/ably/ably-asset-tracking-android/pull/419/commits/8d66606295d2265d6bd7788848cec83bae4d4edd

PR #426, opened 2021-09-02
Fix flakey publisher resolution test

KacperKluka:
It turns out that the `corePublisher.enqueue()` method was behaving unexpectedly from time to time. We were putting quite a few (30 in this case, but I've tested on numbers like 5000) events to the queue in a loop without any delay. This caused the underlying `scope.launch { sendEventChannel.send(event) }` code to sometimes execute out of order. This shouldn't happen in a real-life use, so I'd just added a very small delay to fix that issue :wink:


KacperKluka:
Rerunning integration tests workflow as another flakey test caused it to fail :sweat_smile:

QuintinWillison (changes requested):
As discussed in our standup a few days ago, I'm not comfortable with this fix as it suggests there's something deeper going on in the implementation that is the root cause.

I think this needs more investigation.

KacperKluka:
I've made a research and it looks like nothing is wrong with our code, it behaves as expected. There is no guarantee that the `launch()` calls will be executed in the calling order. I couldn't find information about it in the [docs](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html) but I've found multiple comments that confirm this behaviour (like [this](https://discuss.kotlinlang.org/t/coroutines-order/17713/9) and [this](https://discuss.kotlinlang.org/t/coroutines-execution-order-when-launching-them-sequentially/22069/7)). Also you can look in the coroutines docs about [Dispatchers and threads](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#dispatchers-and-threads) and run the example with multiple launch statements - they will execute in random order.

We have at least 2 possible ways that should allow us to preserve the execution order:
- Use the [start](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-start/index.html) parameter of the `launch()` method. Probably the `UNDISPATCHED` one would do the trick, however, this is part of the experimental API and in my case, the tests stopped producing any results when I tried to use that (but maybe it's just my machine). We should also check if that isn't somehow blocking the main thread and breaking the rest of the SDK.
- Use a custom dispatcher with a single thread which should persist the execution order. I've tried that and this solution works (the test always succeeded) but it requires some careful management of the aforementioned thread. And I'm not sure how it will affect the rest of the SDK as now we're clearly using a dispatcher with multiple threads in the pool. We could use the [newSingleThreadContext](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/new-single-thread-context.html) for that but it's obsolete API (and it produced the same lack of test results as the start param from the first point did for me) so I've tested the `Executors.newSingleThreadExecutor().asCoroutineDispatcher()` which worked fine for that particular test. If we will go this way we need to think about the cleanup of this thread. Because it is a custom dispatcher we need to handle its lifecycle manually. So we need to call `close()` on it when the SDK is stopped, but we can't do that since the events queue should stay alive after the SDK is stopped. We would have to add another method like 'destroy()` that would then clean this dispatcher.

Both solutions seem to require a lot of effort to implement and I'm still wondering if they're worth it :thinking: I doubt that we will receive location updates without at least 3ms of delay between them. So the problem seems to exist only in our test.

QuintinWillison:
I think the use of [newSingleThreadExecutor()](https://developer.android.com/reference/kotlin/java/util/concurrent/Executors#newSingleThreadExecutor()) feels like the most robust solution to me.

I'm not sure that the engineering required to manage the cleanup of that executor is even required, as it should be acceptable to have a single one of these shared across this process (i.e. a singleton for all publisher instances). That is based on the assumption that the tasks running through it are never designed to be blocking in nature.

The implication is that once the first publisher is started, the singleton is created and then lives for the lifespan of the application process. This thing, even if it's hanging on to an idle thread, should be very lightweight - so I think that worry about when it's destroyed is a premature optimisation.

KacperKluka:
As we've discussed I've changed our dispatchers to single-threaded ones https://github.com/ably/ably-asset-tracking-android/pull/426/commits/f9cb4e45be3adaa51ffbfdbb8a00a1f0487f82c3
This should not only help with the test but also overall improve our events queue mechanism :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -54,6 +55,18 @@ internal fun createCorePublisher(
     return DefaultCorePublisher(ably, mapbox, resolutionPolicyFactory, routingProfile, logHandler)
 }
 
+/**
+ * This is a private static single thread dispatcher that will be used for all the [Publisher] instances.
+ * To assure that we process our events in the FIFO order we need to use a single threaded dispatcher.
+ * Because of that, all calls to the [launch] will execute in the order of invocation. Because threads
+ * are expensive resources we don't want to create a separate one for each instance. Therefore, we
+ * have one dispatcher shared across all [Publisher] instances. In an edge case scenario where multiple
+ * instances are active at the same time, this may lead to some slowdowns, as all the work performed
```
QuintinWillison, line 64: While I appreciate your reasons for wanting to add this particular point to the commentary, I think we should keep as far away from premature optimisation in comments also. 😉 

```suggestion
 * instances are active at the same time, this could theoretically lead to some slowdowns, as all the work performed
```

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -40,13 +41,25 @@ internal fun createCoreSubscriber(
     return DefaultCoreSubscriber(ably, initialResolution, trackableId)
 }
 
+/**
+ * This is a private static single thread dispatcher that will be used for all the [Subscriber] instances.
```
QuintinWillison, line 45: The need to repeat this comment in both publisher and subscriber makes me think that maybe we should have a static core utility method to create one of these dispatchers, with a really intentional name like `createSingleThreadDispatcher` and then put this comment on the interface for that method.

QuintinWillison (changes requested):

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/426/commits/9ea1ef6e643acd0fa35834c760d7235ca5783d75

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/426/commits/9ea1ef6e643acd0fa35834c760d7235ca5783d75

QuintinWillison (approved):

PR #427, opened 2021-09-02
Stop the publisher when the PublisherService is destroyed

KacperKluka:
To stay on the safe side we're stopping the publisher when the `PublisherService` is being destroyed (which happen when it's stopped by user or force closed by the OS). Without it there's a chance that there will be multiple active publishers.


publishing-example-app/src/main/java/com/ably/tracking/example/publisher/PublisherService.kt
```diff
@@ -65,6 +66,11 @@ class PublisherService : Service() {
 
     override fun onBind(intent: Intent?): IBinder = binder
 
+    override fun onDestroy() {
+        scope.launch { publisher?.stop() }
```
QuintinWillison, line 70: I think there should be some commentary here, or perhaps on this method, to explain why we're doing this and why it's important.

QuintinWillison (changes requested):

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/PublisherService.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/427/commits/034980adfb4168e89c64c5bcf7f714bacda8aa4b

QuintinWillison (approved):

PR #433, opened 2021-09-07
Change the way of setting build config fields

KacperKluka:
As suggested I've tried to get rid of the 'buildType` param from the closures responsible for setting additional build config fields. In order to use the delegates mechanism we need to explicitly set the delegate object. This forces us to write additional code in each place where we want to use those closures. For example instead of
```groovy
applyVersionConfigFields delegate
```

we would have to write
```groovy
applyVersionConfigFields.delegate = delegate
applyVersionConfigFields()
```

and if we would have forgotten to set the delegate then the method would crash.

To somehow use the delegates mechanism and still keep things simple to use I've added a special helper method `apply` which wraps that delegate assignment and allows us to write the configuration more like a spoken language: `apply versionConfigFields on it` (where `it` is the build type).

WDYT? Does it look better now? It's more complicated than it was before and I'm wondering if we should change it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -142,30 +142,34 @@ subprojects {
                       'UnusedIds'
             }
 
-            def applySecretBuildConfigFields = { buildType ->
-                // Configuration that must not be applied to SDK project main source set builds for the release variant/buildType.
-                // This configuration is, however, required for Android's connectedCheck task (which uses the androidTest source set
-                // with the debug variant/buildType).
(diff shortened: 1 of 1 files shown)


QuintinWillison (approved):
I prefer the elegance of not needing the `buildType.` prefixes within the actions, plus the lack of requirement to even pass that instance to them. Your `apply` function is also clean and readable (though first time I've seen square brackets in Groovy) so I think, on balance, this is definitely an improvement. Thanks!

PR #435, opened 2021-09-15
Fix crash when Mapbox is started and stopped without starting a trip

KacperKluka:
Mapbox wrapper was crashing when passing an empty `tripHistoryString` to the `mapToReplayEvents()`. I've created a test for this issue and fixed it with a simple "if not empty" check :wink:

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
@@ -190,8 +190,10 @@ internal class DefaultMapbox(
                 stopTripSession()
                 mapboxReplayer?.finish()
                 val tripHistoryString = retrieveHistory()
-                val historyEvents = ReplayHistoryMapper().mapToReplayEvents(tripHistoryString)
-                locationHistoryListener?.invoke(LocationHistoryData(historyEvents.toGeoJsonMessages()))
+                if (tripHistoryString.isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
```diff
@@ -190,8 +190,10 @@ internal class DefaultMapbox(
                 stopTripSession()
                 mapboxReplayer?.finish()
                 val tripHistoryString = retrieveHistory()
-                val historyEvents = ReplayHistoryMapper().mapToReplayEvents(tripHistoryString)
-                locationHistoryListener?.invoke(LocationHistoryData(historyEvents.toGeoJsonMessages()))
+                if (tripHistoryString.isNotEmpty()) {
```
QuintinWillison, line 193: For future maintainers, perhaps a comment would help:

```suggestion

                // MapBox's mapToReplayEvents method crashes if passed an empty string,
                // so check it's not empty to prevent that.
                // (see: https://github.com/ably/ably-asset-tracking-android/issues/434)
                if (tripHistoryString.isNotEmpty()) {
```

QuintinWillison (changes requested):

publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/commit/66d38c2f3dda06fd428a15505c052903d5f57e39

PR #436, opened 2021-09-15
Add test checking the order of set route callbacks

KacperKluka:
When I was working on adding the test that checks the order of `mapbox.setRoute()` callbacks I've encountered an unexpected problem. The test was working when I ran it in isolation, however, when it was run as a part of the whole test class suite then it was randomly failing. This is an example of the error it was throwing
```
A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x30 in tid 24409 (Thread-3), pid 24330 (.publisher.test)
```
This leads me to believe that there is some issue with the native C/C++ code as the `SIGSEGV` error means that the process tries to access a memory location that doesn't belong to it. With this in mind, I've tried to reproduce the error when running the test in isolation. I've managed to do that by creating another instance of the Mapbox wrapper and stopping it right before creating the instance that was being tested. Then I changed the first instance to the `MapboxNavigation` to be sure that it isn't caused by some operations we do in the wrapper class. It was failing randomly but always with the same error. Then I've introduced a 1-second delay between the stopping of the first instance and the creation of the second one and it seemed to fix the issue.

I'm out of ideas on how to fix it and I'm wondering why is that happening? I thought that the `MapboxNavigation.onDestroy()` method will synchronously stop the Mapbox and free any resources it uses, but maybe that's not the case? :thinking: 

---

**Update**
I've finally been able to get more logs from this crash:
```
    --------- beginning of crash
2021-09-23 16:45:35.788 950-976/com.ably.tracking.publisher.test A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x30 in tid 976 (Thread-3)
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG: Build fingerprint: 'google/sdk_google_phone_x86/generic_x86:7.0/NYC/4409132:user/release-keys'
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG: Revision: '0'
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG: ABI: 'x86'
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG: pid: 950, tid: 976, name: Thread-3  >>> com.ably.tracking.publisher.test <<<
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x30
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG:     eax 00000030  ebx a7bc77b0  ecx 00000030  edx 00000016
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG:     esi 00000000  edi 00000030
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG:     xcs 00000073  xds 0000007b  xes 0000007b  xfs 0000003b  xss 0000007b
2021-09-23 16:45:35.844 1037-1037/? A/DEBUG:     eip a7b649e3  ebp 8d202588  esp 8d202570  flags 00010206
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG: backtrace:
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #00 pc 000759e3  /system/lib/libc.so (pthread_mutex_lock+35)
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #01 pc 00133640  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libmapbox-common.so (_ZNSt6__ndk15mutex4lockEv+32)
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #02 pc 0028e317  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #03 pc 0028e026  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #04 pc 002b5d01  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #05 pc 002b6319  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #06 pc 00284eac  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #07 pc 002b66d3  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #08 pc 002ec417  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #09 pc 002ecf9d  /data/app/com.ably.tracking.publisher.test-1/lib/x86/libnavigator-android.so
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #10 pc 00074fb2  /system/lib/libc.so (_ZL15__pthread_startPv+210)
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #11 pc 0002026e  /system/lib/libc.so (__start_thread+30)
2021-09-23 16:45:35.846 1037-1037/? A/DEBUG:     #12 pc 0001e046  /system/lib/libc.so (__bionic_clone+70)
```

### DIFF
--- a/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/MapboxTests.kt
+++ b/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/MapboxTests.kt
@@ -1,15 +1,23 @@
 package com.ably.tracking.publisher
 
+import android.annotation.SuppressLint
 import android.app.Notification
 import android.content.Context
 import androidx.core.app.NotificationCompat
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
+import com.ably.tracking.Location
 import com.ably.tracking.connection.Authentication
 import com.ably.tracking.connection.ConnectionConfiguration
 import com.ably.tracking.test.android.common.NOTIFICATION_CHANNEL_ID
+import com.ably.tracking.test.android.common.UnitExpectation
 import com.ably.tracking.test.android.common.createNotificationChannel
+import com.ably.tracking.test.android.common.testLogD
 import com.google.gson.Gson
+import com.mapbox.module.Mapbox_TripNotificationModuleConfiguration
+import com.mapbox.navigation.base.trip.notification.TripNotification
+import com.mapbox.navigation.core.MapboxNavigation
+import org.junit.Assert
 import org.junit.Test
 import org.junit.runner.RunWith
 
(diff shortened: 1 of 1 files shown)


moritzzzzz:
@KacperKluka The `Mapbos.setRoute()` calls I assume will trigger HTTP requests to request routes from the Mapbox API.  As these are asynchronous requests, they will run on a worker thread and will try to write their result back to the memory location that they directed to when initiated. 

`SIGSEGV` points to an access violation.

 By destroying the Mapbox object, you probably also erase the callback, including the memory allocation for storing the result.

--> What happens if you destroy Mapbox and do not re-create it? Does it throw a different error/warning?

> Then I've introduced a 1-second delay between the stopping of the first instance and the creation of the second one and it seemed to fix the issue.

--> This would lead to the assumption that the new Mapbox object now occupies the memory that the worker thread would write to and does not allow writing to it.

To not run into such an issue all async functions have to have finished before reallocating memory. You could also run the requests in an async task that is cancellable and cancel it.

KacperKluka:
@moritzzzzz thanks for your response :bow: 
> The `Mapbos.setRoute()` calls I assume will trigger HTTP requests to request routes from the Mapbox API

It is [calling](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt#L254) the `mapboxNavigation.requestRoutes()` method so I think that's right :wink: 

> By destroying the Mapbox object, you probably also erase the callback, including the memory allocation for storing the result.

This also sounds logical but I think that it shouldn't be the case here. The instance that's being destroyed doesn't perform any operations. It is only created and immediately stopped. The second instance is created, it does some work with callbacks, it waits for the callbacks to return and only then it stops the instance.

> --> What happens if you destroy Mapbox and do not re-create it? Does it throw a different error/warning?

In our case, nothing would happen as the second instance is the one who performs the work. But if you're asking if the error occurs if there's only one instance then the answer is no. Only creating an instance previous to the one that will do the work is making the trouble.

> --> This would lead to the assumption that the new Mapbox object now occupies the memory that the worker thread would write to and does not allow writing to it.

Is it still true if the first Mapbox object was only created and stopped and didn't perform explicitly any other operations? :thinking: 

To summarise: our issue is that we're creating a Mapbox instance that doesn't do anything more than stopping itself and this is causing the next Mapbox instance to fail. So there are no (explicit) ongoing async callbacks in the first instance that we could wait on before using the next instance. Do you think that there are some hidden async things happening when we're starting and stopping the first instance that we're not aware of? :thinking:

KacperKluka:
Closing as we need to wait for Mapbox to resolve their issue first.

PR #437, opened 2021-09-15
Pull request checks: run against HEAD, not merge commit

QuintinWillison:
I've started to get fed up with the "Conflict merging main into" Error instances we get for the docs workflow when running the SDK Upload Action step. It slows us down as it results in checks failing.

I'm going to prototype this particular fix to the problem in this repository for a few weeks, with the plan being to roll it out to workflows in other repositories soon after that as our new, conformed way of using the checkout action.

Full context: https://github.com/ably/sdk-upload-action/issues/19

### DIFF
--- a/.github/workflows/assemble.yml
+++ b/.github/workflows/assemble.yml
@@ -8,7 +8,10 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
+      # Checkout HEAD, don't create merge commit. See https://github.com/ably/sdk-upload-action/issues/19
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
       # Gradle 7 requires Java 11 to run
       - uses: actions/setup-java@v2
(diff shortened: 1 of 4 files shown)


QuintinWillison:
Well, that was a flop. We're [seeing](https://github.com/ably/ably-asset-tracking-android/pull/437/checks?check_run_id=3614811443#step:5:10) on the docs workflow:

    Error: Conflict merging main into 57f7d97f0d310785ab8804b1fd85db8bd2c38c54.

Which really confuses me.

That SHA does, indeed, appear to be `HEAD` of the branch underlying this pull request (at the time of writing this, with just one commit): https://github.com/ably/ably-asset-tracking-android/pull/437/commits/57f7d97f0d310785ab8804b1fd85db8bd2c38c54

So, @owenpearson, does this point back to the issue being with the SDK Upload Action perhaps?

owenpearson:
@QuintinWillison I think you're right. See [my comment](https://github.com/ably/sdk-upload-action/issues/19#issuecomment-920410347) on the sdk-upload-action issue.

QuintinWillison:
I've just kicked the docs workflow off again.

QuintinWillison:
We believe that the change represented by this pull request was the wrong solution, unrelated to the problem we were facing which we're confident has been fixed by https://github.com/ably/sdk-upload-action/pull/26.

PR #439, opened 2021-09-17
Detach a channel when disconnecting from it in the Ably wrapper

KacperKluka:
Calling `detach()` on a channel and waiting for it to succeed during the `disconnect()` method of the `Ably` wrapper class.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -260,8 +260,16 @@ constructor(
                         override fun onSuccess() {
                             channelToRemove.unsubscribe()
                             channelToRemove.presence.unsubscribe()
-                            channels.remove(trackableId)
-                            callback(Result.success(Unit))
+                            channelToRemove.detach(object : CompletionListener {
+                                override fun onSuccess() {
+                                    channels.remove(trackableId)
+                                    callback(Result.success(Unit))
+                                }
+
+                                override fun onError(reason: ErrorInfo) {
+                                    callback(Result.failure(reason.toTrackingException()))
+                                }
+                            })
                         }
 
                         override fun onError(reason: ErrorInfo) {


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -260,8 +260,16 @@ constructor(
                         override fun onSuccess() {
                             channelToRemove.unsubscribe()
                             channelToRemove.presence.unsubscribe()
-                            channels.remove(trackableId)
-                            callback(Result.success(Unit))
+                            channelToRemove.detach(object : CompletionListener {
+                                override fun onSuccess() {
+                                    channels.remove(trackableId)
```
QuintinWillison, line 265: You're mutating `channels` from an Ably callback. Is that safe?

Also, if the detach fails (`onError(ErrorInfo)` path) then are we in an inconsistent state? i.e. the channel is now unsubscribed from messages and presence, yet remains attached.

QuintinWillison (changes requested):

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: The `channels` was already being mutated from the Ably callback (when leaving presence succeeded) and I agree that's not the safest thing to do. Because this change doesn't make it worse than it was before I won't fix that here but I've created an issue for that https://github.com/ably/ably-asset-tracking-android/issues/440.

About the channel being in an inconsistent state when detach fails, it's true. However, at this point, we've left the presence so the channel is already partially closed and won't return to the fully working state if the detach method fails. Also is there anything to unsubscribe from when the channel is detached? I would assume that after detaching all the listeners are already cleared :thinking:  I think it's also a thing that we should consider and make better so I've also mentioned it in the same issue from the previous point. :wink:

QuintinWillison (approved):

PR #445, opened 2021-09-22
Change AppPreferences to the singleton pattern

KacperKluka:
Because we don't want to complicate the app by introducing a DI tool, I've changed the `AppPreferences` to a singleton manually.

### DIFF
--- a/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AppPreferences.kt
+++ b/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AppPreferences.kt
@@ -4,9 +4,21 @@ import android.content.Context
 import androidx.preference.PreferenceManager
 import com.ably.tracking.Accuracy
 
-// TODO - refactor this class to the singleton pattern
-// https://github.com/ably/ably-asset-tracking-android/issues/234
-class AppPreferences(context: Context) {
+class AppPreferences private constructor(context: Context) {
+    companion object {
+        private var instance: AppPreferences? = null
+        fun getInstance(context: Context): AppPreferences {
+            if (instance == null) {
+                synchronized(this) {
+                    if (instance == null) {
+                        instance = AppPreferences(context.applicationContext)
+                    }
+                }
+            }
+            return instance!!
+        }
+    }
+
     private val preferences = PreferenceManager.getDefaultSharedPreferences(context)
     private val LOCATION_SOURCE_KEY = context.getString(R.string.preferences_location_source_key)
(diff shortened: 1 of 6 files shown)


AndyNicks (approved):

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AppPreferences.kt
```diff
@@ -4,9 +4,21 @@ import android.content.Context
...
 import com.ably.tracking.Accuracy
 
-// TODO - refactor this class to the singleton pattern
-// https://github.com/ably/ably-asset-tracking-android/issues/234
-class AppPreferences(context: Context) {
+class AppPreferences private constructor(context: Context) {
+    companion object {
+        private var instance: AppPreferences? = null
+        fun getInstance(context: Context): AppPreferences {
+            if (instance == null) {
```
QuintinWillison, line 11: I would have expected this check to be _within_ the `synchronized` scope. 🤔

QuintinWillison (changes requested):

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AppPreferences.kt
KacperKluka: We have 2 null checks here: one inside and one outside the `synchronized` scope. 
```kotlin
if(instance == null) {
    synchronized {
        if(instance == null) {
            instance = AppPreferences(context)
        }
    }
}
return instance
```
Thanks to this we only run the `synchronized` code when we need to. After the instance will be created the `synchronized` won't be called again saving us some performance :wink: If we only had one null check inside the `synchronized` scope it would look like this
```kotlin
synchronized {
    if(instance == null) {
        instance = AppPreferences(context)
    }
}
return instance
```
and it would call the `synchronized` block each time we would call this method, which is unnecessary.

QuintinWillison: I see. That makes sense. Thanks for explaining.

That pattern has never particularly occurred to me and/or I've not noticed it used elsewhere. I guess it's an alternative to using something like `volatile` on the field.

In my defence there could be an argument that this is a premature optimisation that's slightly counter to the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), but my feelings aren't too strong on that as this does look harmless enough.

QuintinWillison (approved):

PR #452, opened 2021-09-29
Add hepler extension function for showing toasts

KacperKluka:
I've spotted that we had a lot of duplicated `showToast()` methods across the publisher example app so I've extracted the code to helper extension functions, making them available to everything that is or has a `Context` object.

### DIFF
--- a/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AddTrackableActivity.kt
+++ b/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AddTrackableActivity.kt
@@ -7,7 +7,6 @@ import android.os.Bundle
 import android.view.View
 import android.view.inputmethod.EditorInfo
 import android.widget.ArrayAdapter
-import android.widget.Toast
 import androidx.annotation.RequiresPermission
 import androidx.core.content.ContextCompat
 import androidx.core.widget.addTextChangedListener
@@ -86,7 +85,7 @@ class AddTrackableActivity : PublisherServiceActivity() {
                     addTrackable(trackableId)
                 }
             } else {
-                showToast("Insert tracking ID")
+                showLongToast("Insert tracking ID")
             }
         }
     }
@@ -140,7 +139,7 @@ class AddTrackableActivity : PublisherServiceActivity() {
                     )
                     finish()
                 } catch (exception: Exception) {
-                    showToast("Error when adding the trackable")
+                    showLongToast("Error when adding the trackable")
                     hideLoading()
                 }
             }
@@ -159,7 +158,7 @@ class AddTrackableActivity : PublisherServiceActivity() {
             this,
             appPreferences.getS3File(),
             onHistoryDataDownloaded = { onHistoryDataDownloaded(it) },
-            onUninitialized = { showToast("S3 not initialized - cannot download history data") }
+            onUninitialized = { showLongToast("S3 not initialized - cannot download history data") }
         )
     }
 
@@ -173,10 +172,6 @@ class AddTrackableActivity : PublisherServiceActivity() {
         progressIndicator.visibility = View.GONE
         addTrackableButton.isEnabled = true
         trackableIdEditText.isEnabled = true
-    }
-
-    private fun showToast(message: String) {
-        Toast.makeText(this, message, Toast.LENGTH_SHORT).show()
     }
 
     private fun getTrackableId(): String = trackableIdEditText.text.toString().trim()
--- a/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
+++ b/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
@@ -3,10 +3,19 @@ package com.ably.tracking.example.publisher
 import android.content.Context
 import android.view.View
 import android.view.inputmethod.InputMethodManager
+import android.widget.Toast
 
 fun Context.hideKeyboard(view: View) {
     (getSystemService(Context.INPUT_METHOD_SERVICE) as? InputMethodManager)?.hideSoftInputFromWindow(
         view.windowToken,
         0
     )
+}
+
+fun Context.showShortToast(message: String) {
+    Toast.makeText(this, message, Toast.LENGTH_SHORT).show()
+}
+
+fun Context.showLongToast(message: String) {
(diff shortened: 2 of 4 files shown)


publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
```diff
@@ -3,10 +3,19 @@ package com.ably.tracking.example.publisher
...
         view.windowToken,
         0
     )
 }
+
+fun Context.showShortToast(message: String) {
+    Toast.makeText(this, message, Toast.LENGTH_SHORT).show()
+}
+
+fun Context.showLongToast(message: String) {
```
ben-xD, line 19: i dont wanna add random work but these 2 methods could be `showToast()` with optional parameters (e.g. duration) with a default (`LENGTH_SHORT`). This would make it more consistent with the existing makeText API.

ben-xD (approved):

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
KacperKluka: Right, I've also considered the approach you're talking about but in the end, I've chosen the explicit function name way :wink: This might be simply a matter of personal preferences but I'd love to hear @QuintinWillison and @ikbalkaya opinion on that :grin:

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AddTrackableActivity.kt
```diff
@@ -86,7 +85,7 @@ class AddTrackableActivity : PublisherServiceActivity() {
                     addTrackable(trackableId)
                 }
             } else {
-                showToast("Insert tracking ID")
+                showLongToast("Insert tracking ID")
```
ikbalkaya, line 88: Am I right to think that string is not localised here as it's for an example  
🤔

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
```diff
@@ -3,10 +3,19 @@ package com.ably.tracking.example.publisher
 import android.content.Context
 import android.view.View
 import android.view.inputmethod.InputMethodManager
+import android.widget.Toast
 
 fun Context.hideKeyboard(view: View) {
     (getSystemService(Context.INPUT_METHOD_SERVICE) as? InputMethodManager)?.hideSoftInputFromWindow(
         view.windowToken,
         0
```
ikbalkaya, line 11: I don't want to create extra work but I'd definitely benefit from naming of these arguments here, at least 0 here does give me very little clue about what it is

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
ikbalkaya: I'd say Kotlin style goes better with different naming (eg. list.isEmpty, list.isNotEmpty). Additional parameters gives flexibility but I think in this case it's better to stay with separate naming. It makes it easier for clients

ikbalkaya (changes requested):
LGTM except for a minor readability improvement request

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
ben-xD: Okay I don't mind, please resolve this conversation once you're satisfied with everyones comments

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AddTrackableActivity.kt
KacperKluka: That's right, however, I think that we should stick to best practices and move all messages to the `strings.xml` file and make it possible to easily localize the app. It may be useful for someone who would like to use it as an example in some other language :wink: I've created an issue for that https://github.com/ably/ably-asset-tracking-android/issues/455 thanks for spotting this :bow:

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
KacperKluka: I agree with you, that magic number isn't saying anything :sweat_smile: However, that feels a bit outside of this PR's scope so I've put it in another PR https://github.com/ably/ably-asset-tracking-android/pull/456 :wink:

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/AddTrackableActivity.kt
ikbalkaya: Great. I'll resolve this conversation than

ikbalkaya (approved):
LGTM

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/Extensions.kt
QuintinWillison: I've not got a massively strong opinion either way, but if I think the following considerations probably swing this towards keeping with the longer names for these extension methods, not adding a `duration` parameter:

- these extensions are internal to this example app codebase - so consistency with the underlying APIs they use is not such a major concern as they serve a domain with a narrow scope
- these extensions are designed to be used 'quickly' from this example app codebase - so it could be argued that having another parameter to think about, especially when it's 'borrowed' from another API, adds cognitive load to maintainers
- the longer names do exactly what they say on the tin - whereas a default parameter means that at the call sites there is no indication that the toast is 'short' by default - more cognitive load and opportunity for future maintainers to accidentally spawn a short toast when they wanted a long one

Side-note: I had to look up [makeText](https://developer.android.com/reference/android/widget/Toast#makeText(android.content.Context,%20int,%20int)) to understand this conversation (next time, @ben-xD, it would be great if you could turn mentions of third-party / external foundational APIs into links to make it easier for others participating in the conversation).

KacperKluka: Thank you all for your input :bow: Given [Ben's permission](https://github.com/ably/ably-asset-tracking-android/pull/452#discussion_r719092835) I'm resolving this conversation :wink:

PR #456, opened 2021-09-30
Use named parameters for hideKeyboard extension

KacperKluka:
To improve code readability I've extracted the magic number `0` to a `NO_FLAGS` constant. I've tried to use Kotlin's named parameters but it turns out that you can't use those if the function is a non-Kotlin one which is a bit disappointing.


ikbalkaya (approved):
LGTM

PR #461, opened 2021-10-06
Rename workflow job names to match the file and workflow name

KacperKluka:
I've noticed that some of the workflows have wrong names (e.g. the `emulate` workflow's job was named `check`) so I've changed those names to match the workflow filenames.

### DIFF
--- a/.github/workflows/assemble.yml
+++ b/.github/workflows/assemble.yml
@@ -5,7 +5,7 @@ on:
       - main
 
 jobs:
-  check:
+  assemble:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,7 +5,7 @@ on:
       - main
 
 jobs:
-  build:
+  docs:
     runs-on: ubuntu-latest
     permissions:
       deployments: write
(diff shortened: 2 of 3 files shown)


QuintinWillison (commented):
You have indeed spotted something, but it's something that I would prefer not to change!

We have some documentation in [this internal Confluence page](https://ably.atlassian.net/wiki/spaces/SDK/pages/1301413924/GitHub+Standards+and+Best+Practices#Job-Names) (that you don't have access to, @KacperKluka) but I intend to move that out somewhere public at some point soon. To quote from that section:

> Job Names

> This one’s a work in progress at the moment. If in doubt use `check`. We’re also using `build` in some places (e.g. Android Asset Tracking’s [docs workflow](https://github.com/ably/ably-asset-tracking-android/blob/main/.github/workflows/docs.yml)), where the job’s purpose is generating built output that’s intended to be accessible beyond the lifespan of the workflow.

> It’s important because when we create branch protection rules (typically for the default branch, `main`) we want to tick the box to “Require status checks to pass before merging”, which then presents us a list of job names under “Status checks found in the last week for this repository”.

Here's a screenshot of this part of the settings for the `main` branch protection rule in this repository:

Screenshot 2021-10-06 at 15 30 20

**TL;DR:** The names of jobs are relied upon downstream, so I'm keeping them to a small set intentionally.

PR #486, opened 2021-10-21
Extract hardcoded strings from example apps to string resources

KacperKluka:
We should keep all potentially translatable strings in the `strings.xml` resources file to make it easier to localize the app. I've changed the toast helper methods to accept string resource ID to prevent using hardcoded strings in the future.

### DIFF
--- a/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
+++ b/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
@@ -62,7 +62,7 @@ class MainActivity : PublisherServiceActivity() {
                             publisherService.publisher?.stop()
                             publisherService.publisher = null
                         } catch (e: Exception) {
-                            showLongToast(R.string.error_stop_publisher)
+                            showLongToast(R.string.error_stopping_publisher_failed)
                         }
                     } else {
                         showTrackablesList()
--- a/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
+++ b/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
@@ -118,7 +118,7 @@ class MainActivity : PublisherServiceActivity() {
     }
 
     private fun updateLocationSourceMethodInfo() {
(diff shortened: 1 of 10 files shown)


publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
```diff
@@ -62,7 +62,7 @@ class MainActivity : PublisherServiceActivity() {
                             publisherService.publisher?.stop()
                             publisherService.publisher = null
                         } catch (e: Exception) {
-                            showLongToast("Stopping publisher error")
+                            showLongToast(R.string.error_stop_publisher)
```
QuintinWillison, line 65: In this particular case (and a few others where the toast is shown from within a `catch` block) I would have expected it to be useful to display to the user at least the message from `e`. Have you considered that some of these strings may need to take an argument to produce a dynamically formatted output?

QuintinWillison (changes requested):
I think there are some consistency issues with the names you have defined for string resources not matching the textual values they have, or necessarily the class of message they display to the user. 

In terms of consistency, a good example would be `error_stop_publisher` returning value "Stopping publisher error" - whereas `error_publisher_service_not_started` returning a more logically 1:1 mapped value "Publisher service not started".

In terms of message class, in some cases it's a straight up "I failed, sorry" whereas in others it's more like "Things failed, but you can do something to fix this, and this is what it is". I think it might be helpful to differentiate these somehow.

publishing-example-app/src/main/java/com/ably/tracking/example/publisher/MainActivity.kt
KacperKluka: Toasts are not a great tool for showing longer messages since they disappear after a while and the message is gone. Also, the area and text they show are rather small and cramped. Therefore, I would prefer to not show information from exceptions in them. A better place for such errors would be the console log or (if we want users to see them) an alert dialogue. :wink:

QuintinWillison: https://github.com/ably/ably-asset-tracking-android/issues/500

QuintinWillison (approved):

PR #493, opened 2021-10-29
Send raw locations from the Publisher

KacperKluka:
Enabled sending of the raw locations on the same channel on which the enhanced locations are being sent. The raw locations have all the mechanisms that the enhanced ones have, especially resolution and `skippedLocations`. As the raw locations shouldn't be normally published, there is a new option on the builder that enables the raw locations. The commentary on this method states clearly that this should not be enabled in the production environment. Additionally, information about whether the raw locations are enabled is added to the `PresenceData` of the publisher.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -9,7 +9,7 @@ import com.ably.tracking.common.logging.i
 import com.ably.tracking.common.logging.v
 import com.ably.tracking.common.logging.w
 import com.ably.tracking.common.message.getEnhancedLocationUpdate
-import com.ably.tracking.common.message.toJson
+import com.ably.tracking.common.message.toMessageJson
 import com.ably.tracking.common.message.toMessage
 import com.ably.tracking.connection.ConnectionConfiguration
 import com.ably.tracking.logging.LogHandler
@@ -326,7 +326,7 @@ constructor(
     ) {
         val trackableChannel = channels[trackableId]
         if (trackableChannel != null) {
-            val locationUpdateJson = locationUpdate.toJson(gson)
+            val locationUpdateJson = locationUpdate.toMessageJson(gson)
             logHandler?.d("sendEnhancedLocationMessage: publishing: $locationUpdateJson")
             try {
                 trackableChannel.publish(
@@ -358,7 +358,7 @@ constructor(
     ) {
         val trackableChannel = channels[trackableId]
         if (trackableChannel != null) {
-            val locationUpdateJson = locationUpdate.toJson(gson)
+            val locationUpdateJson = locationUpdate.toMessageJson(gson)
             logHandler?.d("sendRawLocationMessage: publishing: $locationUpdateJson")
             try {
                 trackableChannel.publish(
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -24,14 +24,14 @@ import io.ably.lib.types.ChannelOptions
 import io.ably.lib.types.ErrorInfo
 import io.ably.lib.types.Message
 import io.ably.lib.util.Log
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.supervisorScope
-import kotlin.coroutines.resume
-import kotlin.coroutines.resumeWithException
-import kotlin.coroutines.suspendCoroutine
 
 /**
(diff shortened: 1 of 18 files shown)


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -86,6 +86,21 @@ interface Ably {
         callback: (Result<Unit>) -> Unit
     )
 
+    /**
+     * Sends a raw location update to the channel.
+     * Should be called only when there's an existing channel for the [trackableId].
+     * If a channel for the [trackableId] doesn't exist then it just calls [callback] with success.
```
QuintinWillison, line 92: Do you rely on this behaviour? This feels surprising to me, at least when examining this API in isolation. I generally feel it's better to prefer failure (a.k.a. fail early) rather than 'hide' unexpected cases. 🤔

publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
```diff
@@ -200,6 +200,16 @@ interface Publisher {
...
 
+        /**
+         * EXPERIMENTAL API
+         * **OPTIONAL** Enables sending of raw location updates. This should only be enabled for diagnostics.
+         * In the production environment this should be always disabled.
+         * By default this is disabled.
+         *
+         * @return A new instance of the builder with this property changed.
+         */
+        fun enableRawLocations(): Builder
```
QuintinWillison, line 211: I would prefer to remove the `enable` prefix and accept a boolean here.

QuintinWillison (changes requested):
Overall looks very good, thanks. I guess I'm surprised at how much code you needed to add to gain this functionality - it clearly cuts quite deep.

publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/493/commits/cf983ce2676aceaac726873fa72bef6f7d7a5ea5

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: I don't think that we use this behaviour but since it is how all methods in this interface work I would prefer to create a separate issue to change this instead of changing it in this PR :wink: WDYT? Should I create the issue?

QuintinWillison: Yes. Happy to see an issue created for this, under [code-quality](https://github.com/ably/ably-asset-tracking-android/labels/code-quality).

QuintinWillison (approved):

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: Issue created: https://github.com/ably/ably-asset-tracking-android/issues/496

common/src/main/java/com/ably/tracking/common/message/MessageMappers.kt
```diff
@@ -41,6 +42,14 @@ fun Accuracy.toMessage(): AccuracyMessage = when (this) {
     Accuracy.MAXIMUM -> AccuracyMessage.MAXIMUM
 }
 
+fun LocationUpdate.toJson(gson: Gson): String =
```
ikbalkaya, line 45: To me, this name is a bit misleading. It implies that the `LocationUpdate` is going to be serialised but it's a `LocationUpdateMessage` serialization. Is that how this object will be serialised all the time? If not, could you rename this to something like `toMessageJson` ? Thank you 🙏

common/src/main/java/com/ably/tracking/common/message/MessageMappers.kt
```diff
@@ -41,6 +42,14 @@ fun Accuracy.toMessage(): AccuracyMessage = when (this) {
...
+    gson.toJson(
+        LocationUpdateMessage(
+            location.toMessage(),
+            skippedLocations.map { it.toMessage() },
+        )
+    )
+
 fun EnhancedLocationUpdate.toJson(gson: Gson): String =
```
ikbalkaya, line 53: Similar as above

ikbalkaya (changes requested):
I made a request which should improve readability slightly

common/src/main/java/com/ably/tracking/common/message/MessageMappers.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/493/commits/5ebd47d7cb347705074d1d995a72d08ead22e26a

common/src/main/java/com/ably/tracking/common/message/MessageMappers.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/493/commits/5ebd47d7cb347705074d1d995a72d08ead22e26a

ikbalkaya (approved):
Looks good 😎

PR #495, opened 2021-11-02
Add raw locations to the Subscriber SDK

KacperKluka:
I've enabled listening for raw locations in the Subscriber SDK. I've also added this to the Java facade of Subscriber. Additionally, I've added an integration test that checks if there is a raw data flow between publisher and subscriber.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -9,6 +9,7 @@ import com.ably.tracking.common.logging.i
 import com.ably.tracking.common.logging.v
 import com.ably.tracking.common.logging.w
 import com.ably.tracking.common.message.getEnhancedLocationUpdate
+import com.ably.tracking.common.message.getRawLocationUpdate
 import com.ably.tracking.common.message.toJson
 import com.ably.tracking.common.message.toMessage
 import com.ably.tracking.connection.ConnectionConfiguration
@@ -111,6 +112,17 @@ interface Ably {
      * @throws ConnectionException if something goes wrong.
      */
     fun subscribeForEnhancedEvents(trackableId: String, listener: (LocationUpdate) -> Unit)
+
+    /**
+     * Adds a listener for the raw location updates that are received from the channel.
+     * If a channel for the [trackableId] doesn't exist then nothing happens.
+     *
+     * @param trackableId The ID of the trackable channel.
+     * @param listener The function that will be called each time a raw location update event is received.
+     *
+     * @throws ConnectionException if something goes wrong.
+     */
+    fun subscribeForRawEvents(trackableId: String, listener: (LocationUpdate) -> Unit)
 
     /**
      * Joins the presence of the channel for the given [trackableId] and add it to the connected channels.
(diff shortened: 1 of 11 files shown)


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -112,6 +113,17 @@ interface Ably {
...
+    /**
+     * Adds a listener for the raw location updates that are received from the channel.
+     * If a channel for the [trackableId] doesn't exist then nothing happens.
+     *
+     * @param trackableId The ID of the trackable channel.
+     * @param listener The function that will be called each time a raw location update event is received.
+     *
+     * @throws ConnectionException if something goes wrong.
+     */
+    fun subscribeForRawEvents(trackableId: String, listener: (LocationUpdate) -> Unit)
```
QuintinWillison, line 125: I recognise that I'm looking at an internal API here, however I think we should highlight in as many places as possible that these raw location updates will only be received if opted in to (overriding builder default) at publisher end.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Subscriber.kt
```diff
@@ -104,6 +111,17 @@ interface Subscriber {
...
+        /**
+         * EXPERIMENTAL API
+         * **OPTIONAL** Enables subscribing for raw location updates. This should only be enabled for diagnostics.
+         * In the production environment this should be always disabled.
+         * By default this is disabled.
+         *
+         * @param enabled Whether subscribing for raw location updates is enabled.
+         * @return A new instance of the builder with this property changed.
+         */
+        fun rawLocations(enabled: Boolean): Builder
```
QuintinWillison, line 123: I don't feel we need this API at all. If the application developer has not added a consumer to the `rawLocations` shared flow then they are clearly already signalling that they do not care about this data. Same goes for the `rawLocations` method you've added to the subscriber builder in this pull request.

QuintinWillison (changes requested):
I think this is more complicated than it needs be, with too much new API added.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Subscriber.kt
KacperKluka: This `rawLocations()` method was added to prevent us from subscribing to the raw messages when it's unnecessary (see [here](https://github.com/ably/ably-asset-tracking-android/pull/495/files#diff-b61a79b87abd92c047c791b03922a3c91ad39f1e9c3c32c30c32ee55847a47cbR141-R143)) :thinking: If we remove it then we will always have to subscribe for raw locations and push them to the `SharedFlow` even if the user won't be using them. If that's fine then we can safely remove this from the subscriber's builder :wink:

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/495/commits/82cc783b8da37d78b0ccf07eaca8e5b17fec0747

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Subscriber.kt
QuintinWillison: The service will be sending all data regardless, given subscription filtering is a client-side only thing at the subscriber. As discussed in Slack, I think we should remove this API. It can always be introduced in future if there's some perceived benefit - but, for now, the simple availability of the shared flow is ample.

KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/495/commits/f501664c42451cefa819a9aadcd3cb02b97d87d1

ikbalkaya (approved):
LGTM

QuintinWillison (approved):

PR #501, opened 2021-11-04
Send raw locations from the publisher example app

KacperKluka:
I've added a switch in the settings that enables or disables raw locations. It is placed under a new `Debug` section to indicate that this is not meant to be used in the production environment. By default raw locations are disabled.
Screenshot from 2021-11-04 11-15-01

### DIFF
--- a/publishing-example-app/src/main/res/values/strings.xml
+++ b/publishing-example-app/src/main/res/values/strings.xml
@@ -50,6 +50,9 @@
   <string name="preferences_resolution_minimum_displacement_label">Minimum displacement</string>
   <string name="preferences_update_resolution_minimum_displacement_message">Minimum Displacement [m]</string>
   <string name="preferences_update_resolution_minimum_displacement_title">Set Minimum Displacement</string>
+  <string name="preferences_debug_category_title">Debug</string>
+  <string name="preferences_send_raw_locations_key">send_raw_locations</string>
+  <string name="preferences_send_raw_locations_label">Send raw location updates</string>
   <string name="trackable_list_empty_state_header">No assets</string>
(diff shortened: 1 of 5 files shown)


AndyNicks (approved):

publishing-example-app/src/main/res/values/strings.xml
```diff
@@ -50,6 +50,9 @@
   <string name="preferences_resolution_minimum_displacement_label">Minimum displacement</string>
   <string name="preferences_update_resolution_minimum_displacement_message">Minimum Displacement [m]</string>
   <string name="preferences_update_resolution_minimum_displacement_title">Set Minimum Displacement</string>
+  <string name="preferences_debug_category_title">Debug</string>
+  <string name="preferences_send_raw_locations_key">send_raw_locations</string>
```
ikbalkaya, line 54: Just a comment on this and does not require any change but for later keys might be stored in another static (not localised) XML resource

ikbalkaya (approved):
Looks good 👍 I made a comment on the choice of key location. It's not a big deal for now but ideally those sorts of non-localisable constants would be better in a non-localisable resource file.

publishing-example-app/src/main/res/values/strings.xml
KacperKluka: Good idea, I've created an issue for that :wink: https://github.com/ably/ably-asset-tracking-android/issues/502

QuintinWillison: I am resolving this conversation on @ikbalkaya's behalf.

QuintinWillison (approved):

Issue #450, opened 2021-09-28
Calling remove() should remove a trackable that isn't finished being added


KacperKluka:
When you call `add()` and then call `remove()` the given trackable should be removed no matter if it was already added or if it is still being added. 
Currently, we are only handling removing of an already added trackable. This means that if the `add()` method finishes then we can safely call `remove()` and it will do its job. Things look different if we call `remove()` before the `add()` method finishes. 
Because the `add()` (and `track()`) method consists of multiple async steps, the adding of a single trackable requires the handling of multiple events. The trackable is considered as being "added" after it completes all of them. Because of that, it is possible that the remove event will be handled before the final add event is handled, which will result in the trackable successfully being added and the remove not removing it but completing with a success.

We should think of some way that would allow us to detect that we want to remove a trackable that's being added and remove it without completing all of its usual steps.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1252) by [Unito](https://www.unito.io)

PR #506, opened 2021-11-05
Remove unfinished trackable when remove called

ikbalkaya:
This PR should fix https://github.com/ably/ably-asset-tracking-android/issues/450

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -328,6 +328,10 @@ constructor(
                                     event.handler(Result.failure(result.exceptionOrNull()!!))
                                 }
                             }
+                        } else if (state.duplicateTrackableGuard.isCurrentlyAddingTrackable(event.trackable)) {
+                            //This is the case where a trackable hasn't yet finished adding and the removal was
+                            // requested
+                            state.duplicateTrackableGuard.markForRemoval(event.trackable)
                         } else {
                             // notify with false to indicate that it was not removed
                             event.handler(Result.success(false))
@@ -716,7 +720,7 @@ constructor(
             val timeSinceLastSentLocation = currentLocation.timeFrom(lastSentLocation)
             val distanceFromLastSentLocation = currentLocation.distanceInMetersFrom(lastSentLocation)
             return distanceFromLastSentLocation >= resolution.minimumDisplacement ||
-                timeSinceLastSentLocation >= resolution.desiredInterval
+                    timeSinceLastSentLocation >= resolution.desiredInterval
         } else {
             true
         }
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -18,6 +18,7 @@ import com.ably.tracking.common.PresenceData
 import com.ably.tracking.common.createSingleThreadDispatcher
 import com.ably.tracking.common.logging.w
 import com.ably.tracking.logging.LogHandler
+import com.ably.tracking.publisher.guards.DuplicateTrackableGuard
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.channels.Channel
@@ -330,7 +331,7 @@ constructor(
                             }
                         } else if (state.duplicateTrackableGuard.isCurrentlyAddingTrackable(event.trackable)) {
                             //This is the case where a trackable hasn't yet finished adding and the removal was
-                            // requested
+                            // requested. We mark that trackable for removal so that it will not
                             state.duplicateTrackableGuard.markForRemoval(event.trackable)
                         } else {
                             // notify with false to indicate that it was not removed
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -19,6 +19,7 @@ import com.ably.tracking.common.createSingleThreadDispatcher
 import com.ably.tracking.common.logging.w
 import com.ably.tracking.logging.LogHandler
 import com.ably.tracking.publisher.guards.DuplicateTrackableGuard
+import com.ably.tracking.publisher.guards.TrackableRemovalGuard
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.channels.Channel
@@ -239,6 +240,7 @@ constructor(
                         val failureResult = Result.failure<AddTrackableResult>(event.exception)
                         event.handler(failureResult)
                         state.duplicateTrackableGuard.finishAddingTrackable(event.trackable, failureResult)
+                        state.trackableRemovalGuard.removeMarked(event.trackable)
                     }
                     is PresenceMessageEvent -> {
                         when (event.presenceMessage.action) {
@@ -270,22 +272,32 @@ constructor(
                         }
                     }
                     is ConnectionForTrackableCreatedEvent -> {
-                        ably.subscribeForPresenceMessages(
-                            trackableId = event.trackable.id,
-                            listener = { enqueue(PresenceMessageEvent(event.trackable, it)) },
-                            callback = { result ->
-                                try {
-                                    result.getOrThrow()
-                                    request(ConnectionForTrackableReadyEvent(event.trackable, event.handler))
-                                } catch (exception: ConnectionException) {
-                                    ably.disconnect(event.trackable.id, state.presenceData) {
-                                        request(AddTrackableFailedEvent(event.trackable, event.handler, exception))
+                        //only if it wasn't marked for removal
+                        if (state.trackableRemovalGuard.markedForRemoval(event.trackable)) {
+                            state.trackableRemovalGuard.removeMarked(event.trackable)
+                        } else {
+                            ably.subscribeForPresenceMessages(
+                                trackableId = event.trackable.id,
+                                listener = { enqueue(PresenceMessageEvent(event.trackable, it)) },
+                                callback = { result ->
+                                    try {
+                                        result.getOrThrow()
+                                        request(ConnectionForTrackableReadyEvent(event.trackable, event.handler))
+                                    } catch (exception: ConnectionException) {
+                                        ably.disconnect(event.trackable.id, state.presenceData) {
+                                            request(AddTrackableFailedEvent(event.trackable, event.handler, exception))
+                                        }
                                     }
                                 }
-                            }
-                        )
+                            )
+                        }
+
                     }
                     is ConnectionForTrackableReadyEvent -> {
+                        if (state.trackableRemovalGuard.markedForRemoval(event.trackable)){
+                            request(AddTrackableFailedEvent(event.trackable, event.handler, exception = Exception
+                                ("Trackable was marked for removal")))
+                        }
                         ably.subscribeForChannelStateChange(event.trackable.id) {
                             enqueue(ChannelConnectionStateChangeEvent(it, event.trackable.id))
                         }
@@ -332,7 +344,7 @@ constructor(
                         } else if (state.duplicateTrackableGuard.isCurrentlyAddingTrackable(event.trackable)) {
                             //This is the case where a trackable hasn't yet finished adding and the removal was
                             // requested. We mark that trackable for removal so that it will not
-                            state.duplicateTrackableGuard.markForRemoval(event.trackable)
+                            state.trackableRemovalGuard.markForRemoval(event.trackable)
                         } else {
                             // notify with false to indicate that it was not removed
                             event.handler(Result.success(false))
@@ -829,6 +841,8 @@ constructor(
         val rawLocationsPublishingState: LocationsPublishingState<RawLocationChangedEvent> = LocationsPublishingState()
             get() = if (isDisposed) throw PublisherStateDisposedException() else field
         val duplicateTrackableGuard: DuplicateTrackableGuard = DuplicateTrackableGuard()
+            get() = if (isDisposed) throw PublisherStateDisposedException() else field
+        val trackableRemovalGuard: TrackableRemovalGuard = TrackableRemovalGuard()
             get() = if (isDisposed) throw PublisherStateDisposedException() else field
 
         fun dispose() {
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/TrackableRemovalGuard.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/TrackableRemovalGuard.kt
@@ -1,0 +1,18 @@
+package com.ably.tracking.publisher.guards
+
+import com.ably.tracking.publisher.Trackable
+
+class TrackableRemovalGuard {
+
+    /**
+     * A set of trackables that were marked for removal. This should be used to store / retrieve trackables
+     * that did not finished adding.
+     */
+    private val trackablesMarkedForRemoval: MutableSet<Trackable> = mutableSetOf()
+
+    fun markForRemoval(trackable: Trackable) {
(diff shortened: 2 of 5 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -825,6 +848,8 @@ constructor(
             get() = if (isDisposed) throw PublisherStateDisposedException() else field
         val duplicateTrackableGuard: DuplicateTrackableGuard = DuplicateTrackableGuard()
             get() = if (isDisposed) throw PublisherStateDisposedException() else field
+        val trackableRemovalGuard: TrackableRemovalGuard = TrackableRemovalGuard()
+            get() = if (isDisposed) throw PublisherStateDisposedException() else field
 
         fun dispose() {
```
KacperKluka, line 854: You should clear the state of the newly added `trackableRemovalGuard` in the `dispose()` method :wink: To do that you will have to add a method like `clearAll()` to the guard (see the `DuplicateTrackableGuard` for reference).

We should probably also clear a state of this guard when a trackable is removed just as we de with the `duplicateTrackableGuard` (see its `clear()` method) :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -328,6 +347,10 @@ constructor(
                                     event.handler(Result.failure(result.exceptionOrNull()!!))
                                 }
                             }
+                        } else if (state.duplicateTrackableGuard.isCurrentlyAddingTrackable(event.trackable)) {
+                            // This is the case where a trackable hasn't yet finished adding and the removal was
+                            // requested. We mark that trackable for removal so that it will not
```
KacperKluka, line 352: I'm not sure if we need to add this comment, the `isCurrentlyAddingTrackable` in the `if` statement seems quite self-explanatory to me :thinking:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -328,6 +347,10 @@ constructor(
...
+                            state.trackableRemovalGuard.markForRemoval(event.trackable)
```
KacperKluka, line 353: This is a bit too simple implementation due to the presence of `handler`s in the `Request` events :wink: Some operations like adding or removing have a callback `handler` that should be called when the operation is completed. If we won't call the handler then the method will never complete and the `DefaultPublisher`'s `remove()` method will never resume. You need to keep the reference to the `handler` and call it when the specified trackable is removed. Please look at the `DuplicateTrackableGuard` for reference :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -269,22 +272,36 @@ constructor(
...
+                                AddTrackableFailedEvent(
+                                    event.trackable,
+                                    event.handler,
+                                    exception = Exception("Trackable was marked for removal")
+                                )
+                            )
+                        }
                         ably.subscribeForChannelStateChange(event.trackable.id) {
                             enqueue(ChannelConnectionStateChangeEvent(it, event.trackable.id))
                         }
```
KacperKluka, line 307: This will put an `AddTrackableFailedEvent` to the queue and continue the usual flow of a successfully added trackable :thinking: We should stop executing this method when the trackable should be removed, so some sort of `continue` call should be present inside this `if` statement.

Additionally, here we are not actually removing the trackable, we are just saying that adding it has failed. But we still have an active channel in the `Ably` wrapper and other state set, as if the trackable was added. So I'd say we should also trigger the remove trackable flow here or manually clear the state that was set up to this point by the add operation :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -269,22 +272,36 @@ constructor(
...
-                            trackableId = event.trackable.id,
-                            listener = { enqueue(PresenceMessageEvent(event.trackable, it)) },
-                            callback = { result ->
-                                try {
-                                    result.getOrThrow()
-                                    request(ConnectionForTrackableReadyEvent(event.trackable, event.handler))
-                                } catch (exception: ConnectionException) {
-                                    ably.disconnect(event.trackable.id, state.presenceData) {
-                                        request(AddTrackableFailedEvent(event.trackable, event.handler, exception))
+                        // only if it wasn't marked for removal
```
KacperKluka, line 275: I think this comment is not necessary :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -269,22 +272,36 @@ constructor(
...
+                        if (state.trackableRemovalGuard.markedForRemoval(event.trackable)) {
+                            state.trackableRemovalGuard.removeMarked(event.trackable)
+                        } else {
```
KacperKluka, line 278: Like before, here we are simply making some state changes but we are not:
- notifying the add operation `handler`s that adding failed
- notifying the remove operation `handler` that the removing succeeded
- doing any actual removing trackable logic (so for example the Ably channel for the trackable remains active while it should be closed)

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -238,6 +240,7 @@ constructor(
                         val failureResult = Result.failure<AddTrackableResult>(event.exception)
                         event.handler(failureResult)
                         state.duplicateTrackableGuard.finishAddingTrackable(event.trackable, failureResult)
+                        state.trackableRemovalGuard.removeMarked(event.trackable)
```
KacperKluka, line 243: I'm wondering if failure of the adding operation means success of the removing operation :thinking: I think that we should mark that a trackable is removed only from the remove operation finish, not from adding a trackable fail :wink:

KacperKluka (changes requested):
The overall idea is good :muscle: but a few important details are missing :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/TrackableRemovalGuard.kt
```diff
@@ -0,0 +1,22 @@
+package com.ably.tracking.publisher.guards
+
+import com.ably.tracking.publisher.Trackable
+
+/**
+ * This guard class will keep track or trackables that has been requested for removal. The aim of this class is to
```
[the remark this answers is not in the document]
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/506/commits/9569b635edd7308ad46869946cb5d9c3a40218d9

publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/TrackableRemovalGuard.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * A set of trackables that were marked for removal. This should be used to store / retrieve trackables
+     * that did not finished adding.
+     */
+    private val trackables: MutableSet<Trackable> = mutableSetOf()
+
+    fun markForRemoval(trackable: Trackable) = trackables.add(trackable)
+
+    fun markedForRemoval(trackable: Trackable): Boolean = trackables.contains(trackable)
```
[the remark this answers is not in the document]
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/506/commits/31cb9c0c0067a6e0d1967bb13103e51af6968745

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: Hmm 🤔  The only state for this trackable is the set of trackables in contains. There is  already used in sequenceEventQueue method 
Change is here https://github.com/ably/ably-asset-tracking-android/pull/506/commits/7160c81ea07f37329f41def8db9ed6964f8070af

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: Well, that depends. I added it to make it a bit more obvious why this is being done as reading else if on a massively big code block could be tiring, but now you said it is obvious I will remove it.

ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/506/commits/36361c507de5f71ab0860709270df64f0dde3120

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: Could you ping me when you have some time please?

ikbalkaya: I will need some more explanation on this. Could you ping me when you have time please?

ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/506/commits/04feab851d1fc16f9b987f658f8ba3a366d99afd

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/506/commits/04feab851d1fc16f9b987f658f8ba3a366d99afd

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/506/commits/04feab851d1fc16f9b987f658f8ba3a366d99afd

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/506/commits/04feab851d1fc16f9b987f658f8ba3a366d99afd

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: Not sure I understand this one 🤔 But I suppose the change https://github.com/ably/ably-asset-tracking-android/pull/506/commits/04feab851d1fc16f9b987f658f8ba3a366d99afd should also cover that

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -268,23 +271,55 @@ constructor(
...
+                        if (event.result.isSuccess) {
+                            state.trackableRemovalGuard.removeMarked(event.trackable, Result.success(true))
+                        } else {
+                            state.trackableRemovalGuard.removeMarked(
+                                event.trackable,
+                                Result.failure(
+                                    event.result
+                                        .exceptionOrNull()!!
+                                )
+                            )
```
KacperKluka, line 284: Can't we have it in a single line like `Result.failure(event.result.exceptionOrNull()!!)`? :thinking:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -268,23 +271,55 @@ constructor(
...
+                        }
+                        event.handler(Result.failure(RemoveTrackableRequestedException()))
+                        state.duplicateTrackableGuard.finishAddingTrackable(
+                            event.trackable,
+                            Result.failure
+                            (RemoveTrackableRequestedException())
```
KacperKluka, line 290: I would prefer to have it in the same line: `Result.failure(RemoveTrackableRequestedException())` :wink:

publishing-sdk/src/test/java/com/ably/tracking/publisher/guards/TrackableRemovalGuardTest.kt
```diff
@@ -0,0 +1,139 @@
...
+        // given
+        val trackable = Trackable("sample")
+
+        trackableRemovalGuard.markForRemoval(trackable) {
+            // then
+            assertTrue(it.isFailure)
+        }
+        // when
+        trackableRemovalGuard.removeMarked(trackable, Result.failure(Exception("simple")))
+    }
```
KacperKluka, line 138: I'd prefer to modify those tests so the `given when then` are in the usual order :wink: You did something similar in [the two tests](https://github.com/ably/ably-asset-tracking-android/pull/506/files#diff-be03559c892f54bc1f3bab8f42eeb82aec098354696bbb0d203e66836dfdba89R72-R112) above (saving the result in a var and comparing it in the end), it should work here too :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -268,23 +271,55 @@ constructor(
...
-                                    request(ConnectionForTrackableReadyEvent(event.trackable, event.handler))
-                                } catch (exception: ConnectionException) {
-                                    ably.disconnect(event.trackable.id, state.presenceData) {
-                                        request(AddTrackableFailedEvent(event.trackable, event.handler, exception))
+                        if (state.trackableRemovalGuard.isMarkedForRemoval(event.trackable)) {
+                            // Leave Ably channel.
+                            ably.disconnect(event.trackable.id, state.presenceData) { result ->
+                                request(TrackableRemovalRequestedEvent(event.trackable, event.handler, result))
+                            }
+                        } else {
```
KacperKluka, line 299: Instead of this `if else` here we could use the `continue` keyword to skip further processing code for this event just as you did [here](https://github.com/ably/ably-asset-tracking-android/pull/506/files#diff-522c66079d446b893ff101d29587d46ee248760c7a3b0932611c7ba58fce8ed1R317-R322) for the `ConnectionForTrackableReadyEvent` :wink: I think it looks better as we do not introduce another indentation level to this already highly indented block of code :sweat_smile:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -296,9 +331,11 @@ constructor(
                         scope.launch { _trackables.emit(state.trackables) }
                         resolveResolution(event.trackable, state)
                         hooks.trackables?.onTrackableAdded(event.trackable)
-                        val trackableState = state.trackableStates[event.trackable.id] ?: TrackableState.Offline()
+                        val trackableState = state.trackableStates[event.trackable.id]
+                            ?: TrackableState.Offline()
```
KacperKluka, line 335: I wonder why your IDE is wrapping those lines :thinking: The maximum line width should be 120 and this and some other lines do not exceed the limit. Could you check if you have your `Hard wrap at` setting set to `120`? :wink: (Settings -> Editor -> Code Style)

KacperKluka (changes requested):
Looks much better but we can make it even better :grin:

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: I think my settings in IDE settings causes this. I am not doing anything special. But just wondering, wouldn't ktlint catch such code formatting issues?

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
ikbalkaya: That is a good idea 👍

KacperKluka (approved):
LGTM :+1:

PR #511, opened 2021-11-11
Refactor ably wrapper

ikbalkaya:
This pull request aims to improve Ably wrapper by

* Providing the dispatcher for the callback calls
* Reducing disconnect operation call to detaching only
* Added optional ? to a nullable reason parameter as subscribing app would crash after stopping.

For the disconnect operation, I realised that detaching would implicitly end presence from documentation (I added comment into code) so it should be safe to reduce the whole operation to detach. I left channel and presence unsubscription code there. 

After change this test fails on DefaultPublisherTest 
`should fail adding process when adding a trackable that is currently being added and it fails`

I was confused by what this test was trying to do so I couldn't fix it. I'll leave it for now till Monday


ikbalkaya:
I'm having issues with reformatting code to match ktlint style. It will complain about my indentation being 4 and not 5. But I will indent code to 5 and then it will complain otherwise 😩  Help please @KacperKluka

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -268,48 +279,51 @@ constructor(
...
+    /**
+     It should be adequate to only detach from the channel if we want to leave presence. Documentation says :
+     "A member who was present has now left the channel. This may be a result of an explicit request to leave or
+     implicitly when detaching from the channel. Alternatively, if a member’s connection is abruptly disconnected
+     and they do not resume their connection within a minute, Ably treats this as a leave event as the client
+     is no longer present"
+
+     https://docs.ably.io/realtime/presence/
+     * */
+    override fun disconnect(trackableId: String, callback: (Result<Unit>) -> Unit) {
```
KacperKluka, line 291: Do we know if the `detach` will send the `presenceData` when it leaves the presence? We are expecting to see the `presenceData` when leaving the presence and some of the Subscriber SDK logic relies on that :thinking:

publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
```diff
@@ -59,7 +60,7 @@ internal data class PublisherBuilder(
         }
         // All below fields are required and above code checks if they are nulls, so using !! should be safe from NPE
         return DefaultPublisher(
-            DefaultAbly(connectionConfiguration!!, logHandler),
+            DefaultAbly(connectionConfiguration!!, logHandler, singleThreadDispatcher),
```
KacperKluka, line 63: Here you are using the same single thread dispatcher that the core publisher's events queue uses. This is quite dangerous as any blocking code in the `DefaultAbly` can cause the whole `Publisher` to block and in the worst-case scenario end in a deadlock :thinking: Is there any particular reason why you are reusing this dispatcher instead of creating a new one?

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -167,13 +168,13 @@ interface Ably {
...
 
-class DefaultAbly
 /**
  * @throws ConnectionException if something goes wrong during Ably SDK initialization.
  */
-constructor(
+class DefaultAbly(
     connectionConfiguration: ConnectionConfiguration,
     private val logHandler: LogHandler?,
+    val callbackDispatcher: CoroutineDispatcher
```
KacperKluka, line 177: I think it should be a `private val` :wink:

KacperKluka (changes requested):
I'm not sure how wrapping the callbacks in a `scope.launch()` will make the code safer :thinking:
Given we are using a single-threaded dispatcher for the launch we have a guarantee that the `launch()` calls will be executed in the order in which they were called. However, why do we need this? :thinking: 
I noticed that in two places we are mutating the `channels` map from this scope so I guess this gives us some safety in terms of modifying this state, but then we are still accessing the state without any protection (all the `channels[trackableId]?.let {...}` calls).

Given that the only state that needs protection is the `channels` map maybe it would be better to use some multithread-safe data structure like [`ConcurrentHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) to handle everything related to thread safety? :thinking: WDYT? :wink:

ikbalkaya:
Superseded by https://github.com/ably/ably-asset-tracking-android/pull/513

Issue #440, opened 2021-09-17
Refactor the Ably wrapper to be more secure and error-proof


KacperKluka:
There are a couple of places in which the `Ably` wrapper can be improved e.g. currently, we are modifying the `channels` list in the callback from Ably so we're mutating the state outside of the "main thread". We should fix this and look for other dangerous things and fix them too.
Another thing to consider is how to assert a consistent state for the channel during the disconnect operation. What if only leaving presence succeeded but detaching failed? Should the channel then be considered as closed or do we want to introduce some new state for it?

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1226) by [Unito](https://www.unito.io)

KacperKluka:
We've decided that we will wait for https://github.com/ably/ably-asset-tracking-android/issues/522 to be resolved before working on this issue.

KacperKluka:
We need to wait for https://github.com/ably/ably-asset-tracking-android/issues/660 as well because the Ably wrapper is a part of the shared code used by both the publisher and subscriber.

davyskiba:
Investigation conclusions so far:
1. Ably java has an internal thread-safe channels map we could rely on instead of holding our own
2. once the channel enters the failed state only thing we can do with it is attach it, after which it starts operating normally

PR #513, opened 2021-11-16
Make ably wrapper safer

ikbalkaya:
This pull request aims to improve Ably wrapper (for a little bit) by
* Using a Collections.synchronisedMap to synchronise the underlying map  for channels in order to make it  thread safe
* Moving unsubscription operations to detach success callback to maintain a single point for success.
* Re-enter to presence if detach fails.

For now, reconnection happens with an empty listener. That is because it's solely to undo effect of presence.leave and registering a listener will highly complicate the situation. However it's still not completely safe as reconnection might also fail and thus the channel is in inconsistent state. I think we should aim to reduce the amount of state changing operations to 1 but it does not seem possible for now.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -293,11 +293,12 @@ constructor(
                                 callback(Result.failure(reason.toTrackingException()))
                                 // reenter but do not handle reenter results as it will make things more complicated
                                 // for now
-                                channelToRemove.presence.enter(Gson().toJson(presenceData),object :CompletionListener{
-                                    override fun onSuccess() { }
+                                channelToRemove.presence.enter(Gson().toJson(presenceData),
+                                    object : CompletionListener {
+                                        override fun onSuccess() {}
 
-                                    override fun onError(reason: ErrorInfo?) { }
-                                })
+                                        override fun onError(reason: ErrorInfo?) {}
+                                    })
                             }
                         })
                     }
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -9,8 +9,8 @@ import com.ably.tracking.common.logging.i
 import com.ably.tracking.common.logging.v
 import com.ably.tracking.common.logging.w
 import com.ably.tracking.common.message.getEnhancedLocationUpdate
-import com.ably.tracking.common.message.toMessageJson
 import com.ably.tracking.common.message.toMessage
+import com.ably.tracking.common.message.toMessageJson
 import com.ably.tracking.connection.ConnectionConfiguration
 import com.ably.tracking.logging.LogHandler
 import com.google.gson.Gson
@@ -24,15 +24,16 @@ import io.ably.lib.types.ChannelOptions
 import io.ably.lib.types.ErrorInfo
 import io.ably.lib.types.Message
 import io.ably.lib.util.Log
-import kotlin.coroutines.resume
-import kotlin.coroutines.resumeWithException
-import kotlin.coroutines.suspendCoroutine
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.supervisorScope
-import java.util.concurrent.ConcurrentHashMap
+import java.util.Collections
+import kotlin.collections.HashMap
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
 
 /**
  * Wrapper for the [AblyRealtime] that's used to interact with the Ably SDK.
@@ -178,7 +179,7 @@ constructor(
 ) : Ably {
     private val gson = Gson()
     private val ably: AblyRealtime
-    private val channels: ConcurrentHashMap<String, Channel> = ConcurrentHashMap()
+    private val channels = Collections.synchronizedMap(HashMap<String,Channel>())
     private val scope = CoroutineScope(Dispatchers.IO + SupervisorJob())
 
     init {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -177,7 +178,7 @@ constructor(
 ) : Ably {
     private val gson = Gson()
     private val ably: AblyRealtime
-    private val channels: MutableMap<String, Channel> = mutableMapOf()
+    private val channels: ConcurrentHashMap<String, Channel> = ConcurrentHashMap()
```
KacperKluka, line 181: I know that I've suggested using the `ConcurrentHashMap()` but after looking into [the docs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) and running a small test I'm starting to think that it's not the right tool for us :disappointed: 

In the class summary in the docs you can find that it is:
```
A hash table supporting full concurrency of retrievals and high expected concurrency for updates.
[...]
Retrieval operations (including get) generally do not block, so may overlap with update operations 
(including put and remove). Retrievals reflect the results of the most recently completed update 
operations holding upon their onset. (More formally, an update operation for a given key bears 
a happens-before relation with any (non-null) retrieval for that key reporting the updated value.)
```

I've checked and the `map.get()` and `map.put()` operations doesn't seem to be thread secure. In my test I've run this line of code `map["test_counter"] = map["test_counter"] + 1` in a multi-threaded fashion and there were races happening and the result was unpredictable. When I've replaced it with `map.compute("test_counter") { key, value -> value + 1 }` then there were no races and everything was looking good.

Given the above, I wonder if we will need to add some safety mechanisms by ourselves. Our main goal is to provide synchronized access to the `channels` map. We would like to both access and modify it in a secure way. The `ConcurrentHashMap` doesn't seem to serve this purpose. I'm a bit afraid that our only way would be to use `mutex`es or some other locks but we need to be careful with them. We probably don't want to block the whole `Ably` on a single operation but rather do this on a trackable basis :thinking: Let me know what you think about this :wink:

KacperKluka (changes requested):
We're on a good way but we need to make a few adjustments :wink:

common/src/main/java/com/ably/tracking/common/Ably.kt
ikbalkaya: There is another map called Synchronised map and I think read/writes are fully synchronised with that one with a performance drawback but I don't think we need a map with high performance
https://www.baeldung.com/java-synchronizedmap-vs-concurrenthashmap

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -269,37 +270,43 @@ constructor(
...
 
-                                override fun onError(reason: ErrorInfo) {
-                                    callback(Result.failure(reason.toTrackingException()))
+                            override fun onError(reason: ErrorInfo) {
+                                callback(Result.failure(reason.toTrackingException()))
+                                // reconnect
+                                connect(trackableId, presenceData) {
+                                    // dealing with connection result will make things more complicated here so leaving
+                                    // it blank
                                 }
```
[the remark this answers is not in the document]
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/513/commits/428bb63a34c0156c00165669fe1d23ae908b9b15

common/src/main/java/com/ably/tracking/common/Ably.kt
ikbalkaya: As you pointed out  `ConcurrentHashMap` is not completely thread safe as opposed to its name suggests. So we normally could use a mutex (or might have used an actor) to protect the map ourselves. However I think we are good to go with the synchronisedMap that Java collections framework gives us. It's basically going to do job for us and synchronise the underlying map to it. The only downside I see here is that we have to import a java framework. 
Change is here https://github.com/ably/ably-asset-tracking-android/pull/513/commits/7cc5255c6fcf90f81f5313668fbe0cdac15bb25e

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -269,37 +271,48 @@ constructor(
...
 
-                        override fun onError(reason: ErrorInfo) {
-                            callback(Result.failure(reason.toTrackingException()))
-                        }
+                            override fun onError(reason: ErrorInfo) {
+                                callback(Result.failure(reason.toTrackingException()))
+                                // reenter but do not handle reenter results as it will make things more complicated
+                                // for now
+                                channelToRemove.presence.enter(
+                                    Gson().toJson(presenceData),
```
KacperKluka, line 298: The Gson parsing part is not correct, it should be `gson.toJson(presenceData.toMessage())` :wink:

KacperKluka (changes requested):

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: So now with `SynchronizedMap` both accessing and modification of the `channels` map is multithread safe :+1: However, I think it doesn't make it much better than before :thinking: I mean that now we are sure there will be no races when looking into the `channels` map, but there still can be a case that one channel is being used by multiple threads at the same time. For example, thread A can get a channel, try to use it for sending but at the same time thread B can also get the channel and close it. 

I think that to make the Ably wrapper secure we need to make sure that each operation is atomic, meaning that if it starts then we should make sure no one will interrupt it while it's working. So there won't be a case like I've described above, that one thread is trying to send and at the same time, the channel it's using is being closed by another thread. WDYT? :wink:

ikbalkaya: In this case I think there is a misunderstanding of the issue from my side. Reframing the issue for a more concrete understanding will do it. I think this would be a good suit after I do the refactoring work. What do you think?

KacperKluka: No worries, I think that my description of the issue could be better :sweat_smile:  Yes, we should wait for the refactor as the API of the `Ably` wrapper might change due to that :wink:

KacperKluka:
I've noticed that there is no comment on why this is in the draft state now, so I'll add it before I forget it :sweat_smile:. 
The reason behind this PR being a draft is that we want to wait for the event queue refactoring https://github.com/ably/ably-asset-tracking-android/pull/525 to be completed before moving on with this PR :wink:

KacperKluka:
We've decided on a call today that we will close this PR and carry on the conversation in its [corresponding issue](https://github.com/ably/ably-asset-tracking-android/issues/440)

PR #514, opened 2021-11-17
Send resolution data as a part of the location updates

KacperKluka:
I've added the resolution data as an optional part of both raw and enhanced location updates.


AndyNicks (approved):

KacperKluka:
This issue requirements have changed so I'll create a new PR for it.

PR #521, opened 2021-11-19
Add documentation for the core publisher and core subscriber events

KacperKluka:
I've added some docs to the events used in both `CorePublisher` and `CoreSubscriber`. I've tried to only comment the whole event classes as the params are almost always self explanatory :wink:


QuintinWillison (changes requested):
I think the repeated "Event created" prefix in these comments is perhaps an indication that they are, of themselves, a bit of an anti-pattern right now...

Rather than describing _how_ and _when_ these event classes are used in their commentary, could you switch to describing _what_ they convey or ask for? For example, for `StartEvent` where you currently have:

> Event created when wanting to start the `CoreSubscriber`

replace with:

> Start the `CoreSubscriber`

Does that make sense?

QuintinWillison (approved):

Issue #326, opened 2021-05-07
Logger / Logging Support


QuintinWillison:
Requirements:

- Events occurring in the library implementation need to be logged at a range of levels (debug, info, verbose, etc..).
- The app developer needs to be able to control where these log messages go.
- By extension, the app developer should be able to contribute their own messages to the same logging destination(s).

The obvious choice is [SLF4J](https://github.com/qos-ch/slf4j). In which case there should be no need for us to expose any additional API (beyond adding the `slf4j-api` mandatory dependency) to support logging as it'll be up the app developer to choose their own logging implementation (e.g. `slf4j-simple` during development) or even write their own.

We could consider using [kotlin-logging](https://github.com/MicroUtils/kotlin-logging) on our implementation-side if it brings code-quality benefits for us, but it should not be necessary to expose this or any other Kotlin logging specialities transitively, or otherwise, through our API.

See [Marco Behler](https://github.com/marcobehler)'s [How To Do Logging In Java](https://www.marcobehler.com/guides/java-logging) as it explains things very nicely for those unfamiliar with SLF4J.

KacperKluka:
> The example apps need to compile with the `simple` implementation

The `simple` logger logs to the console which isn't very useful in Android because we should be logging to the `Logcat`. Therefore, I've used the `logback` (namely [logback-android](https://github.com/tony19/logback-android)) instead. It looks like the best choice for our use case because not only it supports `Logcat` out of the box, but also allows to create custom log handlers (called [appenders](http://logback.qos.ch/manual/appenders.html)) by just creating a class and adding it to the configuration file. :wink:

KacperKluka:
> APK size for app distributors should be assessed as part of this work

I've built debug versions of the publishing example app in two variants:
- current version of the app from the `main` branch with `Timber` logging library: 27.0 MB
- app without `Timber` but with `slf4j-api` and `logback-android`: 27.2 MB

So we're increasing the app size by 0.2MB and this is without using a tool like proguard. After obfuscation and code shrinking we'd probably save some more space :wink:

kavalerov:
Just to record internal discussion (https://ably-real-time.slack.com/archives/C01EPJENRD0/p1620922044288300  and below):What exactly is the purpose of this work, in terms of user value? 

Things that don't allow for the SDK to work properly (like permission refusal) should probably be raised as exceptions, that are handled by the app developer. In what use case do we expect this logger to be used?

kavalerov:
[...]

PR #526, opened 2021-11-24
Add log statements to publisher SDK

KacperKluka:
I've added a few verbose log statements to the `Mapbox` and `CorePublisher` classes to help us debug issues with location updates.

Here's an example of log output from receiving location to sending it:
```
14:38:00.786 [DefaultMapbox] Enhanced location received from Mapbox: Location[null 37.274426,-121.946426 acc=20 et=+18h22m11s583ms alt=0.0 vel=54.06 bear=10.857853]
14:38:00.786 [DefaultCorePublisher] Enhanced location received: Location(latitude=37.2744264173863, longitude=-121.94642617300786, altitude=0.0, accuracy=20.0, bearing=10.857853, speed=54.06, time=1637761080786)
14:38:00.786 [DefaultCorePublisher] Enhanced location changed event received Location(latitude=37.2744264173863, longitude=-121.94642617300786, altitude=0.0, accuracy=20.0, bearing=10.857853, speed=54.06, time=1637761080786)
14:38:00.786 [DefaultCorePublisher] Processing enhanced location for trackable: xxxxx. Location(latitude=37.2744264173863, longitude=-121.94642617300786, altitude=0.0, accuracy=20.0, bearing=10.857853, speed=54.06, time=1637761080786)
14:38:00.787 [DefaultCorePublisher] Trackable: xxxxx will send enhanced location Location(latitude=37.2744264173863, longitude=-121.94642617300786, altitude=0.0, accuracy=20.0, bearing=10.857853, speed=54.06, time=1637761080786)
14:38:00.788 sendEnhancedLocationMessage: publishing: {"intermediateLocations":[{"geometry":{"coordinates":[-121.94654799999999,37.273921,0.0],"type":"Point"},"properties":{"accuracyHorizontal":0.0,"bearing":10.431928,"speed":0.0,"time":1.637761080128E9},"type":"Feature"}],"location":{"geometry":{"coordinates":[-121.94642617300786,37.2744264173863,0.0],"type":"Point"},"properties":{"accuracyHorizontal":20.0,"bearing":10.857853,"speed":54.06,"time":1.637761080786E9},"type":"Feature"},"skippedLocations":[],"type":"PREDICTED"}

```

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/logging/TagCreator.kt
+++ b/common/src/main/java/com/ably/tracking/common/logging/TagCreator.kt
@@ -1,0 +1,6 @@
+package com.ably.tracking.common.logging
+
+/**
+ * Creates logging tag based on the name of the [loggingObject]'s class.
+ */
+fun createLoggingTag(loggingObject: Any): String = "[${loggingObject::class.simpleName}]"
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -16,6 +16,8 @@ import com.ably.tracking.common.ConnectionStateChange
 import com.ably.tracking.common.PresenceAction
 import com.ably.tracking.common.PresenceData
 import com.ably.tracking.common.createSingleThreadDispatcher
+import com.ably.tracking.common.logging.createLoggingTag
+import com.ably.tracking.common.logging.v
 import com.ably.tracking.common.logging.w
 import com.ably.tracking.logging.LogHandler
 import com.ably.tracking.publisher.guards.DuplicateTrackableGuard
@@ -82,6 +84,7 @@ constructor(
     private val areRawLocationsEnabled: Boolean?,
     private val sendResolutionEnabled: Boolean,
 ) : CorePublisher {
+    private val TAG = createLoggingTag(this)
     private val scope = CoroutineScope(singleThreadDispatcher + SupervisorJob())
     private val sendEventChannel: SendChannel<Event>
     private val _locations = MutableSharedFlow<LocationUpdate>(replay = 1)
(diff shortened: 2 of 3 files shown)


QuintinWillison (approved):
I don't want to block us landing this PR based on these observations, but I do want to make the observations as it may be a good idea to add them to our issues backlog for future work:

1. I can see what you're aiming to do with the `createLoggingTag` utility function, but it actually ends up with very repetitive code - with, effectively, two APIs required for every call to log - being: (1) method `v` on `logHandler`; and (2) use val `TAG`. I would prefer to see `TAG` implicit in each call to log - i.e. have an intermediary to replace `TAG` - perhaps called `LOG` on which you call `v` to automatically use the `TAG` (i.e. inject at one point in the file). This, perhaps, intersects with #326 - being more the pattern you would observe with SLF4J, for example.
2. Threading. The implication of the current design means that the app developer's implementation of log handler is going to get called from multiple threads. Elsewhere in this SDK we make a promise to call on the main thread / coroutine scope. So, it feels to me like we should be able to make similar promises here. Again, intersects with #326 as SLF4J implementations will likely do this for free.

AndyNicks:
Does the logHandler?.v() function add a timestamp to a log?

QuintinWillison:
So, I guess, in interim - regarding threading - it would be worth ensuring that the log handler interface clearly documents that it can be called from multiple threads.

KacperKluka:
> Does the logHandler?.v() function add a timestamp to a log?

@AndyNicks No, it doesn't, should we create an issue for that? :wink: 

---

> So, I guess, in interim - regarding threading - it would be worth ensuring that the log handler interface clearly documents that it can be called from multiple threads.

@QuintinWillison I think the info is already there in the interface docs :wink: 

```kotlin
/**
 * Gets called when a log message is sent from the SDK.
 * This method is called synchronously and can be called from any thread.
 * Don't run blocking code in this method as this can cause the SDK to malfunction.
 *
 * @param level The importance level of the message.
 * @param message The message text.
 * @param throwable Optional error object.
 */
fun logMessage(level: LogLevel, message: String, throwable: Throwable? = null)
```

QuintinWillison:
Perfect, thanks for clarifying @KacperKluka - I hadn't gone to depth of checking the docs myself this morning because I was short of time.

AndyNicks:
Yes please

AndyNicks (approved):

ikbalkaya (approved):
Looks good to me

KacperKluka:
I've created the issue about adding timestamps to logs: https://github.com/ably/ably-asset-tracking-android/issues/530

PR #529, opened 2021-11-25
Add trackable worker tests

ikbalkaya:
This pull request aims to add tests for AddTrackableWorker and prepare some ground  work for future unit tests. This pull request includes:

*  Extraction of some concrete dependencies to AddTrackable worker in order to provide fake implementation for worker tests. 
* Updating types to new interface type in other classes / interfaces that use those dependencies
* Creation of some fake classes that helps with AddTrackable tests. Please note that fake classes only has functionality that is needed for existing tests. In the future these classes will likely be improved to be used in other test cases. 
* Tests for AddTrackableWorker 
* Fixes on AddTrackableWorker as a result of some test failures
* A new suspendingConnect function for Ably interface that makes it easy to invoke connect() from coroutines

Note to reviewers:
Please review https://github.com/ably/ably-asset-tracking-android/pull/525 before this one


publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+
+class AddTrackableWorkerTest {
+    private lateinit var worker: AddTrackableWorker
+
+    // dependencies
+    private val resultCallbackFunction = mockk<ResultHandler<StateFlow<TrackableState>>>()
+    private val ably = FakeAbly(true)
```
KacperKluka, line 22: Why are you using fakes instead of mocks? We could get rid of the `FakeXyz` classes and simply use the `mockk` library to both mock and verify the behaviour of each of the classes :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherProperties.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.ably.tracking.TrackableState
+import com.ably.tracking.common.PresenceData
+import com.ably.tracking.publisher.guards.DuplicateTrackableGuard
+import kotlinx.coroutines.flow.MutableStateFlow
+
+/**
+ * This interface is intended to make it easy to create fake properties for tests
+ * */
+internal interface PublisherProperties {
```
KacperKluka, line 11: Creating an interface just to make the tests easier to write doesn't sound good :thinking: If something can't be tested without creating an interface then it's fine to do so, but in this situation I think that a better thing to do would be to add some test helpers that would create the required objects in the desired configuration :thinking: Especially that the `PublisherProperties` is just a container of objects, it does not hold any business logic that you might want to fake :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/DublicateTrackableGuard.kt
```diff
@@ -4,11 +4,23 @@ import com.ably.tracking.publisher.AddTrackableHandler
 import com.ably.tracking.publisher.AddTrackableResult
 import com.ably.tracking.publisher.Trackable
 
+/**
+ * Interface for [DublicateTrackableGuardImpl] is created to make it possible to create test fakes.
+ * */
+internal interface DuplicateTrackableGuard {
```
KacperKluka, line 10: Same situation as with the `PublisherProperties` :thinking: Do we really need this interface and do we want to mock/fake the behaviour of this class? I guess that if the `DuplicateTrackableGuard` is well enough tested then it shouldn't be a problem to simply use it as it is in unit tests of some other classes, WDYT? :wink:

KacperKluka (changes requested):
Sorry for being so picky but this is a major change in the SDK and we need to be very careful with it :grin: Overall I love the fact that we will be finally able to test the domain logic in small chunks with the workers queue refactor :smile:

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -175,6 +175,17 @@ interface Ably {
         callback: (Result<Unit>) -> Unit
     )
 
+    /**
+     * A suspending version of [connect]
+     * */
+    suspend fun suspendingConnect(
```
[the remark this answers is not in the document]
ikbalkaya: Good idea. Here it is https://github.com/ably/ably-asset-tracking-android/pull/529/commits/6e85b5101259a594682ac67479eef691c3933bc4

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -290,7 +291,13 @@ constructor(
...
                                 } catch (exception: ConnectionException) {
                                     ably.disconnect(event.trackable.id, properties.presenceData) {
-                                        request(AddTrackableFailedEvent(event.trackable, event.handler, exception))
+                                        request(
+                                            AddTrackableFailedEvent(
+                                                event.trackable,
+                                                event.handler,
+                                                exception
+                                            )
+                                        )
```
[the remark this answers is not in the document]
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/529/commits/3417d6c58ec855c44dbab7eaeac1dd1ebba3f663

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
ikbalkaya: Short answer is that I have not used mocks for faking behaviour before. I used them to verify some behaviour but not for faking them. I'll attempt to use mockk. Looks like you have already created an extension that I can use

publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherProperties.kt
ikbalkaya: While  I am also sceptical on this instance whether we need this interface, tests can be used to expose some design flaws in code. Looking at this; 
https://en.wikipedia.org/wiki/Dependency_inversion_principle

Even in this particular instance this interface was introduced to help for testing, it is not only useful in that situation. I think any dependency that is injected to another class should have a common interface and not expose any implementation details.

publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/DublicateTrackableGuard.kt
ikbalkaya: This was done for the same reason as above. However I think your point is valid in terms that it exposes some design flaw. So I'd say, if you don't think this should be interfaced, I'd say it then shouldn't be a dependency. So maybe duplicate trackable guarding is responsibility of the work queue which enqueues / processes the works. But I don't think it's in the scope of this PR

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
ikbalkaya: Here is I replaced fake Ably with mockk Ably 😉

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
```diff
@@ -0,0 +1,148 @@
...
+        publisherProperties.trackables.add(trackable)
+        publisherProperties.trackableStateFlows[trackable.id] = MutableStateFlow(TrackableState.Online)
+        // when
+        val result = worker.doWork(publisherProperties)
+        // then
+        Assert.assertNull(result.asyncWork)
+        Assert.assertTrue(result.syncWorkResult is AddTrackableWorkResult.AlreadyIn)
+        // also make sure it has the right content
+        val alreadyIn = result.syncWorkResult as AddTrackableWorkResult.AlreadyIn
+        Assert.assertTrue(alreadyIn.handler == resultCallbackFunction)
```
KacperKluka, line 98: We should be using `Assert.assertEquals` instead because this will give us more information when the assertion fails :wink:

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            // given
+            val publisherProperties = spyk(FakeProperties(FakeDuplicateGuard(false)))
+            ably.mockSuspendingConnectSuccess(trackable.id)
+            // when
+            val result = worker.doWork(publisherProperties)
+
+            // then
+            // first make sure there is an asyncwork
+            Assert.assertNotNull(result.asyncWork)
+            result.asyncWork?.let {
```
KacperKluka, line 115: Since you check if the `asyncWork` is not null I think we can safely use the `!!` operator to simplify this :wink:

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            // given
+            val publisherProperties = spyk(FakeProperties(FakeDuplicateGuard(false)))
+            ably.mockSuspendingConnectFailure(trackable.id)
+            // when
+            val result = worker.doWork(publisherProperties)
+
+            // then
+            // first make sure there is an asyncwork
+            Assert.assertNotNull(result.asyncWork)
+            result.asyncWork?.let {
```
KacperKluka, line 138: Like above :wink:

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
```diff
@@ -0,0 +1,148 @@
...
+import io.mockk.spyk
+import io.mockk.verify
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.runBlocking
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+
+class AddTrackableWorkerTest {
```
KacperKluka, line 20: Please add vertical spaces in all tests methods so the `given when then` comments have a vertical space before them, for example:
```
        // given
        val publisherProperties = FakeProperties(FakeDuplicateGuard(false))

        // when
        val result = worker.doWork(publisherProperties)

        // then
        Assert.assertNotNull(result.asyncWork)
```

KacperKluka (changes requested):
Looks good but we can make a few improvements :wink:

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/FakeProperties.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.ably.tracking.publisher.guards.DuplicateTrackableGuard
+import kotlinx.coroutines.flow.MutableStateFlow
+
+internal class FakeProperties(override val duplicateTrackableGuard: DuplicateTrackableGuard) : PublisherProperties {
+    override val trackables: MutableSet<Trackable> = mutableSetOf()
+    override val trackableStateFlows: MutableMap<String, MutableStateFlow<TrackableState>> = mutableMapOf()
+    override var presenceData: PresenceData = PresenceData("properties")
+}
+
+internal class FakeDuplicateGuard(private val currentlyAdding: Boolean) : DuplicateTrackableGuard {
```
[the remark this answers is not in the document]
ikbalkaya: I created an issue for this https://github.com/ably/ably-asset-tracking-android/issues/551

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
```diff
@@ -0,0 +1,148 @@
...
+    private lateinit var worker: AddTrackableWorker
+
+    // dependencies
+    private val resultCallbackFunction = mockk<ResultHandler<StateFlow<TrackableState>>>()
```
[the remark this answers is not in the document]
ikbalkaya: I will address this in another PR. See https://github.com/ably/ably-asset-tracking-android/issues/551

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
ikbalkaya: Do you really think this is necessary?

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
ikbalkaya: Fair enough https://github.com/ably/ably-asset-tracking-android/pull/529/commits/b9163eecc3e1e192fb71895a370fa60702de4c70

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/529/commits/8ceee9d5b50cf8429a556b71687e73698b2ae28b

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
KacperKluka: Thanks but there are still a few `assertTrue()` in your tests when they should be `assertEquals`, please change them too :wink: 
Also, I think that the order of the arguments should be inverted :thinking: In the `assertEquals` the first argument is the expected value and the second one is the actual one, so in this case, it should probably be `Assert.assertEquals(resultCallbackFunction, alreadyIn.handler)` because we expect to see `resultCallbackFunction` and the `alreadyIn.handler` is the thing that we receive. It is important because then you have clear and precise assertion error messages and you don't have to wonder what's actually wrong in the test :wink:

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
KacperKluka: Yes, I think that clean and well-written tests are very important in maintaining a healthy codebase :wink: So we should aim to make them as good as our production code :grin: However, if this will cause you a lot of trouble due to the chain of PRs I'm fine with creating an issue for that and making the change later :wink:

ikbalkaya: Ah no, my objection is that I do not think it is a good idea to forcely unwrap an optional. Consider that we added !! operator and then later 
`Assert.assertNotNull(result.asyncWork)`
 is not needed and removed. Wouldn't you be left with a code that unwraps a potentially null value. Of course for this particular tests it's on the next line, but in any client situation I wouldn't use !! operator for an optional value.

publishing-sdk/src/test/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorkerTest.kt
ikbalkaya: Thanks for your suggestions. I applied your suggestion to the place of comment. I will address others in a different PR.

ikbalkaya: https://github.com/ably/ably-asset-tracking-android/issues/558

KacperKluka (approved):
LGTM :+1:

PR #537, opened 2021-12-02
Connection for trackable created worker

ikbalkaya:
Worker implementation for ConnectionForTrackableCreatedEvent

Note to reviewers: Please review https://github.com/ably/ably-asset-tracking-android/pull/529 before this one

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -208,6 +208,11 @@ interface Ably {
     fun disconnect(trackableId: String, presenceData: PresenceData, callback: (Result<Unit>) -> Unit)
 
     /**
+     * A suspending version of [disconnect]
+     * */
+    suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Boolean>
+
+    /**
      * Cleanups and closes all the connected channels and their presence. In the end closes Ably connection.
      *
      * @param presenceData The data that will be send via the presence channels.
@@ -379,13 +384,14 @@ constructor(
      * A suspend version of the [DefaultAbly.disconnect] method. It waits until disconnection is completed.
      * @throws ConnectionException if something goes wrong during disconnect.
      */
-    private suspend fun disconnect(trackableId: String, presenceData: PresenceData) {
-        suspendCoroutine<Unit> { continuation ->
+    override suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Boolean> {
+        return suspendCoroutine { continuation ->
             disconnect(trackableId, presenceData) {
                 try {
-                    continuation.resume(it.getOrThrow())
+                    it.getOrThrow()
+                    continuation.resume(Result.success(true))
                 } catch (exception: ConnectionException) {
-                    continuation.resumeWithException(exception)
+                    continuation.resume(Result.failure(exception))
                 }
             }
         }
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -382,7 +382,6 @@ constructor(
 
     /**
      * A suspend version of the [DefaultAbly.disconnect] method. It waits until disconnection is completed.
-     * @throws ConnectionException if something goes wrong during disconnect.
      */
     override suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Unit> {
         return suspendCoroutine { continuation ->
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -210,7 +210,7 @@ interface Ably {
     /**
      * A suspending version of [disconnect]
      * */
-    suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Boolean>
+    suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Unit>
 
     /**
      * Cleanups and closes all the connected channels and their presence. In the end closes Ably connection.
@@ -384,12 +384,12 @@ constructor(
      * A suspend version of the [DefaultAbly.disconnect] method. It waits until disconnection is completed.
      * @throws ConnectionException if something goes wrong during disconnect.
      */
-    override suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Boolean> {
+    override suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Unit> {
         return suspendCoroutine { continuation ->
             disconnect(trackableId, presenceData) {
                 try {
                     it.getOrThrow()
-                    continuation.resume(Result.success(true))
+                    continuation.resume(Result.success(Unit))
                 } catch (exception: ConnectionException) {
                     continuation.resume(Result.failure(exception))
                 }
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
@@ -31,11 +31,7 @@ internal class ConnectionCreatedResultHandler : WorkResultHandler {
                 corePublisher.request(ConnectionForTrackableReadyEvent(workResult.trackable, workResult.handler))
             }
             is ConnectionCreatedWorkResult.PresenceFail -> corePublisher.request(
-                AddTrackableFailedEvent(
-                    workResult.trackable,
-                    workResult.handler,
-                    workResult.exception
-                )
+                AddTrackableFailedEvent(workResult.trackable, workResult.handler, workResult.exception)
             )
         }
         return null
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
@@ -1,6 +1,5 @@
 package com.ably.tracking.publisher.workerqueue.resulthandlers
 
-import com.ably.tracking.ConnectionException
 import com.ably.tracking.publisher.AddTrackableFailedEvent
 import com.ably.tracking.publisher.ConnectionForTrackableReadyEvent
 import com.ably.tracking.publisher.CorePublisher
@@ -20,10 +19,7 @@ internal class ConnectionCreatedResultHandler : WorkResultHandler {
                     TrackableRemovalRequestedEvent(
                         trackable = workResult.trackable,
                         handler = workResult.handler,
-                        result = if (workResult.successfulDisconnect) Result.success(Unit) else Result.failure(
-                            workResult.exception
-                                as ConnectionException
-                        )
+                        result = workResult.result
                     )
                 )
 
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
@@ -1,0 +1,44 @@
+package com.ably.tracking.publisher.workerqueue.resulthandlers
+
+import com.ably.tracking.ConnectionException
+import com.ably.tracking.publisher.AddTrackableFailedEvent
+import com.ably.tracking.publisher.ConnectionForTrackableReadyEvent
+import com.ably.tracking.publisher.CorePublisher
+import com.ably.tracking.publisher.TrackableRemovalRequestedEvent
+import com.ably.tracking.publisher.workerqueue.ConnectionCreatedWorkResult
+import com.ably.tracking.publisher.workerqueue.WorkResult
+import com.ably.tracking.publisher.workerqueue.WorkResultHandler
+import com.ably.tracking.publisher.workerqueue.workers.Worker
+
+internal class ConnectionCreatedResultHandler : WorkResultHandler {
+    override fun handle(
+        workResult: WorkResult,
+        corePublisher: CorePublisher
+    ): Worker? {
+        when (workResult) {
+            is ConnectionCreatedWorkResult.RemovalRequested ->
+                corePublisher.request(
+                    TrackableRemovalRequestedEvent(
+                        trackable = workResult.trackable,
+                        handler = workResult.handler,
+                        result = if (workResult.successfulDisconnect) Result.success(Unit) else Result.failure(
+                            workResult.exception
+                                as ConnectionException
+                        )
+                    )
+                )
+
+            is ConnectionCreatedWorkResult.PresenceSuccess -> {
+                corePublisher.request(ConnectionForTrackableReadyEvent(workResult.trackable, workResult.handler))
+            }
+            is ConnectionCreatedWorkResult.PresenceFail -> corePublisher.request(
+                AddTrackableFailedEvent(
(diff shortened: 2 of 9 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+
+            is ConnectionCreatedWorkResult.PresenceSuccess -> {
+                corePublisher.request(ConnectionForTrackableReadyEvent(workResult.trackable, workResult.handler))
+            }
+            is ConnectionCreatedWorkResult.PresenceFail -> corePublisher.request(
+                AddTrackableFailedEvent(
+                    workResult.trackable,
+                    workResult.handler,
+                    workResult.exception
+                )
```
KacperKluka, line 38: This can be safely joined into one line without reaching the line break limit :wink: 
```kotlin
AddTrackableFailedEvent(workResult.trackable, workResult.handler, workResult.exception)
```

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+        when (workResult) {
+            is ConnectionCreatedWorkResult.RemovalRequested ->
+                corePublisher.request(
+                    TrackableRemovalRequestedEvent(
+                        trackable = workResult.trackable,
+                        handler = workResult.handler,
+                        result = if (workResult.successfulDisconnect) Result.success(Unit) else Result.failure(
+                            workResult.exception
+                                as ConnectionException
+                        )
```
KacperKluka, line 26: I think that when we correct the `ably.disconnect()` suspending method signature then this whole `if` will become unnecessary and we will be able to simply pass forward the `result` from the `disconnect` here :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
```diff
@@ -47,3 +49,24 @@ internal sealed class AddTrackableWorkResult() : WorkResult() {
...
+        val trackable: Trackable,
+        val handler: ResultHandler<StateFlow<TrackableState>>,
+        val successfulDisconnect: Boolean,
+        val exception: ConnectionException? = null
+    ) : ConnectionCreatedWorkResult()
+
+    internal data class PresenceSuccess(
+        val trackable: Trackable,
+        val handler: ResultHandler<StateFlow<TrackableState>>,
+        val presenceUpdateListener: (presenceMessage: PresenceMessage) -> Unit
```
KacperKluka, line 64: This is never used - perhaps we should remove it? :thinking:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
```diff
@@ -47,3 +49,24 @@ internal sealed class AddTrackableWorkResult() : WorkResult() {
...
+
+internal sealed class ConnectionCreatedWorkResult : WorkResult() {
+    internal data class RemovalRequested(
+        val trackable: Trackable,
+        val handler: ResultHandler<StateFlow<TrackableState>>,
+        val successfulDisconnect: Boolean,
+        val exception: ConnectionException? = null
+    ) : ConnectionCreatedWorkResult()
+
+    internal data class PresenceSuccess(
```
KacperKluka, line 61: I think we should be more verbose with names and rename it to `SubscribeToPresenceSuccess` :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
```diff
@@ -47,3 +49,24 @@ internal sealed class AddTrackableWorkResult() : WorkResult() {
...
+        val trackable: Trackable,
+        val handler: ResultHandler<StateFlow<TrackableState>>,
+        val presenceUpdateListener: (presenceMessage: PresenceMessage) -> Unit
+    ) : ConnectionCreatedWorkResult()
+
+    internal data class PresenceFail(
```
KacperKluka, line 67: I think we should be more verbose and rename it to `SubscribeToPresenceFailed` :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
```diff
@@ -0,0 +1,76 @@
...
+        }
+        val presenceData = properties.presenceData.copy()
+        return SyncAsyncResult(
+            asyncWork = {
+                subscribeToPresenceMessages(presenceData)
+            }
+        )
+    }
+
+    private suspend fun subscribeToPresenceMessages(presenceData: PresenceData): ConnectionCreatedWorkResult {
```
KacperKluka, line 56: Why it's here and not in the `Ably` interface? :thinking:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
```diff
@@ -0,0 +1,76 @@
...
+                        result.getOrThrow()
+                        continuation.resume(
+                            ConnectionCreatedWorkResult.PresenceSuccess(trackable, handler, presenceUpdateListener)
+                        )
+                    } catch (exception: ConnectionException) {
+                        ably.disconnect(trackable.id, presenceData) {
+                            continuation.resume(ConnectionCreatedWorkResult.PresenceFail(trackable, handler, exception))
+                        }
+                    }
+                }
```
KacperKluka, line 72: I think that too much is happening here, instead of returning the `WorkResult` I'd suggest simply returning a success or failure and then using it to perform more actions. Thanks to this you'll be, for example, able to use suspending `disconnect` instead of the callback-based one :wink:

KacperKluka (changes requested):
Some improvements are needed :wink:

KacperKluka:
I've hoped to find another PR that would contain tests for this worker but it looks like there isn't one. Please add tests for this worker as a part of this PR :bow:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
ikbalkaya: I didn't want to pollute Ably interface with more suspending functions. Instead return a suspending result for the use case. Ably interface refactorings is not part of this refactoring work (I know I did one in one of previous PRs but I think it gets out of the scope). If you still think this should move to Ably, please add an issue and I'll adress it there

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
ikbalkaya: These are subclasses for ConnectionCreatedWorkResult and I think they do not need to be verbose as they are already scoped inside a particular result

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
ikbalkaya: As above

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -207,6 +207,11 @@ interface Ably {
      */
     fun disconnect(trackableId: String, presenceData: PresenceData, callback: (Result<Unit>) -> Unit)
 
+    /**
+     * A suspending version of [disconnect]
+     * */
+    suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Boolean>
```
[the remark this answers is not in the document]
ikbalkaya: The original intention was to return whether the disconnect success or failed, but I see that those are already handled by Result class. I'll update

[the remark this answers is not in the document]
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/537/commits/646dd114c1682ef61caf7b3b1495e00fe40c6f88

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/ConnectionCreatedResultHandler.kt
ikbalkaya: [I'm wondering  ](https://github.com/ably/ably-asset-tracking-android/pull/537/commits/9aad9b1309824fbbcb68583f08aec0226d0709cb)

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
ikbalkaya: I added this for the future iteration of refactorings. Just like other callbacks we will probably end up dealing with this callback in a result handler. See CorePublisher class and see how ConnectionCreatedWorker is invoked

```
  workerQueue.execute(
                            ConnectionCreatedWorker(
                                event.trackable,
                                event.handler,
                                ably
                            ) { presenceMessage ->
                                enqueue(PresenceMessageEvent(event.trackable, presenceMessage))
                            }
                        )
```
`enqueue(PresenceMessageEvent(event.trackable, presenceMessage))
` 
will have to be handled from the corresponding work handler. However this refactoring will happen in the future iterations

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
ikbalkaya: Good point 👍  However as in others I intentionally kept the async block as it's original implementation. I am going to add an issue to address this later

ikbalkaya: https://github.com/ably/ably-asset-tracking-android/issues/554

ikbalkaya:
> I've hoped to find another PR that would contain tests for this worker but it looks like there isn't one. Please add tests for this worker as a part of this PR 🙇

I'm going to add test PRs for workers after refactoring for workers is complete.

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
KacperKluka: I think that we should follow the YAGNI rule and don't add stuff that will probably be used sometime in the future :wink: If we will need it then we can always add it in the future, if it's not needed now, then let's not add it :grin:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
KacperKluka: Yes, they are scoped, but what does exactly `PresenceSuccess` mean in the scope of `ConnectionCreated`? :thinking: Does it mean that the presence was successfully joined? Or that we've successfully subscribed to the presence channel? This is not obvious, therefore, I still think that the name should change to something that would explicitly tell you what has happened :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
KacperKluka: You're right, it makes sense to not add anything to the Ably interface as this feels a bit outside of the scope of the refactoring :wink:

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -379,13 +384,14 @@ constructor(
...
+    override suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Unit> {
+        return suspendCoroutine { continuation ->
             disconnect(trackableId, presenceData) {
                 try {
-                    continuation.resume(it.getOrThrow())
+                    it.getOrThrow()
+                    continuation.resume(Result.success(Unit))
                 } catch (exception: ConnectionException) {
-                    continuation.resumeWithException(exception)
+                    continuation.resume(Result.failure(exception))
```
KacperKluka, line 394: Because you've changed the error path behaviour the documentation of this function needs to be updated :wink: We should remove this line as it is no longer valid:
```
@throws ConnectionException if something goes wrong during disconnect.
```

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
```diff
@@ -0,0 +1,71 @@
...
+            // Leave Ably channel.
+            val presenceData = properties.presenceData.copy()
+            return SyncAsyncResult(
+                asyncWork = {
+                    val result = ably.disconnect(trackable.id, presenceData)
+                    if (result.isSuccess) {
+                        ConnectionCreatedWorkResult.RemovalRequested(trackable, handler, result)
+                    } else {
+                        ConnectionCreatedWorkResult.RemovalRequested(trackable, handler, result)
+                    }
```
KacperKluka, line 39: Both branches of this `if / else` execute the same code (I wonder why lint isn't complaining about it :thinking:). Therefore, we don't need this `if` and this can be simplified :wink: 
```kotlin
ConnectionCreatedWorkResult.RemovalRequested(trackable, handler, result)
```

KacperKluka (changes requested):
Almost there :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
ikbalkaya: Well. As far as I know YAGNI means "you aren't gonna need it" as I tried to explain in comment above "you are gonna need it". I make this intent explicit for now so that it is not forgotten later.

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/results/WorkResult.kt
ikbalkaya: All cases were extracted from the original code as use case results. In this case you can for example read it as
"I have tried to create a connection, and the presence succeeded as a result of this work, what should I do?" 
You can ask the same question for other 2 cases. As they are both part of`ConnectionCreatedWorkResult` it is only possible to use them on those particular cases

common/src/main/java/com/ably/tracking/common/Ably.kt
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/pull/537/commits/0c45da6151272009587fbc8f9f4eee3a16c3863b

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
ikbalkaya: Very interesting of IDE is not complaining about this. Change is here https://github.com/ably/ably-asset-tracking-android/pull/537/commits/0c45da6151272009587fbc8f9f4eee3a16c3863b

KacperKluka (approved):
LGTM :+1:

PR #538, opened 2021-12-02
Rename handler to callback function

ikbalkaya:
This PR renames Handler to CallbackFunction and updates corresponding files and reference names

Please review https://github.com/ably/ably-asset-tracking-android/pull/537 before this one

PR #539, opened 2021-12-02
Add trackable failed worker

ikbalkaya:
This PR contains AddTrackableFailedEvent refactoring 

Please review https://github.com/ably/ably-asset-tracking-android/pull/538 before this one

Note for this particular worker:
In the current implementation, the callback is in the middle of execution, so I didn't find it safe to change the order of operations here to return a result. This implementation will have to change a little so that we provide a work result and work result handler. I have created an issue for this. https://github.com/ably/ably-asset-tracking-android/issues/546


KacperKluka (changes requested):
Overall looks good but tests for the new worker are missing :wink:

ikbalkaya:
> Overall looks good but tests for the new worker are missing 😉

As others, I shall add test PRs after all worker refactorings

KacperKluka (approved):
LGTM :+1:

PR #540, opened 2021-12-03
Presence message worker

ikbalkaya:
PresenceMessageEvent refactoring implementation
Please review https://github.com/ably/ably-asset-tracking-android/pull/539 before this one

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/EventWorkerQueue.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/EventWorkerQueue.kt
@@ -1,7 +1,9 @@
 package com.ably.tracking.publisher.workerqueue
 
+import com.ably.tracking.publisher.AdhocEvent
 import com.ably.tracking.publisher.CorePublisher
 import com.ably.tracking.publisher.DefaultCorePublisher
+import com.ably.tracking.publisher.Request
 import com.ably.tracking.publisher.workerqueue.resulthandlers.getWorkResultHandler
 import com.ably.tracking.publisher.workerqueue.workers.Worker
 import kotlinx.coroutines.CoroutineScope
@@ -42,6 +44,13 @@ internal class EventWorkerQueue(
     private fun handleWorkResult(workResult: WorkResult) {
         val resultHandler = getWorkResultHandler(workResult)
         val nextWorker = resultHandler.handle(workResult, corePublisher)
-        nextWorker?.let { corePublisher.request(it.event) }
+        nextWorker?.let {
+            it.event.apply {
+                when (this) {
+                    is Request<*> -> corePublisher.request(this)
+                    is AdhocEvent -> corePublisher.enqueue(this)
+                }
+            }
+        }
     }
 }
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/PresenceMessageWorker.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/PresenceMessageWorker.kt
@@ -8,7 +8,7 @@ import com.ably.tracking.publisher.Event
 import com.ably.tracking.publisher.PresenceMessageEvent
 import com.ably.tracking.publisher.PublisherProperties
 import com.ably.tracking.publisher.Trackable
-import com.ably.tracking.publisher.workerqueue.SyncAsyncResult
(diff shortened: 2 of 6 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/EventWorkerQueue.kt
```diff
@@ -44,6 +46,13 @@ internal class EventWorkerQueue(
...
         val resultHandler = getWorkResultHandler(workResult)
         val nextWorker = resultHandler.handle(workResult, corePublisher)
-        nextWorker?.let { corePublisher.request(it.event) }
+        nextWorker?.let {
+            it.event.apply {
+                when (this) {
+                    is Request<*> -> corePublisher.request(this)
+                    is AdhocEvent -> corePublisher.enqueue(this)
+                }
+            }
```
KacperKluka, line 55: I would choose some other scope function for this use-case, as the `apply` does not seem very idiomatic here :thinking: According to the [docs](https://kotlinlang.org/docs/scope-functions.html#apply) the `apply` function is mostly used for object configuration and here we're doing something different :wink: 

I'd either use the `let` function but with a named parameter, something like this
```suggestion
            it.event.let { event ->
                when (event) {
                    is Request<*> -> corePublisher.request(event)
                    is AdhocEvent -> corePublisher.enqueue(event)
                }
            }
```

or we can use `run` which will still use `this` as the event
```suggestion
            it.event.run {
                when (this) {
                    is Request<*> -> corePublisher.request(this)
                    is AdhocEvent -> corePublisher.enqueue(this)
                }
            }
```

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/PresenceMessageWorker.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.ably.tracking.publisher.Event
+import com.ably.tracking.publisher.PresenceMessageEvent
+import com.ably.tracking.publisher.PublisherProperties
+import com.ably.tracking.publisher.Trackable
+import com.ably.tracking.publisher.workerqueue.results.SyncAsyncResult
+
+internal class PresenceMessageWorker(
+    private val trackable: Trackable,
+    private val presenceMessage: PresenceMessage,
+    private val corePublisher: CorePublisher
```
KacperKluka, line 16: I don't think that we should be referencing the `CorePublisher` from any worker :thinking: Wasn't the main reason behind having `WorkResultHandlers` to decouple the workers from the queue? If I understand it correctly we should create a work result handler for this worker and use the core publisher in that result handler :wink:

KacperKluka (changes requested):
Looks good but there are a few things to improve :wink: Also there are no tests, please add tests for the worker as a part of this PR :wink:

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/EventWorkerQueue.kt
ikbalkaya: In order to avoid conflicts from merge, I will address stylistic changes on next set of PRs. I will link the issue here when I create it

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/PresenceMessageWorker.kt
ikbalkaya: I remember chatting about this. Core publisher is here only because it is implementing the functions that are used in this particular worker. As everything was inside the same class before refactoring, an easy way would be to define the functions as CorePublisher and use them on worker. However these functions will become part of another interface (Eg. SubscriptionEditor) and is going to replace CorePublisher as dependency. But is not part of this PR. I shall create an issue for this.

ikbalkaya: https://github.com/ably/ably-asset-tracking-android/issues/555

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/EventWorkerQueue.kt
ikbalkaya: https://github.com/ably/ably-asset-tracking-android/issues/556

KacperKluka (approved):
LGTM :+1:

PR #541, opened 2021-12-03
Trackable removal requested worker implementation 

ikbalkaya:
Refactoring for TrackableRemovalRequestedEvent 

Please review https://github.com/ably/ably-asset-tracking-android/pull/540 before this one

PR #542, opened 2021-12-03
Set active trackable worker

ikbalkaya:
Refactoring for SetActiveTrackableEvent

Please review 
https://github.com/ably/ably-asset-tracking-android/pull/541 before this one

Some notes:
* For this and some previous refactorings I did not provide a work result and work result handlers. I did not do this because this classes might likely undergo another iteration (in later stage of refactoring)
* I provided CorePublisher as dependency to do some works (eg. setDestination) Those functions uncover the need for different interfaces for those particular responsibilities. I did not do this refactoring as it was not in scope of this PR


publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/SetActiveTrackableWorker.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.ably.tracking.publisher.PublisherProperties
+import com.ably.tracking.publisher.SetActiveTrackableEvent
+import com.ably.tracking.publisher.Trackable
+import com.ably.tracking.publisher.workerqueue.results.SetActiveTrackableResult
+import com.ably.tracking.publisher.workerqueue.results.SyncAsyncResult
+
+internal class SetActiveTrackableWorker(
+    private val trackable: Trackable,
+    private val callbackFunction: ResultCallbackFunction<Unit>,
+    private val publisher: CorePublisher,
```
KacperKluka, line 16: As I've mentioned in a previous PR I believe that the worker should not have access to the `CorePublisher` as this is the job of the work result handler, right? Therefore any code using the `publisher` should be moved to the handler.

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/SetActiveTrackableWorker.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    override fun doWork(properties: PublisherProperties): SyncAsyncResult {
+        if (properties.active != trackable) {
+            properties.active = trackable
+
+            // In the future consider moving following lines to handler
+            hooks.trackables?.onActiveTrackableChanged(trackable)
+            trackable.destination?.let {
+                publisher.setDestination(it, properties)
+            }
```
KacperKluka, line 30: Please move the `publisher.setDestination()` to the work result handler :wink:

KacperKluka (changes requested):
Looks good but needs a few changes :wink: Also please add tests for the worker

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/SetActiveTrackableWorker.kt
KacperKluka: Now I understand that this is the right reasoning :wink: Publisher here is used because it has too many responsibilities and this shouldn't imply that we have to move it to the work handler.

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/SetActiveTrackableWorker.kt
KacperKluka: This indicates that the `publisher` has too many responsibilities and needs some more refactoring, I've created an issue for this https://github.com/ably/ably-asset-tracking-android/issues/568

PR #543, opened 2021-12-06
Stop worker 

ikbalkaya:
This PR contains 
* Refactoring of work inside StopEvent
* Marking WorkerQueue and Worker functions as `suspend` as it was before. That's because original stop operation used a suspending function from Ably. I wanted to keep it compatible with original one.

Like some previous others there is no WorkResult or Result handler for this worker. If a need arises it will be covered in other PRs


KacperKluka (changes requested):
The stopping procedure is an exception to our rule that we don't want to block the queue processing. Therefore it feels wrong to modify the whole `Worker` API and change the `doWork` to a suspending function in order to support this single case. Also, we've made the `doWork` a non-suspendable function on purpose as we didn't want to block the queue by mistake. 
Therefore, we need to find some other way to implement the stopping operations in the new "workers world" :thinking:

KacperKluka (approved):
Instead of changing the `doWork` to a suspending function, I've used the `runBlocking` block inside the `StopWorker` to make it block the publisher while it's stopping.

KacperKluka:
I've created the issue for improving the blocking workers mechanism https://github.com/ably/ably-asset-tracking-android/issues/570 :wink:

PR #547, opened 2021-12-07
Allow to disable the enhanced location predictions in the publisher

KacperKluka:
I've added the option to disable the predictions to the `Publisher` builder. To disable the location predictions we need to set the `navigatorPredictionMillis` to `0`.
I've also added a switch to the publisher example app that enables or disables the predictions.


AndyNicks (approved):

ikbalkaya (approved):
Looks good to me 😎

QuintinWillison (approved):

PR #548, opened 2021-12-07
Clear the route when the destination is reached

KacperKluka:
I've added an arrival observer which notifies us when we reach the destination. Then we clear the route in order to improve the Mapbox behaviour after reaching the destination.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
@@ -170,7 +170,7 @@ internal class DefaultMapbox(
     private var mapboxReplayer: MapboxReplayer? = null
     private var locationHistoryListener: (LocationHistoryListener)? = null
     private var locationObserver: LocationObserver? = null
-    private var arrivalObserver: ArrivalObserver? = null
+    private lateinit var arrivalObserver: ArrivalObserver
 
     init {
         setupTripNotification(notificationProvider, notificationId)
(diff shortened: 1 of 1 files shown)


AndyNicks (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
```diff
@@ -167,6 +170,7 @@ internal class DefaultMapbox(
     private var mapboxReplayer: MapboxReplayer? = null
     private var locationHistoryListener: (LocationHistoryListener)? = null
     private var locationObserver: LocationObserver? = null
+    private var arrivalObserver: ArrivalObserver? = null
```
ikbalkaya, line 173: I can see that this is being initialised in init block. Shouldn't this be a non-optional val if I'm not missing something 🤔

ikbalkaya (changes requested):
Looks good overall 😎 Just one clarification needed (not necessarily a code change request)

publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
KacperKluka: Good idea :+1: However, I couldn't use `val` for some reason so I've changed it to a non-null `lateinit var` :wink: https://github.com/ably/ably-asset-tracking-android/pull/548/commits/94d35cafa3b40665d5ad3e597c52c64f20eef8f0

ikbalkaya (approved):

QuintinWillison (approved):

PR #562, opened 2021-12-13
Change ResolutionConstraints to an interface

KacperKluka:
I've changed the `ResolutionConstraints` from a sealed class to an interface. Because of that I had to add a few checks in the `DefaultResolutionPolicy` to make sure we are using the expected resolution constraints type.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/ConfigurationModels.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/ConfigurationModels.kt
@@ -331,7 +331,7 @@ internal fun DefaultResolutionSet.getResolution(isNear: Boolean, hasSubscriber: 
 /**
  * Specifies factors which contribute towards deciding the tracking [Resolution] for a [Trackable].
  */
-sealed class ResolutionConstraints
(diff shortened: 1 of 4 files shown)


AndyNicks (approved):

QuintinWillison (commented):
I'm not very sure why we're making this change, for the time being... there is no clear reason, for me, what this buys for anybody - us or the app developer. 🤔 

I think `sealed class` is more communicative in terms of the current capability constraints of the SDK. And, as far as I can see, the move to change this to an `interface` can be made in the future at the point when we support user implementations of this concept.

Or am I missing something?

KacperKluka:
@QuintinWillison the thing with `sealed class`es is that you can't add new implementations outside of the original module. This is the error you get when you try to add a new implementation: 
```
Inheritance of sealed classes or interfaces from different module is prohibited
```
This means that when you are using AAT as a dependency you can't add another `ResolutionConstraints`. Changing it to an interface allows users to create their own implementations and use them in their own `ResolutionPolicy' implementations :wink:

QuintinWillison (approved):
I was missing something, indeed! Thanks for reminding me of the details in this area. 😄

PR #572, opened 2021-12-17
Mapbox v2 migration draft

KacperKluka:
A rough and dirty Mapbox v2 migration attempt.
The code compiles but we get a missing dependency runtime crash.


KacperKluka:
Closing as I'm about to create a proper migration PR :wink:

PR #574, opened 2021-12-20
Migrate to Mapbox v2

KacperKluka:
I've migrated Mapbox dependencies that we use to the newest possible versions. The major changes in the Publisher SDK were the way of retrieving the history data and the way of setting the destination. The major change in the publisher example app was switching from map markers to map annotations.
I couldn't use the newest `com.mapbox.maps:android` in version `10.2.0` as it didn't work with the newest stable Mapbox navigation SDK. Therefore, I've used the newest version that worked which is `10.1.0`.

While doing this I also had to do a few things not directly related to Mapbox:
- upgrade the gradle version as without it the `resValue` function required to setup Mapbox in the example app wouldn't properly work
- change the way of calling the `SetDestinationActivity` from the deprecated `startActivityForResult` to a result handler approach as the ktlint was failing on this

### DIFF
--- a/publishing-example-app/build.gradle
+++ b/publishing-example-app/build.gradle
@@ -16,6 +16,8 @@ android {
         vectorDrawables.useSupportLibrary = true
         applicationId 'com.ably.tracking.example.publisher'
         proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        // Required to initialize Mapbox and use MapView in the layout XML files.
+        resValue "string", "mapbox_access_token", property('MAPBOX_ACCESS_TOKEN')
     }
 }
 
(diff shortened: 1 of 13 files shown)


AndyNicks (approved):

publishing-example-app/build.gradle
```diff
@@ -16,6 +16,8 @@ android {
         vectorDrawables.useSupportLibrary = true
         applicationId 'com.ably.tracking.example.publisher'
         proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        // Required to initialize Mapbox and use MapView in the layout XML files.
+        resValue "string", "mapbox_access_token", property('MAPBOX_ACCESS_TOKEN')
```
QuintinWillison, line 20: I've pulled this branch in locally and done a `grep` for `mapbox_access_token` and can't seem to find where it's being used.

Also, if it is being used, my preference would be to preserve case for this property - as we have elsewhere - i.e. keep it all caps.

QuintinWillison (approved):
One small query, but benign in nature so happy for it to be fixed (if a fix is needed) in a subsequent PR, after the imminent release.

publishing-example-app/build.gradle
KacperKluka: This is a part of the Mapbox configuration in the publisher example app :wink: You can find more details about it [here](https://docs.mapbox.com/android/maps/guides/install/#configure-credentials) under `Configure your public token`. Before the migration, you could set it up programmatically during the runtime but now you have to put it into XML resources.

QuintinWillison: Still feels somewhat opaque to me, and I don't think the MapBox documentation you pointed me towards provides any clarity either. To quote that documentation:

> There are many ways to configure your public access token. Many of the examples and code snippets on this site assume your token is stored in a file in your project with other string values. If you would like to manage your public access token this way, open your project's R.strings.xml file and add the following string resource:
> `<string name="mapbox_access_token">MAPBOX_ACCESS_TOKEN</string>`

I have no idea what "many ways" means there, as it seems there is only this one way illustrated. Very odd.

PR #580, opened 2021-12-30
Add sources and javadoc JARs to the published library

KacperKluka:
I've added two additional JARs to the published library packages. One of them contains the javadocs and the other contains the source code. In order to do that I had to add the `dokka` plugin to all modules that are being published. 
Thanks to that, when we use the library from a maven repository we can lookup the code and view the docs.

### DIFF
--- a/publish.gradle
+++ b/publish.gradle
@@ -3,6 +3,22 @@ apply plugin: 'maven-publish'
 final githubActor = findProperty('GITHUB_ACTOR')
 final githubToken = findProperty('GITHUB_TOKEN')
 
+task sourcesJar(type: Jar) {
+    // Copy files from main source directories to the JAR.
+    from android.sourceSets.main.java.srcDirs
+    // Specify the JAR type (it will be appended to its filename).
+    archiveClassifier.set("sources")
+}
+
+task dokkaJavadocJar(type: Jar) {
+    // Generate the javadocs.
+    dependsOn dokkaJavadoc
+    // Copy files from the javadocs directory to the jar.
+    from dokkaJavadoc.outputDirectory
+    // Specify the JAR type (it will be appended to its filename).
+    archiveClassifier.set("javadoc")
+}
+
 afterEvaluate {
     // Based on: https://developer.android.com/studio/build/maven-publish-plugin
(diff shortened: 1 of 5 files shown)


publish.gradle
```diff
@@ -3,13 +3,34 @@ apply plugin: 'maven-publish'
 final githubActor = findProperty('GITHUB_ACTOR')
 final githubToken = findProperty('GITHUB_TOKEN')
 
+task sourcesJar(type: Jar) {
+    // Copy files from main source directories to the JAR.
+    from android.sourceSets.main.java.srcDirs
+    // Specify the JAR type (it will be appended to its filename).
+    archiveClassifier.set("sources")
+}
```
ikbalkaya, line 11: I have added a similar task in https://github.com/ably/ably-asset-tracking-android/blob/d83b579b668f0cd19aa2d48943928245f702e960/publish-module.gradle. 

In there the sourceSets that are looked at are slightly different (android.sourceSets if an Android library). Rest seems to be the same. I wonder if there is a way to validate that all sources are included

ikbalkaya (changes requested):
I realised that might be a need to check whether all source sets are looked at in sourcesJar tasks. Otherwise looks good

publish.gradle
QuintinWillison: If it will definitely fail in future, then YAGNI should apply here.

Though, I'm a little concerned when @ikbalkaya says:

> I wonder if there is a way to validate that all sources are included

Does this mean there's a risk of this somehow failing silently in future? (sorry, I've not looked in depth myself yet)

KacperKluka: I've added a pure Java/Kotlin module and as soon as I've tried to sync the project with Gradle I've got this error
```
Caused by: groovy.lang.MissingPropertyException: Could not get unknown property 'android' for task ':purekotlin:sourcesJar' of type org.gradle.api.tasks.bundling.Jar.
```
So it looks like it fails fast, exactly how we want it to :+1:

ikbalkaya: My assumption was that it would fail silently. If it fails fast, I'm fine with how it is right now 👍

ikbalkaya (approved):

QuintinWillison (approved):

PR #585, opened 2022-01-13
Create connection created worker tests

KacperKluka:
I've added tests for `ConnectionCreatedWorker`. To do that I had to update the mocking library, which introduced a bug for which I've luckily found a workaround.


QuintinWillison (approved):
Relates to: https://github.com/mockk/mockk/issues/777

PR #587, opened 2022-01-14
Add set active trackable worker tests

KacperKluka:
I've added tests for the `SetActiveTrackableWorker`.

While working on this I've also noticed that when the new active trackable doesn't have a destination, the previous destination is not cleared. This means that we can end up with a new active trackable, but destination is still used from previously active trackable. I've created an issue for this https://github.com/ably/ably-asset-tracking-android/issues/588.


KacperKluka:
In order to not forget about removing the destination problem mentioned in the PR description, I've decided to include that change as a part of this PR. This will reduce the risk of forgetting about this as the original issue will soon be closed https://github.com/ably/ably-asset-tracking-android/issues/588 and hopefully will reduce the merge conflicts when the whole refactoring integration branch will be merged.

KacperKluka:
After rebasing the refactoring work on the main branch the destination problem was fixed on the integration branch, therefore https://github.com/ably/ably-asset-tracking-android/pull/587#issuecomment-1018328845 is not relevant anymore :wink:

QuintinWillison (approved):

PR #499, opened 2021-11-03
Publish to maven

ikbalkaya:
This pull request should contain all works done that publishes libraries to Maven central
This article is followed
https://proandroiddev.com/publishing-android-libraries-to-mavencentral-in-2021-8ac9975c3e52

PR #592, opened 2022-01-19
Publish to the MavenCentral

KacperKluka:
Added configuration and workflow for publishing the SDKs to the MavenCentral.
The new workflow has temporarily disabled the version tag checkout in order to test it without merging the whole work to the `main` branch.

I've based this configuration on:
- https://github.com/gradle-nexus/publish-plugin
- https://proandroiddev.com/publishing-android-libraries-to-mavencentral-in-2021-8ac9975c3e52
- https://medium.com/mobile-app-development-publication/upload-to-mavencentral-made-easy-for-android-library-30d2b83af0c7
- https://github.com/ably/ably-asset-tracking-android/pull/499
- https://central.sonatype.org/publish/publish-gradle/

### DIFF
--- a/.github/workflows/publish-maven-central.yml
+++ b/.github/workflows/publish-maven-central.yml
@@ -16,8 +16,9 @@ jobs:
 
     steps:
 
-#    Temporarily disable git branch checkout to test this on other branches than the main branch
-#    - uses: actions/checkout@v2
+    - uses: actions/checkout@v2
+#      Temporarily disable tag checkout to test this workflow. According to checkout docs, if we don't specify any ref then:
+#      "When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, uses the default branch."
 #      with:
 #        ref: v${{ github.event.inputs.version }}
 
--- a/.github/workflows/publish-maven-central.yml
+++ b/.github/workflows/publish-maven-central.yml
@@ -16,9 +16,10 @@ jobs:
 
     steps:
 
-    - uses: actions/checkout@v2
-      with:
-        ref: v${{ github.event.inputs.version }}
+#    Temporarily disable git branch checkout to test this on other branches than the main branch
+#    - uses: actions/checkout@v2
+#      with:
+#        ref: v${{ github.event.inputs.version }}
 
     # Gradle 7 requires Java 11 to run
     - uses: actions/setup-java@v2
@@ -32,8 +33,11 @@ jobs:
         ORG_GRADLE_PROJECT_MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
         ORG_GRADLE_PROJECT_ABLY_API_KEY: ${{ secrets.ABLY_API_KEY }}
         ORG_GRADLE_PROJECT_GOOGLE_MAPS_API_KEY: ${{ secrets.GOOGLE_MAPS_API_KEY }}
-        ORG_GRADLE_PROJECT_GITHUB_ACTOR: ${{ github.actor }}
-        ORG_GRADLE_PROJECT_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        ORG_GRADLE_PROJECT_OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+        ORG_GRADLE_PROJECT_OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
+        ORG_GRADLE_PROJECT_SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
+        ORG_GRADLE_PROJECT_SIGNING_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
+        ORG_GRADLE_PROJECT_SIGNING_KEY: ${{ secrets.SIGNING_KEY_RING_FILE_BASE64 }}
       run: |
-        echo "Publishing to GitHub Packages version ${{ github.event.inputs.version }}..."
-        ./gradlew publish -PpublishTarget=GitHubPackages
+        echo "Publishing to MavenCentral version ${{ github.event.inputs.version }}..."
+        ./gradlew publishToSonatype -PpublishTarget=MavenCentral closeAndReleaseSonatypeStagingRepository
--- a/.github/workflows/publish-maven-central.yml
+++ b/.github/workflows/publish-maven-central.yml
@@ -38,7 +38,7 @@ jobs:
         ORG_GRADLE_PROJECT_OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
         ORG_GRADLE_PROJECT_SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
         ORG_GRADLE_PROJECT_SIGNING_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
-        ORG_GRADLE_PROJECT_SIGNING_KEY: ${{ secrets.SIGNING_KEY_RING_FILE_BASE64 }}
+        ORG_GRADLE_PROJECT_SIGNING_KEY_BASE64: ${{ secrets.SIGNING_KEY_RING_FILE_BASE64 }}
       run: |
         echo "Publishing version ${{ github.event.inputs.version }} to MavenCentral..."
         ./gradlew publishToSonatype -PpublishTarget=MavenCentral closeAndReleaseSonatypeStagingRepository
--- a/.github/workflows/publish-maven-central.yml
+++ b/.github/workflows/publish-maven-central.yml
@@ -39,5 +39,5 @@ jobs:
         ORG_GRADLE_PROJECT_SIGNING_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
         ORG_GRADLE_PROJECT_SIGNING_KEY: ${{ secrets.SIGNING_KEY_RING_FILE_BASE64 }}
       run: |
-        echo "Publishing to MavenCentral version ${{ github.event.inputs.version }}..."
+        echo "Publishing version ${{ github.event.inputs.version }} to MavenCentral..."
         ./gradlew publishToSonatype -PpublishTarget=MavenCentral closeAndReleaseSonatypeStagingRepository
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -34,6 +34,9 @@ jobs:
         ORG_GRADLE_PROJECT_GOOGLE_MAPS_API_KEY: ${{ secrets.GOOGLE_MAPS_API_KEY }}
         ORG_GRADLE_PROJECT_GITHUB_ACTOR: ${{ github.actor }}
         ORG_GRADLE_PROJECT_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        ORG_GRADLE_PROJECT_SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
+        ORG_GRADLE_PROJECT_SIGNING_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
+        ORG_GRADLE_PROJECT_SIGNING_KEY: ${{ secrets.SIGNING_KEY_RING_FILE_BASE64 }}
       run: |
         echo "Publishing version ${{ github.event.inputs.version }} to GitHub Packages..."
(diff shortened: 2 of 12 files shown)


.github/workflows/publish-maven-central.yml
```diff
@@ -0,0 +1,43 @@
...
+      env:
+        ORG_GRADLE_PROJECT_MAPBOX_DOWNLOADS_TOKEN: ${{ secrets.MAPBOX_DOWNLOADS_TOKEN }}
+        ORG_GRADLE_PROJECT_MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
+        ORG_GRADLE_PROJECT_ABLY_API_KEY: ${{ secrets.ABLY_API_KEY }}
+        ORG_GRADLE_PROJECT_GOOGLE_MAPS_API_KEY: ${{ secrets.GOOGLE_MAPS_API_KEY }}
+        ORG_GRADLE_PROJECT_OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+        ORG_GRADLE_PROJECT_OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
+        ORG_GRADLE_PROJECT_SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
+        ORG_GRADLE_PROJECT_SIGNING_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
+        ORG_GRADLE_PROJECT_SIGNING_KEY: ${{ secrets.SIGNING_KEY_RING_FILE_BASE64 }}
```
QuintinWillison, line 40: This is not able to be as simple as this. You need to:

1. Take this string value, run it through a base64 decode and then save the result to a sensible filename representing what it is - I would suggest perhaps `signing.key.gpg` - ready to be used by the Gradle task. I would suggest doing this in a discrete step in the workflow, modelled on [what we do here for AWS Amplify JSON](https://github.com/ably/ably-asset-tracking-android/blob/27b5763968bfdaeea18665c2faa2ba0556535976/.github/workflows/assemble.yml#L19), but with the additional complication that you also need to add base64 decode.
2. Pass that file to [the `signing` plugin](https://docs.gradle.org/current/userguide/signing_plugin.html) using `secretKeyRingFile` - that is, not using `useInMemoryPgpKeys`.

To achieve the base64 encode and copy it to my clipboard to set up the GitHub secret for you, as described here, I used `openssl base64 < signing.key.gpg | pbcopy`. I expect that the reverse will look something like...

Save secret to file (first line in discrete workflow step):

```bash
echo "$SIGNING_KEY_RING_FILE_BASE64" > signing.key.gpg.base64
```

Then decode (second line in discrete workflow step):

```bash
openssl base64 -d -in signing.key.gpg.base64 -out signing.key.gpg
```

publish.gradle
```diff
@@ -48,3 +73,15 @@ afterEvaluate {
         }
     }
 }
+
+if (isPublishingToMavenCentral) {
+    // Sign the artifacts
+    signing {
+        useInMemoryPgpKeys(
```
QuintinWillison, line 80: I've left more detail in the comment I left in the new workflow, but I don't believe we need to be using this in-memory route. We should be using the basic `keyId`, `password` and `secretKeyRingFile` properties.

setup-mavencentral-publishing.gradle
```diff
@@ -0,0 +1,15 @@
+// MavenCentral publishing Sonatype configuration
+
+nexusPublishing {
+    repositories {
+        sonatype {
+            username = findProperty('OSSRH_USERNAME') ?: ""
+            password = findProperty('OSSRH_PASSWORD') ?: ""
+
+            // According to the plugin docs the "nexusUrl" and "snapshotRepositoryUrl" have to be added as we've registered after February 2021.
```
QuintinWillison, line 9: Does that match up with [Joel's comment](https://issues.sonatype.org/browse/OSSRH-68447?focusedCommentId=1120336&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1120336)?

publish.gradle
```diff
@@ -27,15 +30,37 @@ afterEvaluate {
...
+                    licenses {
+                        license {
+                            name = 'Apache License 2.0'
+                            url = 'https://github.com/ably/ably-asset-tracking-android/blob/main/LICENSE'
+                        }
+                    }
+                    scm {
+                        connection = 'scm:git:github.com/ably/ably-asset-tracking-android.git'
+                        developerConnection = 'scm:git:ssh://github.com/ably/ably-asset-tracking-android.git'
+                        url = 'https://github.com/ably/ably-asset-tracking-android/tree/main'
```
QuintinWillison, line 56: The `tree/main` suffix here looks superfluous and overly specific. 🤔

build.gradle
```diff
@@ -45,6 +46,14 @@ allprojects {
     // on incrementing this value, ensure to also increment versionCode in android defaultConfig (also in this file)
     version = '1.0.0-beta.15'
 
+    // Values used to publish the SDK to maven repositories.
+    ext {
+        PUBLISH_GROUP_ID = group
+        PUBLISH_ARTIFACT_ID = project.name // This returns the module/project name, like "publishing-sdk".
+        PUBLISH_VERSION = version
+        PUBLISH_POM_NAME = "${PUBLISH_GROUP_ID}:${PUBLISH_ARTIFACT_ID}"
```
QuintinWillison, line 54: This smells of configuration over convention (counter to what [we like](https://en.wikipedia.org/wiki/Convention_over_configuration))... do we really have to specify this?

.github/workflows/publish-maven-central.yml
```diff
@@ -0,0 +1,43 @@
...
+      run: |
+        echo "Publishing to MavenCentral version ${{ github.event.inputs.version }}..."
```
QuintinWillison, line 42:
```suggestion
        echo "Publishing version ${{ github.event.inputs.version }} to Maven Central..."
```

publish.gradle
```diff
@@ -48,3 +73,15 @@ afterEvaluate {
         }
     }
 }
+
+if (isPublishingToMavenCentral) {
```
QuintinWillison, line 77: This begs a couple of questions for me:

1. Should we not consider signing for all distribution platforms? i.e. is there a reason we can't sign when we release to GitHub Packages?
2. Why is this the only place you're testing `isPublishingToMavenCentral`? I'm now worried that you would release to Maven Central even if `publishTarget` wasn't `MavenCentral` 🤔

QuintinWillison (changes requested):
Lots to like here. Thanks! But, alas, a bit more work to do before we can test this out, please...

build.gradle
KacperKluka: According to the [Sonatype docs](https://central.sonatype.org/publish/requirements/#project-name-description-and-url), we have to specify the `name` in the POM. Also, they suggest that it's a common practice to create the name from group and artifact ID:
```
A common and acceptable practice for name is to assemble it from the coordinates using Maven properties:

<name>${project.groupId}:${project.artifactId}</name>
```

publish.gradle
KacperKluka: I've based this on the [official requirements docs](https://central.sonatype.org/publish/requirements/#scm-information) example 
```xml
<scm>
  <connection>scm:git:git://github.com/simpligility/ossrh-demo.git</connection>
  <developerConnection>scm:git:ssh://github.com:simpligility/ossrh-demo.git</developerConnection>
  <url>http://github.com/simpligility/ossrh-demo/tree/master</url>
</scm>
```
Now that I look at it the `connection` field differs from the official example so I'm going to change it :wink:

setup-mavencentral-publishing.gradle
KacperKluka: Yes, as far as I understand it they must have changed their URLs after February 2021 and the publish plugin we're using uses the old ones by default. That's why we have to manually override them and specify the `https://s01.oss.sonatype.org/` URLs :wink:

publish.gradle
KacperKluka: 1. I think that there are no reasons for us to not sign the artifacts when releasing them to GitHub Packages :thinking: I just wanted to introduce as few changes as possible in the old workflow but I think that signing won't break anything :wink: I'll remove this `if` check and always sign the artifacts.
2. That's because to publish to MavenCentral you need to use a special gradle task, so it won't get published when you run the regular `publish` task :wink: However, I think we should be able to add the check in the [MavenCentral setup script](.github/workflows/publish-maven-central.yml) so it won't get configured if we're not targeting MavenCentral :wink:

.github/workflows/publish-maven-central.yml
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/592/commits/c7a61837b912e7b98d9e1f3a211539de4fabf185

.github/workflows/publish.yml
```diff
@@ -35,5 +35,5 @@ jobs:
         ORG_GRADLE_PROJECT_GITHUB_ACTOR: ${{ github.actor }}
         ORG_GRADLE_PROJECT_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        echo "Publishing version ${{ github.event.inputs.version }}..."
-        ./gradlew publish
+        echo "Publishing to GitHub Packages version ${{ github.event.inputs.version }}..."
```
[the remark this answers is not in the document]
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/592/commits/c7a61837b912e7b98d9e1f3a211539de4fabf185

.github/workflows/publish-maven-central.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+
+    steps:
+
+#    Temporarily disable git branch checkout to test this on other branches than the main branch
```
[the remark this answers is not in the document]
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/592/commits/1c27526167ffd96fd18325597c05ffdc8fedf37b

publish.gradle
KacperKluka: According to the [signing plugin docs](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:in-memory-keys) the in-memory route is the correct one for CI setups :thinking: Is there any reason why we shouldn't use it?

.github/workflows/publish-maven-central.yml
KacperKluka: I think it is as simple as that :sweat_smile: In [one](https://proandroiddev.com/publishing-android-libraries-to-mavencentral-in-2021-8ac9975c3e52) of the linked articles it's stated that:
```
signing.key: the base64 encoded private key you exported earlier
```
Additionally in the [signing plugin docs](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:in-memory-keys) they say that the key should be in the ASCII-Armored format which means that it should be base64 encoded as far as I understand it :thinking: WDYT?

publish.gradle
QuintinWillison: Because it's more complicated than we need. Also, that advice is based - I believe - on assumptions around how those CI systems are setup. I would _not_ use this route, for example, with a static Jenkins instance. However, GitHub runners are ephemeral so I am happy that there is no risk and - thus - would prefer this simpler route, please.

setup-mavencentral-publishing.gradle
QuintinWillison: Cool. It's also going to be pretty binary, I guess, in that it will work or it will not work when we try it. 😁

build.gradle
QuintinWillison: No worries, sounds sensible... I'm just surprised this isn't a default somehow - given we're simply combining fields from this config. 🤷

publish.gradle
KacperKluka: As we've briefly discussed on our call, the current setup seems like the simplest one as the signing plugin expects the key to be a Base64 encoded string (which is exactly what we have in the secrets). Therefore, I've added an explanatory comment about our signing approach https://github.com/ably/ably-asset-tracking-android/pull/592/commits/ef18043ec116473fed3d5faa6c0c82710667a7d8

.github/workflows/publish-maven-central.yml
KacperKluka: As discussed on our call I've added an explanation about the current approach which seems to be the simplest one https://github.com/ably/ably-asset-tracking-android/pull/592/commits/ef18043ec116473fed3d5faa6c0c82710667a7d8

QuintinWillison (approved):

PR #593, opened 2022-01-20
Revert refactoring changes

KacperKluka:
We've mistakenly merged some refactoring work that shouldn't yet be landed, thus we're reverting the changes.


QuintinWillison (approved):
Approved without inspection.

PR #600, opened 2022-01-24
Add timestamps to log statements

KacperKluka:
I've added the timestamp in milliseconds since the epoch at the start of each log statement. Here's an example on how it looks:
```
25-01-22 14:22:23.843: [DefaultMapbox] Stop and close Mapbox
25-01-22 14:22:23.887: [DefaultMapbox] Destroy the MapboxNavigation instance
25-01-22 14:22:23.918: [DefaultMapbox] Change location engine resolution
25-01-22 14:22:24.034: io.ably.lib.transport.ConnectionManager: requestState(): requesting closing; id = rK0Y4LadyZ
```


QuintinWillison (changes requested):
I've suggested to @KacperKluka that we format these as real time to millisecond (`MMM`) precision - e.g. hh:mm:ss.MMM

QuintinWillison (approved):

PR #603, opened 2022-01-26
Add connection ready worker

KacperKluka:
I've added the `ConnectionReadyWorker` with tests. During the refactor I've noticed that we may simplify this worker but it seems to be out of scope of this PR, so I'll create an issue for it https://github.com/ably/ably-asset-tracking-android/issues/604 :wink:


QuintinWillison:
@KacperKluka and I have had quite a long discussion around the testing approach and I have, as a result, pushed https://github.com/ably/ably-asset-tracking-android/pull/603/commits/99ad068a039f1d707806afcd4c236800b74a2019 to illustrate my preferred approach which is a 1:1 relationship between the test function and the operation being tested - in that a single operation will have multiple outcomes verified. Currently the balance is more for a 1:1 relationship between the test function and the outcomes individually, which is OTT in my view and does end up with a lot of repetition and potentially (in time) test execution time overhead.

KacperKluka:

PR #621, opened 2022-02-04
Tweak Maven meta

QuintinWillison:
Looking at [how we appear on Maven Central for v.1.0.0](https://search.maven.org/artifact/com.ably.tracking/publishing-sdk/1.0.0/aar):

Screenshot 2022-02-04 at 06 20 58

I didn't like:

- Having two `Ably` entities under `Developers` (fixed in https://github.com/ably/ably-asset-tracking-android/commit/30c53a73afd3b81280573d700c6076f9eeb6ba2a)
- Having a specific branch name specified under `Source code` (fixed in https://github.com/ably/ably-asset-tracking-android/commit/9c107c232401971c062ff1e346e1d739cb437c69)

I also noticed, from looking at another publisher's POM, that there is a `tag` node in `scm` which allows us to specify accurately the tagged commit which built this release (added in https://github.com/ably/ably-asset-tracking-android/commit/16738436381c0424849d1e56b1666370df524295). Others appear to be doing this wrong, perhaps because [the docs that I could find](https://maven.apache.org/pom.html#SCM) also specify `HEAD` as the value for this node. 🙄 

The other publisher's POM, for comparison, [how it appears on Maven Central](https://search.maven.org/artifact/com.google.firebase/firebase-admin/8.1.0/jar):

Screenshot 2022-02-04 at 06 17 13

Noting that they also have the same issue we had with their product/company (!?) name appearing twice under `Developers`. From inspecting other POMs, I could see that `organization` and `organizationUrl` appear to not be required for a `developer` block.

### DIFF
--- a/publish.gradle
+++ b/publish.gradle
@@ -65,7 +65,7 @@ afterEvaluate {
                     scm {
                         connection = 'scm:git:git://github.com/ably/ably-asset-tracking-android.git'
                         developerConnection = 'scm:git:ssh://github.com/ably/ably-asset-tracking-android.git'
-                        url = 'https://github.com/ably/ably-asset-tracking-android/tree/main'
+                        url = 'https://github.com/ably/ably-asset-tracking-android'
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


QuintinWillison:
The [failing check](https://github.com/ably/ably-asset-tracking-android/runs/5063048271?check_suite_focus=true) is caused by a known flakey test (#623).

AndyNicks (approved):

KacperKluka (approved):

Issue #626, opened 2022-02-08
Send batches of location updates


KacperKluka:
### Challenge
We want to send fewer location messages but still preserve the accuracy of the location updates. In order to still be able to animate the asset position smoothly, we will need to notify the subscriber when will the next batch of locations be sent, so they can use that time to animate the previous locations.

### Suggested solution:
To implement this behaviour we will have to make changes in both SDKs:
#### Publisher
- Keep the frequent location engine updates rate ignoring the resolution requests for the best locations accuracy. This will probably require some changes in the `ResolutionPolicy` implementation but it's still to be decided whether this should be a completely new resolution policy or just a param on the already created `DefaultResolutionPolicy` (I'd suggest the second approach).
- Keep the possibility to modify the location sending resolution in order to limit the number of messages sent. All locations updates that were not sent should be included in the `skippedLocations` list of the location update message. This is how it works already so no changes are required.
- Publish the expected interval after which the new locations batch will be sent (for subscriber-side rendering). We have a debug feature of sending the current trackable resolution which we may use to provide the interval information. However, there's still a chance that the `minimumDisplacement` will be exceeded before the `desiredInterval`. This will result in the messages being sent earlier than the subscriber expects them. Not sure if that will be a problem though.

#### Subscriber
- Use the location updates interval to smoothly animate the courier marker on the map. The bigger part of this should be implemented by the SDK users, the only thing from the SDK side is to expose the location updates interval. However, we will implement this behaviour in our example app to show how it can be done.

PR #627, opened 2022-02-09
Enable setting a constant location engine resolution

KacperKluka:
I've added a method to the `Publisher` builder that allows to specify whether the location engine should have a dynamic resolution calculated by the `ResolutionPolicy`or should it have a constant resolution that won't change throughout the publisher lifetime.

Resolves a part of https://github.com/ably/ably-asset-tracking-android/issues/626

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
@@ -7,6 +7,7 @@ import androidx.annotation.RequiresPermission
 import com.ably.tracking.BuilderConfigurationIncompleteException
 import com.ably.tracking.ConnectionException
 import com.ably.tracking.LocationUpdate
+import com.ably.tracking.Resolution
 import com.ably.tracking.TrackableState
 import com.ably.tracking.connection.ConnectionConfiguration
 import com.ably.tracking.logging.LogHandler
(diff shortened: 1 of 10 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
```diff
@@ -243,6 +244,18 @@ interface Publisher {
          */
         fun rawHistoryDataCallback(callback: (String) -> Unit): Builder
 
+        /**
+         * **OPTIONAL** Enables using a constant location engine resolution.
+         * If the [resolution] is not null then instead of using [ResolutionPolicy] to calculate a dynamic resolution for the location engine
```
QuintinWillison, line 249: This could be clearer, I think. If I understand correctly, this resolution will override any policy coming via the `resolutionPolicy` method on this builder. That needs to be stated explicitly, using the word override. Also, what would happen if I called this method and _then_ called the `resolutionPolicy` method on the builder? I would expect the policy to then replace this, if that makes sense. We need to think about the state of the builder as an atomic whole - it must make intrinsic sense in and of itself.

QuintinWillison (changes requested):

publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
KacperKluka: This is a bit more complicated :sweat_smile: The `ResolutionPolicy` has two methods both called `resolve`. One of them is for resolving the location updates sending resolution (let's call it Ably resolution) and the other one calculates the location engine resolution. With this new builder method, we are only "overriding" the location engine resolution resolve method. So I don't think that the order of the builder method calls should matter here :thinking:

QuintinWillison: ok, fair enough, so the two concepts can (and must) co-exist at the same time on any given builder instance.

If that ☝️ statement is correct then please feel free to resolve this conversation on my behalf.

QuintinWillison (approved):

AndyNicks (approved):

PR #628, opened 2022-02-09
Send resolution data from publisher

KacperKluka:
I've used the already implemented debug capability of sending resolutions. I've made it enabled by default and added a check that will only perform resolution related actions if the new resolution is different than the current one.

Resolves a part of https://github.com/ably/ably-asset-tracking-android/issues/626

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
@@ -215,10 +215,8 @@ interface Publisher {
         fun rawLocations(enabled: Boolean): Builder
 
         /**
-         * EXPERIMENTAL API
-         * **OPTIONAL** Enables sending of calculated resolutions. This should only be enabled for diagnostics.
-         * In the production environment this should be always disabled.
-         * By default this is disabled.
+         * **OPTIONAL** Enables sending of calculated resolutions.
+         * By default this is enabled.
          *
          * @param enabled Whether the sending of calculated resolutions is enabled.
          * @return A new instance of the builder with this property changed.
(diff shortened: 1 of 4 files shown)


AndyNicks (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
```diff
@@ -215,10 +215,8 @@ interface Publisher {
         fun rawLocations(enabled: Boolean): Builder
 
         /**
-         * EXPERIMENTAL API
-         * **OPTIONAL** Enables sending of calculated resolutions. This should only be enabled for diagnostics.
-         * In the production environment this should be always disabled.
-         * By default this is disabled.
+         * **OPTIONAL** Enables sending of calculated resolutions.
+         * By default this is enabled.
```
QuintinWillison, line 219: This is quite a change to behaviour. This is a good example of a change where we need to 'remember' (collectively) to ensure this is highlighted as a (potentially breaking?) change to runtime behaviour. Impacts:

1. It should appear in changelog as such for next release
2. Next release should perhaps have more than a `patch` revision bump (I know we're planning to make this part of a `minor` bump for next release, perhaps)

cc @AndyNicks @stmoreau 

A while back @tbedford described the way that one the engineering teams he's worked with elsewhere covered this sort of scenario, and I wonder if it's something we should consider adopting as an SDK Team... the basic mechanism was that when a change was made that was changelog-worthy for the next release then the changelog was modified as an atomic part of that change (i.e. within the same PR). I like that, a lot. We would need to define a standard for how we do this as a team, so I'm not proposing this as a change for this PR, but just flagging it here in this context for awareness. I'll create a cross-SDKs issue for this and will reference it in this conversation shortly.

QuintinWillison (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
QuintinWillison: https://github.com/ably/engineering/issues/17

PR #640, opened 2022-02-16
Send publisher resolution data via presence data

KacperKluka:
Instead of sending the resolution data as separate messages on the channel we should send them as the publisher presence data. I've also added a test that verifies that we send publisher resolution data.


AndyNicks (approved):

ikbalkaya (approved):
LGTM

PR #650, opened 2022-02-22
CI build reports

QuintinWillison:
Expose more information as artifacts, in particular for us to use to debug when CI runs have failed.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -33,48 +33,56 @@ jobs:
         if: always()
 
       - uses: actions/upload-artifact@v2
+        name: Build Reports for common
         if: always()
         with:
           name: api-level-${{ matrix.api-level }}-common-build-reports
           path: common/build/reports
 
       - uses: actions/upload-artifact@v2
+        name: Build Reports for integration-testing-app
         if: always()
         with:
           name: api-level-${{ matrix.api-level }}-integration-testing-app-build-reports
           path: integration-testing-app/build/reports
 
       - uses: actions/upload-artifact@v2
+        name: Build Reports for publishing-example-app
         if: always()
         with:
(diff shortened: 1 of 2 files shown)


KacperKluka (commented):
Overall it looks good but I'm wondering why you have chosen those specific modules for `check` and `emulate` workflows? :thinking: For example, in the `emulate` workflow we're uploading reports for the `common` and `core-sdk` which do not contain any integration tests on their own :thinking:
Also, the new `ui-sdk` module seems to be missing in the `check` workflow :wink:

ikurek (approved):
Looks good to me 🙂

ikbalkaya (approved):
LGTM 👍

QuintinWillison:
@KacperKluka Regarding this question:

> I'm wondering why you have chosen those specific modules for check and emulate workflows?

This was because I initially prototyped this PR with a `find` command in the workflow to reveal for me all folder structures that contained reports (`index.html` was a good search term). I didn't look to see if those reports actually contained any contents. Honestly, I don't think it matters too much and I would rather catch more than less, especially in case tests do end up getting added to those other modules. 🤷 

Regarding:

> Also, the new ui-sdk module seems to be missing in the check workflow

I'll take a look, as I also did a `find` in the CI context for that workflow and don't recall surfacing anything for that module. Or I forgot. Or I made a mistake. 🤔

KacperKluka (approved):
@QuintinWillison Thanks for the explanation, LGTM :+1:

QuintinWillison:
I'm going to leave the `find` commands in the workflows for the time being (https://github.com/ably/ably-asset-tracking-android/pull/650/commits/95f5e7523bc15c055cc94dd6e9a569b99a55e0ce). They're harmless and might assist later on if we find any reports are missing.

PR #652, opened 2022-02-23
Upgrade Mapbox to version 2.3.0 and other dependencies to the newest versions

KacperKluka:
I've upgraded the Mapbox Navigation SDK to the `2.3.0` version and the Mapbox Maps SDK to the `10.3.0` version.

After upgrading, I've started to see weird lint errors and it turns out that the way to fix them was to update the remaining libraries. This included:
- Gradle
- Kotlin
- ktlint (I've also removed [this](https://github.com/ably/ably-asset-tracking-android/pull/652/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7L232-L233) which I believe is not required anymore :wink: )
- Android libraries
- Testing libraries

Finally, I had to explicitly add the `fragment` dependency in version `1.4.1` as this was required to obtain the `getFragment()` method on the new `FragmentContainerView` object which lint told us to use :man_shrugging: :


AndyNicks (approved):

QuintinWillison (approved):
Now, that's what I call a spring clean! 😋

PR #656, opened 2022-03-15
Add the WorkerFactory for easier Worker creation

KacperKluka:
I've added the `WorkerFactory` that's responsible for creating all the workers. Thanks to this we can hide the implementation details like usage of the Ably wrapper or log handler when creating a worker.
This is a groundwork for the upcoming PRs that will replace the event queue with the new worker queue.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/WorkResultHandler.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/WorkResultHandler.kt
@@ -1,6 +1,7 @@
 package com.ably.tracking.publisher.workerqueue.resulthandlers
 
 import com.ably.tracking.publisher.CorePublisher
+import com.ably.tracking.publisher.workerqueue.WorkerFactory
 import com.ably.tracking.publisher.workerqueue.results.WorkResult
 import com.ably.tracking.publisher.workerqueue.workers.Worker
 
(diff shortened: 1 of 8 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/WorkResultHandler.kt
```diff
@@ -17,10 +18,11 @@ internal interface WorkResultHandler<in T : WorkResult> {
...
      * @param corePublisher: This is a temporary reference of [CorePublisher] that is kept here to maintain
      * compatibility with refactored code.
+     * @param workerFactory: A [WorkerFactory] that will create the optional [Worker].
      *
      * Implementors must delegate work to [corePublisher]  if the required [Worker]s have not been implemented yet.
      *
      * @return an optional [Worker] if implementors decide there is a need to add another worker to the queue.
      * **/
-    fun handle(workResult: T, corePublisher: CorePublisher): Worker?
+    fun handle(workResult: T, corePublisher: CorePublisher, workerFactory: WorkerFactory): Worker?
```
ikbalkaya, line 27: Instead of passing workerFactories to `handle() `methods wouldn't it be cleaner to pass it as object dependency. Also considering the `handle` returns a` Worker?` it won't always need a factory to create one

ikbalkaya (changes requested):
Generally looks good. The only thing from my side : `WorkResultHandler`'s `handle` can do without a factory if there is not a strong reason to keep it there

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/resulthandlers/WorkResultHandler.kt
KacperKluka: Great idea :grin: https://github.com/ably/ably-asset-tracking-android/pull/656/commits/33f6a01e7c5f30f0fe8fe88d35274979e9710042

ikbalkaya (approved):
Looks good 😎

PR #657, opened 2022-03-15
Move log statements from the event queue to the workers

KacperKluka:
I've moved log statements from the `CorePublisher` to the appropriate `Worker`s in preparation for the upcoming PR that will replace the event queue with the worker queue.


ikbalkaya (approved):
Looks good 👍 An observation from me is that our dependencies get deeper and more complicated. It might arguably be a good idea to adopt a DI framework 🤔

KacperKluka:

PR #658, opened 2022-03-15
Connect add trackable process workers

KacperKluka:
The add trackable process was previously represented by 3 events and now is represented by 3 workers. In order to correctly chain them together I had to add some callbacks that are used in the latter workers to the prior ones. This is a groundwork for the upcoming work that will replace the event queue with the worker queue.


KacperKluka:

PR #607, opened 2022-01-27
Integration: Refactor event queue

KacperKluka:
This is an aggregate PR of the whole event queue refactoring work. All refactoring PRs should be landed on `integration/refactor-event-queue` branch. After all the work is done, this PR will be landed on the `main` branch and the refactoring work will be completed. Until that, the PR should stay in the "draft state".
Additionally, this PR will show the state of tests for the refactoring work.

PR #661, opened 2022-03-23
Add timeout to stopping the publisher

KacperKluka:
I've added a configurable timeout to the Publisher's `stop()` method. 
Also I've changed the behaviour of the `StopWorker`related to setting the `isStopped` flag. Previously it was only set if the stopping was successful. Now, it is set no matter if the stopping procedure was completed successfully or there was an error. This makes sense as after calling the `stop()` method we should treat the Publisher as stopped.

We agreed that all of the refactoring work should be on https://github.com/ably/ably-asset-tracking-android/pull/607 but since this is an API change I think it deserves its own PR :wink:


ikbalkaya (approved):
Looks good to me

QuintinWillison (approved):

PR #664, opened 2022-04-04
Improve add trackable process

KacperKluka:
Previously, the add operation was failing if there was an error when subscribing to presence messages (after successfully entering the presence).
Now, the add operation won't fail but will return a new state which indicates that the trackable is online but is not used optimally (due to lack of presence data). In the meantime the SDK will listen for the channel state changes and will try to resubscribe to presence data each time the channel becomes online. If it succeeds in subscribing to presence the state of the trackable will be updated to notify that it is now using all of its potential.

### DIFF
--- a/core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
+++ b/core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
@@ -11,14 +11,15 @@ sealed class TrackableState {
     object Online : TrackableState()
 
     /**
-     * Trackable state is [Publishing] when it's being actively tracked but is neither able to detect nor receive data from subscribers.
+     * Trackable state is [Publishing] when its locations are being published but it is not able to detect subscribers or receive data from them.
+     * This state allows the trackable to be actively tracked, however, its features are limited compared to the [Online] state.
      * This state can change to either [Online] or [Offline] or [Failed].
      */
     object Publishing : TrackableState()
 
     /**
-     * Trackable state is [Offline] when it's connecting or recovering from an error and hopefully will soon be back in the [Online].
-     * This state can change to either [Online] or [Failed].
+     * Trackable state is [Offline] when it's connecting or recovering from an error and hopefully will soon be back in the [Online] or [Publishing].
+     * This state can change to either [Online] or [Publishing] or [Failed].
      */
     data class Offline(val errorInformation: ErrorInformation? = null) : TrackableState()
 
--- a/core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
+++ b/core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
@@ -11,10 +11,10 @@ sealed class TrackableState {
     object Online : TrackableState()
 
     /**
-     * Trackable state is [OnlineNonOptimal] when it's being actively tracked but is neither able to detect nor receive data from subscribers.
+     * Trackable state is [Publishing] when it's being actively tracked but is neither able to detect nor receive data from subscribers.
      * This state can change to either [Online] or [Offline] or [Failed].
      */
-    object OnlineNonOptimal : TrackableState()
+    object Publishing : TrackableState()
 
     /**
      * Trackable state is [Offline] when it's connecting or recovering from an error and hopefully will soon be back in the [Online].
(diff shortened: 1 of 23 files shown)


core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
```diff
@@ -5,17 +5,26 @@ package com.ably.tracking
...
+     * This state can change to either [Offline] or [Failed].
      */
     object Online : TrackableState()
 
     /**
-     * Trackable state is [Offline] when it's connecting or recovering from an error and hopefully will soon be back in the [Online]. This state can change to either [Online] or [Failed].
+     * Trackable state is [OnlineNonOptimal] when it's being actively tracked but is neither able to detect nor receive data from subscribers.
+     * This state can change to either [Online] or [Offline] or [Failed].
+     */
+    object OnlineNonOptimal : TrackableState()
```
ikbalkaya, line 17: I think this name should change to something else. `NonOptiomal` immediately sounds like to me, that it's working but somehow with low frequency of data etc. How about `OnlineNotActive` or `OnlineNonReceiving`?

ikbalkaya (changes requested):
Not a very detailed review, but it looks good, except for a naming suggestion

core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
KacperKluka: That's a good suggestion, thanks :bow: How about being very explicit and changing it to `OnlineNotSubscribed`? :thinking: @AndyNicks @ikbalkaya WDYT

core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
```diff
@@ -5,17 +5,26 @@ package com.ably.tracking
...
     /**
-     * Trackable state is [Online] when it's being actively tracked. This state can change to either [Offline] or [Failed].
+     * Trackable state is [Online] when it's being actively tracked.
+     * This state can change to either [Offline] or [Failed].
      */
     object Online : TrackableState()
 
     /**
-     * Trackable state is [Offline] when it's connecting or recovering from an error and hopefully will soon be back in the [Online]. This state can change to either [Online] or [Failed].
+     * Trackable state is [OnlineNonOptimal] when it's being actively tracked but is neither able to detect nor receive data from subscribers.
```
QuintinWillison, line 14: I'm not sure that the use of the word 'neither' is correct here, as my understanding is that this state is seen when transitioning from offline to online, therefore transitory in nature also.

core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
QuintinWillison: Perhaps we don't try to overload 'online' anymore and instead go for a single word that represents what it's doing in this transitory state - I'm thinking perhaps either `Publishing` or `Presenting`.

Then I would update the commentary for the `Online` state to make it clear that it includes everything expected of `Presenting`.

I also think it might be worth documenting the possible state transitions, perhaps by indicate on each state's commentary whether it's final and, if not, what state(s) could follow it.

QuintinWillison (changes requested):
I've had a fairly brief skim through (for lack of time) and it all looks sensible to me... just a need, as identified by @ikbalkaya, to perhaps rename that state.

core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
KacperKluka: I like the `Publishing` name as it describes exactly what we`re doing - we are publishing this trackable's locations, nothing more, nothing less :wink:
Here I've renamed the state https://github.com/ably/ably-asset-tracking-android/pull/664/commits/6ba46df352155c8bb1c8cba1ad2c465b09dca0f3 and here I've updated the docs https://github.com/ably/ably-asset-tracking-android/pull/664/commits/eae1f34149040989a02ceb47ec6e5cb2ef7f4b48

core-sdk/src/main/java/com/ably/tracking/ConfigurationModels.kt
KacperKluka: I've changed the docs to make them clearer https://github.com/ably/ably-asset-tracking-android/pull/664/commits/eae1f34149040989a02ceb47ec6e5cb2ef7f4b48 :wink:

QuintinWillison (approved):

ikbalkaya (approved):

PR #665, opened 2022-04-05
Handle connection resume exception

KacperKluka:
From time to time we might get the "Connection resume failed" exception from `ably-android` due to the fact that a new connection was established. Because of that, all messages that we were trying to send using the old connection will receive the error. This should not be propagated immediately to the SDK users and it should be handled on the AAT side.

Therefore, I've wrapped all the Ably SDK calls that might throw this exception in a method that will retry those calls if a "connection resume failed" error is received. https://github.com/ably/ably-asset-tracking-android/pull/665/commits/cfadcae2fb075295f72d9fe769498ffd31a1235a
To do that in a clean way I had to first refactor the `Ably` wrapper to use coroutines (but still have the callback-based API). https://github.com/ably/ably-asset-tracking-android/pull/665/commits/72ce740fdbbd0330813f44d1a3e668113bae9022

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -318,37 +318,65 @@ constructor(
     }
 
     override fun disconnect(trackableId: String, presenceData: PresenceData, callback: (Result<Unit>) -> Unit) {
-        if (channels.contains(trackableId)) {
-            val channelToRemove = channels[trackableId]!!
+        scope.launch {
+            if (channels.contains(trackableId)) {
+                val channelToRemove = channels[trackableId]!!
+                try {
+                    disconnectChannel(channelToRemove, presenceData)
+                    channels.remove(trackableId)
+                    callback(Result.success(Unit))
+                } catch (exception: ConnectionException) {
+                    callback(Result.failure(exception))
+                }
+            } else {
+                callback(Result.success(Unit))
+            }
+        }
+    }
+
+    private suspend fun disconnectChannel(channel: Channel, presenceData: PresenceData) {
+        leavePresence(channel, presenceData)
+        channel.unsubscribe()
+        channel.presence.unsubscribe()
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -546,4 +611,57 @@ constructor(
             ably.close()
         }
     }
+
+    /**
+     * Performs the [operation] and if a "connection resume" exception is thrown it waits for the [channel] to
+     * reconnect and retries the [operation], otherwise it rethrows the exception. If the [operation] fails for
+     * the second time the exception is rethrown no matter if it was the "connection resume" exception or not.
+     */
+    private suspend fun retryIfChannelConnectionFails(channel: Channel, operation: suspend () -> Unit) {
```
QuintinWillison, line 620: A couple of thoughts here:

1. Perhaps this method could have a better name, including resume, so perhaps `retryChannelOperationIfConnectionResumeFails`.
2. Could you reduce potential for the `operation` to be doing something different (e.g. on a different channel) by making the `operation` take the `Channel` as an argument? Looking at the upstream call sites, you're having to pass the `channel` variable twice - once to this first arg and then for a second time within the `operation` implementation.

QuintinWillison (changes requested):

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: Good suggestions :+1: 
1. https://github.com/ably/ably-asset-tracking-android/pull/665/commits/02542e29ea43b97d28e1124b9afdf9efb04d535c
2. https://github.com/ably/ably-asset-tracking-android/pull/665/commits/0336a82cb3aa465563fbb9ce1210901ba21ab177

QuintinWillison (approved):

ikbalkaya (approved):
Warning, light review! LGTM

PR #671, opened 2022-05-24
Add prefix to secret names, as used for their instances at org-level.

QuintinWillison:
Matching changes made in response to the creation of https://github.com/ably/ably-java/pull/791, creating the need for Maven signing and OSSRH secrets to be used by more than one repository, meaning they had to be moved to org-level.


KacperKluka (approved):

ikbalkaya (approved):
LGTM 👍

PR #678, opened 2022-06-05
Do not destroy the Mapbox instance when removing the last trackable

KacperKluka:
Previously, when you removed the last trackable from the publisher the whole instance of Mapbox was being closed and destroyed. Now, it only stops the trip and location updates instead. Stopping the publisher still should both stop the trip and close the Mapbox instance.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -432,7 +432,7 @@ constructor(
     override fun stopLocationUpdates(properties: PublisherProperties) {
         properties.isTracking = false
         mapbox.unregisterLocationObserver()
-        mapbox.stopAndClose()
+        mapbox.stopTrip()
     }
 
     @RequiresPermission(anyOf = [Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION])
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -66,6 +66,7 @@ internal interface CorePublisher {
     fun removeCurrentDestination(properties: PublisherProperties)
     fun startLocationUpdates(properties: PublisherProperties)
     fun stopLocationUpdates(properties: PublisherProperties)
+    fun closeMapbox()
     fun processNextWaitingEnhancedLocationUpdate(properties: PublisherProperties, trackableId: String)
     fun saveEnhancedLocationForFurtherSending(properties: PublisherProperties, trackableId: String, location: Location)
     fun retrySendingEnhancedLocation(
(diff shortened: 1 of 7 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -179,6 +180,10 @@ constructor(
         workerFactory = DefaultWorkerFactory(ably, hooks, this, policy, mapbox, this, logHandler)
         workerQueue = DefaultWorkerQueue(properties, scope, workerFactory)
         ably.subscribeForAblyStateChange { enqueue(workerFactory.createWorker(WorkerParams.AblyConnectionStateChange(it))) }
+        mapbox.setLocationHistoryListener { historyData -> scope.launch { _locationHistory.emit(historyData) } }
```
ikbalkaya, line 183: Is this change related to this  🤔

publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
```diff
@@ -249,13 +254,20 @@ internal class DefaultMapbox(
...
-            logHandler?.v("$TAG Stop and close Mapbox")
+            logHandler?.v("$TAG Stop trip and location updates")
             mapboxReplayer?.stop()
+            mapboxReplayer?.seekTo(0.0)
+            mapboxNavigation.stopTripSession()
+        }
+    }
+
+    override fun close() {
+        runBlocking(mainDispatcher) {
```
ikbalkaya, line 267: I wonder why this is blocking on main dispatcher? Wouldn't it cause ANR on Android - or is it a case that Mapbox running on UI thread ?

ikbalkaya (commented):
Code looks good. I have some minor questions

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
KacperKluka: Yes, but I agree that Github chose to display it in a weird way :sweat_smile: I moved code from the middle of this block and put it in the `registerLocationObserver()` method directly below this one. :wink: So something like this:
**Before**
```
init {
    doSomething()
    codeToExtract()
    doSomething()
}
```
**After**
```
init {
    doSomething()
    doSomething()
}

private fun newMethod() {
    codeToExtract()
}
```

publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
KacperKluka: We have to use the main dispatcher because the `MapboxNavigation` is annotated by `@UiThread` which 
> Denotes that the annotated method or constructor should only be called on the UI thread.

It should be safe from ANRs, as we're using coroutines so they suspend the thread but do not block it :wink:

ikbalkaya: [Here](https://proandroiddev.com/how-to-deadlock-your-android-app-with-runblocking-8dab02c2624d#:~:text=This%20is%20why%20runBlocking%20with,t%20start%20the%20targeted%20code.) It says

>  if you write this code on the UI thread, you will deadlock your Android app forever.

KacperKluka: Oh right, true :sweat_smile: However, it has been implemented like this for a long time now and everything seems to be fine, so I think it shouldn't block this PR :thinking: Nevertheless, this is a very valid point and I've created an issue to discuss it in the future https://github.com/ably/ably-asset-tracking-android/issues/680 :bow:

ikbalkaya (approved):

PR #690, opened 2022-06-15
Remove runtime secrets from workflows that don't need them

KacperKluka:
I removed the runtime secrets (Ably key, Mapbox access key, Google maps key) from workflows that don't need them. To do that, I had to make them optional in gradle configuration files.


QuintinWillison (approved):
Ooh, very nice. A lovely incremental improvement. Great work! 🚀

PR #692, opened 2022-06-17
Add location updates validation

KacperKluka:
It turns out that the gson is not very secure and can break Kotlin's null safety. I guess that's because it uses the reflection mechanism from Java and therefore is allowed to put a null in a non-nullable field. Because of that, I've added a few checks that make sure that the objects we receive are in a valid format. If that's not the case we are simply logging it to the log handler.


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -494,7 +499,12 @@ constructor(
         channels[trackableId]?.let { channel ->
             try {
                 channel.subscribe(EventNames.RAW) { message ->
-                    listener(message.getRawLocationUpdate(gson))
+                    val locationUpdate = message.getRawLocationUpdate(gson)
+                    if (locationUpdate != null) {
+                        listener(locationUpdate)
+                    } else {
+                        logHandler?.w("Could not deserialize a raw location update message")
```
ikbalkaya, line 506: This is probably a product question but I am wondering how risky this would be just to log a warning

common/src/main/java/com/ably/tracking/common/GsonHelpers.kt
```diff
@@ -0,0 +1,14 @@
+package com.ably.tracking.common
+
+import com.google.gson.Gson
+import com.google.gson.JsonParseException
+
+/**
+ * Deserializes the [json] and returns an object of type [classOfT] or a null if something goes wrong.
+ */
+fun <T> Gson.fromJsonOrNull(json: String?, classOfT: Class<T>): T? =
```
ikbalkaya, line 9: As this function already returns an optional type, I think 'OrNull' part could be dropped.

ikbalkaya (changes requested):
Minor comments

common/src/main/java/com/ably/tracking/common/GsonHelpers.kt
KacperKluka: The problem is that there's already a function called `fromJson()` that does not return a null but throws an exception in case anything goes wrong :sweat_smile: That's why I've added the `OrNull` suffix to this method :wink:

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: After a discussion on an [internal channel](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1656498487121459?thread_ts=1656331319.885969&cid=C01EPJENRD0) we've decided that an invalid message is a protocol error and should change the trackable state to `Failed`. I'll adjust this PR accordingly :wink:

KacperKluka:
Changed to draft as we have to wait for an infrastructure for assigning error codes to complete this PR.

KacperKluka:
The [error code](https://github.com/ably/ably-asset-tracking-common/tree/main/specification#error-codes) is ready so we can continue the work.

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/692/commits/58cab833d4a88c94d2c98d419ca3a681be7a1379

KacperKluka: I've also added a test that checks the failure of the subscriber channel :wink: https://github.com/ably/ably-asset-tracking-android/pull/692/commits/992b1bd47053e0877003104bfc6bcdc836c7f13c

davyskiba (approved):
LGTM

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -515,30 +522,73 @@ constructor(
...
+    private fun processReceivedLocationUpdateMessage(
+        channel: Channel,
+        trackableId: String,
+        presenceData: PresenceData,
+        message: Message,
+        listener: (LocationUpdate) -> Unit,
+        isRawLocation: Boolean
+    ) {
+        val locationUpdate =
+            if (isRawLocation) message.getRawLocationUpdate(gson) else message.getEnhancedLocationUpdate(gson)
```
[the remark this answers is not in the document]
KacperKluka: I generally prefer to use `{}` as well but here I treat this `if(a) x else y` as a ternary operator `a ? x : y` and would prefer to keep it in one line, would that be okay with you :wink:

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -515,30 +522,73 @@ constructor(
...
+        if (locationUpdate != null) {
+            listener(locationUpdate)
+        } else {
+            logHandler?.e("Could not deserialize ${if (isRawLocation) "raw" else "enhanced"} location update message")
```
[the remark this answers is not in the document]
KacperKluka: Good idea :+1: https://github.com/ably/ably-asset-tracking-android/pull/692/commits/1de5741f8114495b3c880733ac20d30bce4dc048

davyskiba (approved):

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: Because Ikbal is away and we have to merge this PR I'm going to resolve this conversation. Please feel free to let me know if you are not happy with the PR and would like to add some changes, I'll add them in a separate PR :bow:

KacperKluka:
I'm going to use my admin powers to merge this PR despite the change requests. However, those requests seem to be fixed so I hope it is okay. Let me know if you feel otherwise and we'll work on that in another PR :wink:

PR #693, opened 2022-06-17
Update ably-android version to 1.2.14

KacperKluka:
Updated the `ably-android` version which should fix its transitive dependencies' vulnerabilities


QuintinWillison (approved):
Good call with https://github.com/ably/ably-asset-tracking-android/pull/693/commits/0a17b5b64664a26bba2088e66abe87a44a5b725c - thanks, @KacperKluka. 😉

PR #699, opened 2022-06-23
Fix exception throwing issues when connecting to Ably

KacperKluka:
I've noticed that when the connection operation fails (which is either when starting the subscriber or adding a trackable to the publisher) instead of throwing a `ConnectionException` we are throwing an `Exception` that holds a `ConnectionException`. Because of that, our `try / catch` was not catching the error and it caused the system to malfunction.


KacperKluka:
Converting to draft as I've found more issues related to this piece of code and exceptions.

PR #703, opened 2022-06-24
Request a new auth token if connecting to Ably fails due to wrong token capabilities

KacperKluka:
When a new trackable is added and a connection to Ably is being established, we are entering the presence of a channel. If that operation fails due to the token capabilities, we request a new token and retry the operation. If it fails for the second time we treat that as an error.


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -254,6 +255,29 @@ constructor(
...
+            channel.enterPresenceSuspending(presenceData)
+        } catch (connectionException: ConnectionException) {
+            if (connectionException.errorInformation.code == AUTH_TOKEN_CAPABILITY_ERROR_CODE) {
+                try {
+                    ably.auth.renew()
+                } catch (ablyException: AblyException) {
+                    throw ablyException.errorInfo.toTrackingException()
+                }
+                channel.attachSuspending()
+                channel.enterPresenceSuspending(presenceData)
```
QuintinWillison, line 274: This could be some time later, at which point this `presenceData` could have been superseded by another call to `enterChannelPresence`, so I'm wondering if this solution might be too simplistic.

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -254,6 +255,29 @@ constructor(
...
+     * a new auth token is requested and the operation is retried once more.
+     * @throws ConnectionException if something goes wrong or the retry fails
+     */
+    private suspend fun enterChannelPresence(channel: Channel, presenceData: PresenceData) {
+        try {
+            channel.enterPresenceSuspending(presenceData)
+        } catch (connectionException: ConnectionException) {
+            if (connectionException.errorInformation.code == AUTH_TOKEN_CAPABILITY_ERROR_CODE) {
+                try {
+                    ably.auth.renew()
```
QuintinWillison, line 269: I just had a quick check of the `ably-android` source and I see that `AblyRealtime`'s implementation of `onAuthUpdated(string, boolean)` [is called](https://github.com/ably/ably-java/blob/faa89afa8b6dc4fb535a6f488e3b776c4a0ab8a7/lib/src/main/java/io/ably/lib/rest/Auth.java#L833) by explicitly setting that second argument (`waitForResponse`) to `false`. I believe that means that this method call looks blocking but is actually just setting off an asynchronous operation, which I expect is unlikely to complete before you call `channel.attachSuspending()` below. Therefore, I am wondering if this could ever work reliably, or at all. 🤔

QuintinWillison (changes requested):
I'm not sure this would work.

Perhaps we should work on finding a way to test this. Perhaps with a local token auth implementation within the test suite. 🤔

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: I've been heavily testing it manually and it worked each time :grin: The method `ably.auth.renew()` is actually blocking as it calls `assertValidToken()` which calls `requestToken()` which calls the `authCallback.getTokenRequest()`. The last one returns the new token (so it makes a blocking call to the user's token provider) and therefore after this method is executed we already have the new token :wink:

KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/703/commits/717be7121fb2eb760505033c3e054126e4824d9a

ikbalkaya: This is indeed an implicit async call and because as @QuintinWillison mentioned `waitForResponse` is false, it is possible for function to return without getting a response

This is where function returns

https://github.com/ably/ably-java/blob/faa89afa8b6dc4fb535a6f488e3b776c4a0ab8a7/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L944

On the next line there is a while loop that checks whether a connection state has changed and in case of `waitForResponse` is true. I will try to see whether I can transfer the state change asynchronously to this library. But if we cannot it might likely be a change request on ably-java

ikbalkaya: As a follow up to this, I wasn't able to locate any spec that enforces this behaviour in [re-authentication](https://docs.ably.io/client-lib-development-guide/features/#RTN22) or [authroriazation](https://docs.ably.io/client-lib-development-guide/features/#RTC8). Therefore I am going to open an issue for ably-java

QuintinWillison: As an update, @ikbalkaya's work on `ably-java` has landed now under https://github.com/ably/ably-java/pull/816. Once `ably-android` v.1.2 patch has been released, then we'll need to bring that in here in order to resolve this work. Or perhaps as a discrete PR whose underlying branch can then become the target for this PR.

QuintinWillison:
Hmm, GitHub has me confused. My intention was to remove @ikbalkaya's soon-to-be-very stale review approval, whilst we keep this PR in a Draft state awaiting the newer `ably-android` dependency. I don't seem to be able to remove that approval so perhaps safest that I re-request that review so the state isn't approved. 😅

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -254,6 +256,45 @@ constructor(
...
+        try {
+            channel.enterPresenceSuspending(presenceData)
+        } catch (connectionException: ConnectionException) {
+            if (connectionException.errorInformation.code == AUTH_TOKEN_CAPABILITY_ERROR_CODE) {
+                val renewAuthResult = renewAuthSuspending()
+
+                renewAuthResult.errorInfo?.let {
+                    throw it.toTrackingException()
+                }
+                Log.d("Ably", Thread.currentThread().name)
```
KacperKluka, line 274: Do we need this log or is it a leftover from debugging? :wink:

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -254,6 +256,45 @@ constructor(
...
+            try {
+                ably.auth.renewAuth { success, tokenDetails, errorInfo ->
+                    continuation.resume(RenewAuthResult(success, tokenDetails, errorInfo))
+                }
+            } catch (e: Exception) {
+                e.printStackTrace()
+                continuation.resumeWithException(e)
+            }
+        }
+    }
```
KacperKluka, line 296: I think we don't need that `RenewAuthResult` as we only care if the operation was successful or not. So we could follow the pattern used in other suspending Ably wrappers, and simplify it to either complete or throw an error if something goes wrong, WDYT? :wink: So instead of 
```kotlin
ably.auth.renewAuth { success, tokenDetails, errorInfo ->
    continuation.resume(RenewAuthResult(success, tokenDetails, errorInfo))
}
```
we could have
```kotlin
ably.auth.renewAuth { success, tokenDetails, errorInfo ->
    if (success) {
        continuation.resume(Unit)
    else {
        continuation.resumeWithException(errorInfo.toTrackingException())
    }
}
```

common/src/main/java/com/ably/tracking/common/Ably.kt
ikbalkaya: Oops, that's a leftover 🤦 0eeabac19456565afdcd6a09337181bd981702b0

common/src/main/java/com/ably/tracking/common/Ably.kt
ikbalkaya: I'm going to have to disagree with this. Even if we do not use results in this particular case, situations might arise where we will need them. Plus this is more like a mapping from the original ably-java function.

KacperKluka: I'm a big fan of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and [KISS](https://en.wikipedia.org/wiki/KISS_principle) rules and according to them, if we do not need something right now, we should not add it because it adds complexity and noise to the code :wink: So I would like to keep the code as simple as possible and to only solve our current needs :wink: WDYT?

ikbalkaya: This particular functions aims to be a suspending version of `renewAuth`. Doesn't it have to fulfill the requirement of `renewAuth`? I do not think this is a matter of YAGNI because then I would have created a function called `renewAuthSuspendingOnlyForCompletion` which IMO more superfluous. I am also not a big fan of cutting small corners that might harm the overall consistency of code.

KacperKluka: Hmm if that's how you interpret this class name then maybe we've chosen the wrong name for it :thinking: Nevertheless, I would like to land this PR so let's leave it as it is for now, as everything works well :+1:

ikbalkaya (approved):

KacperKluka (commented):
As the author of this PR I can only comment, so please treat this comment as my approval of this PR :wink: LGTM :+1:

common/src/main/java/com/ably/tracking/common/Ably.kt
QuintinWillison: https://github.com/ably/ably-asset-tracking-android/pull/703/commits/5c1dc7dedf969d0054067f145568329cbcf7532f

QuintinWillison (approved):

PR #708, opened 2022-07-08
Fix error when trying to connect to a previously closed channel

KacperKluka:
We've encountered an error when we tried to add a trackable, that was already added and removed a while ago. It looks like the issue was happening due to the underlying `channel` being in a `detached` state after its removal. A [simple fix](https://github.com/ably/ably-asset-tracking-android/pull/708/commits/77c26468cc23d0b22d2cfd325d521f3d09e8e488) of checking the state and attaching to the channel before attempting to enter its presence seems to solve the issue.

As a part of this PR, I've changed the way we use Ably SDK to a coroutines-based approach. This was done to escape a callback hell that was slowly being created due to the callback-based Ably API :sweat_smile:


ikbalkaya (approved):
This looks good for now,

However I think in the future we should adopt a bit more reactive way of dealing with state change actions by listening to channel/connection state change.

Issue #711, opened 2022-07-14
Replace ambiguous @JvmSynthetic annotation with a type alias


KacperKluka:
We are using the `@JvmSynthetic` annotation to mark methods in our Kotlin API that should not be visible in the Java version of the API. However, the name of the annotation isn't very meaningful and doesn't describe our use case. What we could do to improve this and make it very explicit is create a `typealias` that would be used instead of that annotation. Something like this
```kotlin
typealias KotlinOnlyApi = JvmSynthetic
```
then we could replace
```kotlin
@JvmSynthetic
suspend fun stop()
```
with
```kotlin
@KotlinOnlyApi
suspend fun stop()
```

The only downside of this is that we can't put the `typealias` in the `common` module and reuse it in both SDKs. Instead, we will have to create the alias in each of the SDK modules (`publishing-sdk` and `subscribing-sdk`) which means code duplication, but in this case that probably won't be a real problem.

QuintinWillison:
This is a great idea, however I might refine the name. The `Api` suffix is, perhaps, superfluous. So, maybe, just `KotlinOnly` is enough. What do you think?

I would also be interested to consider what package this goes under, as that will add context too.

KacperKluka:
It turns out that we might want to take it even further and put this type alias in its own module/project (as suggested [here](https://github.com/ably/ably-asset-tracking-android/pull/716#discussion_r922123757)). Given the above and the fact that this issue is a nice-to-have we're removing it from the `1.1.0` milestone.

PR #716, opened 2022-07-15
Replace @JvmSynthetic with a custom type alias annotation

KacperKluka:
The `@JvmSynthetic` annotation is not very descriptive and does not explain why we are using it. That's why we've introduced a type alias for it, to change its name and make it more meaningful.

### DIFF
--- a/core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
+++ b/core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
@@ -5,6 +5,9 @@ data class ConnectionConfiguration(val authentication: Authentication)
 typealias TokenRequestCallback = (TokenParams) -> TokenRequest
 typealias JwtCallback = (TokenParams) -> String
 
+// We have to duplicate the KotlinOnly annotation from the "common" module as this module does not have access to it
+private typealias KotlinOnly = JvmSynthetic
+
 sealed class Authentication(
     val clientId: String,
     val basicApiKey: String?,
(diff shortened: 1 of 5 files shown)


core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
```diff
@@ -5,6 +5,9 @@ data class ConnectionConfiguration(val authentication: Authentication)
 typealias TokenRequestCallback = (TokenParams) -> TokenRequest
 typealias JwtCallback = (TokenParams) -> String
 
+// We have to duplicate the KotlinOnly annotation from the "common" module as this module does not have access to it
+private typealias KotlinOnly = JvmSynthetic
```
QuintinWillison, line 9: I think we should aim to do this properly, which may mean another project/module to be the canonical root for these annotations.

I'm thinking along the lines of a new repository for "Ably Kotlin Common", with the aspiration that we will be using it soon enough for `ably-java` also.

QuintinWillison (changes requested):
Regarding the `typealias` duplication, I think we need to halt this work for now, as I think to do this properly there's work to be done outside of this repository.

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
KacperKluka: Good idea, I didn't like that duplication as well :sweat_smile: Would it be okay to put it in a separate module in the AAT project for now?

QuintinWillison: I would rather create a new repository for it. Let's discuss in Slack the naming and then I shall share it here.

ikbalkaya:
This will be addressed elsewhere https://github.com/ably/ably-asset-tracking-android/issues/711

PR #724, opened 2022-08-03
Show the trip notification when it gets removed after stopping trip session

KacperKluka:
On Android 24 and below the notification is disappearing when the last trackable is removed (which causes Mapbox trip session stop). This should not happen as this notification is likely used for a foreground service as well. To fix that we are manually showing the notification after trip session stops on Android 24 and below.


davyskiba (approved):
LGTM

PR #725, opened 2022-08-04
Add documentation about location animator flows

KacperKluka:
I added information about the frequency of updates from the flows from `CoreLocationAnimator`. Additionally, I added information about optional animator configuration.

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -47,13 +47,30 @@ dependencies {
 Now you should have access to the `LocationAnimator`, `CoreLocationAnimator` and `Position` classes.
 
 To use the animation logic you need to create a new instance of the `CoreLocationAnimator` and use its shared flows to receive both map marker and map camera location updates.
+You can add optional configuration options during `CoreLocationAnimator` creation:
+
+- `intentionalAnimationDelayInMilliseconds` - A constant delay added to the animation duration. It helps to smooth out movement when we receive a location update later than we've expected.
+  The higher value the less realtime the animation becomes.
+- `animationStepsBetweenCameraUpdates` - How often should the camera updates be sent. Setting a higher value might improve UX but if it's too big it can lead to map marker moving out of the screen.
 
 ```kotlin
-val locationAnimator: LocationAnimator = CoreLocationAnimator()
+val locationAnimator: LocationAnimator = CoreLocationAnimator(
+  intentionalAnimationDelayInMilliseconds,
+  animationStepsBetweenCameraUpdates,
+)
+```
+
+The animator exposes two flows, one for map marker (`positionsFlow`) and the other for camera updates (`cameraPositionsFlow`).
+The map marker flow emits 60 positions per second, so you should simply update marker's position each time the flow emits a new position.
+The camera flow emits 1 position when a new camera update is ready, so you can either move the camera there or animate its movement by yourself.
+
+```kotlin
+// 60 positions per second
 locationAnimator.positionsFlow
     .onEach { mapMarkerPosition -> setMapMarkerPosition(mapMarkerPosition) }
     .launchIn(scope)
 
+// 1 position when a camera update is ready
 locationAnimator.cameraPositionsFlow
     .onEach { cameraPosition -> moveCamera(cameraPosition) }
     .launchIn(scope)
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -50,10 +50,10 @@ To use the animation logic you need to create a new instance of the `CoreLocatio
 The `CoreLocationAnimator` class takes the following configuration options:
 
 - `intentionalAnimationDelayInMilliseconds` - A constant delay added to the animation duration. This helps to smooth out movement when a location update is received later than expected due to network latency.
-  The higher value the less realtime the animation becomes.
+  The higher value the less realtime the animation becomes. The default value is 2000 milliseconds.
 - `animationStepsBetweenCameraUpdates` - How many map marker steps need to be animated before a camera update is sent.
-  One step is created for each location from a location update, so this can be thought of as amount of locations between camera updates.
-  Setting a higher value might improve UX but, if set too high, may cause the map marker to move out of the screen.
+  One step is created for each location from a location update, so this can be thought of as the number of locations between camera updates.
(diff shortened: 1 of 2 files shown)


davyskiba (approved):
LGTM

UPDATING.md
```diff
@@ -47,13 +47,30 @@ dependencies {
 Now you should have access to the `LocationAnimator`, `CoreLocationAnimator` and `Position` classes.
 
 To use the animation logic you need to create a new instance of the `CoreLocationAnimator` and use its shared flows to receive both map marker and map camera location updates.
+You can add optional configuration options during `CoreLocationAnimator` creation:
+
+- `intentionalAnimationDelayInMilliseconds` - A constant delay added to the animation duration. It helps to smooth out movement when we receive a location update later than we've expected.
+  The higher value the less realtime the animation becomes.
+- `animationStepsBetweenCameraUpdates` - How often should the camera updates be sent. Setting a higher value might improve UX but if it's too big it can lead to map marker moving out of the screen.
```
owenpearson, line 54:
```suggestion
- `animationStepsBetweenCameraUpdates` - How often should the camera updates be sent. Setting a higher value might improve UX but, if set too high, may cause the map marker to move out of the screen.
```

Also, I think "How often should the camera updates be sent" is vague. Judging by the name 'animationStepsBetweenCameraUpdate' I would assume that this is the number of animation steps the location animator will send between each camera update, however, judging by this description it sounds like it's actually the frequency of the camera updates. If it's the latter we should make it clear what the units are, ie if I set it to `1` does that mean 1 update per second?

UPDATING.md
```diff
@@ -47,13 +47,30 @@ dependencies {
 Now you should have access to the `LocationAnimator`, `CoreLocationAnimator` and `Position` classes.
 
 To use the animation logic you need to create a new instance of the `CoreLocationAnimator` and use its shared flows to receive both map marker and map camera location updates.
+You can add optional configuration options during `CoreLocationAnimator` creation:
+
+- `intentionalAnimationDelayInMilliseconds` - A constant delay added to the animation duration. It helps to smooth out movement when we receive a location update later than we've expected.
```
owenpearson, line 52:
```suggestion
- `intentionalAnimationDelayInMilliseconds` - A constant delay added to the animation duration. This helps to smooth out movement when a location update is received later than expected due to network conditions.
```

UPDATING.md
```diff
@@ -47,13 +47,30 @@ dependencies {
 Now you should have access to the `LocationAnimator`, `CoreLocationAnimator` and `Position` classes.
 
 To use the animation logic you need to create a new instance of the `CoreLocationAnimator` and use its shared flows to receive both map marker and map camera location updates.
+You can add optional configuration options during `CoreLocationAnimator` creation:
```
owenpearson, line 50:
```suggestion
The `CoreLocationAnimator` class takes the following configuration options:
```

UPDATING.md
owenpearson, line 52: Should we also include the default values for both of these options?

UPDATING.md
```diff
@@ -47,13 +47,30 @@ dependencies {
...
+
+```kotlin
+val locationAnimator: LocationAnimator = CoreLocationAnimator(
+  intentionalAnimationDelayInMilliseconds,
+  animationStepsBetweenCameraUpdates,
+)
+```
+
+The animator exposes two flows, one for map marker (`positionsFlow`) and the other for camera updates (`cameraPositionsFlow`).
+The map marker flow emits 60 positions per second, so you should simply update marker's position each time the flow emits a new position.
```
owenpearson, line 64:
```suggestion
The map marker flow emits 60 positions per second, so you should simply update the marker's position each time the flow emits a new position.
```

owenpearson (changes requested):

UPDATING.md
KacperKluka: I was wondering about it but if we do, then each time we change those values in the code we will also have to remember to update them here (and I think we will most likely forget to do that :sweat_smile: ) WDYT? :wink:

UPDATING.md
KacperKluka: Good idea, let me know if now it is more explicit :wink: https://github.com/ably/ably-asset-tracking-android/pull/725/commits/8be9a59fd100e3071515e84de4048d37d6279253

UPDATING.md
owenpearson: I did consider that when I was writing my comment and I think it's a valid point, but if I was a developer reading this documentation for the first time I would definitely want to know what the default values were. Maybe we could add comments next to the defaults values in the code just to remind developers that if we change them we also need to update the docs?

UPDATING.md
```diff
@@ -47,13 +47,32 @@ dependencies {
 Now you should have access to the `LocationAnimator`, `CoreLocationAnimator` and `Position` classes.
 
 To use the animation logic you need to create a new instance of the `CoreLocationAnimator` and use its shared flows to receive both map marker and map camera location updates.
+The `CoreLocationAnimator` class takes the following configuration options:
+
+- `intentionalAnimationDelayInMilliseconds` - A constant delay added to the animation duration. This helps to smooth out movement when a location update is received later than expected due to network latency.
+  The higher value the less realtime the animation becomes.
+- `animationStepsBetweenCameraUpdates` - How many map marker steps need to be animated before a camera update is sent.
+  One step is created for each location from a location update, so this can be thought of as amount of locations between camera updates.
```
owenpearson, line 55:
```suggestion
  One step is created for each location from a location update, so this can be thought of as the number of locations between camera updates.
```

Looks much better 😄  just one small additional suggestion

owenpearson (changes requested):

UPDATING.md
KacperKluka: Fixed as a part of https://github.com/ably/ably-asset-tracking-android/pull/725/commits/45a3511993f21cd0baaf1d818fe87f359dc6e9f7 :grin:

UPDATING.md
KacperKluka: Good idea :+1: https://github.com/ably/ably-asset-tracking-android/pull/725/commits/45a3511993f21cd0baaf1d818fe87f359dc6e9f7

owenpearson (approved):
LGTM 👍

PR #727, opened 2022-08-04
Upgrade Mapbox Nav SDK to 2.7.0

KacperKluka:
I upgraded to the newest version `2.7.0-rc.1` which includes the special asset tracking profile. Therefore, we can stop using the snapshot repository so I've removed anything related to it as well.


davyskiba (approved):
LGTM

Issue #728, opened 2022-08-04
Add option to enable or disable the cycling profile in the publisher example app


KacperKluka:
We should add a new preference to the settings screen that would allow example app users to enable or disable the cycling profile.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2508) by [Unito](https://www.unito.io)

PR #729, opened 2022-08-04
Allow to enable or disable the cycling profile

KacperKluka:
Added a new API that will allow users to enable or disable the cycling profile during publisher creation.
I enabled the profile in the example app for now and created an issue to make it configurable in the future https://github.com/ably/ably-asset-tracking-android/issues/728.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
@@ -160,7 +160,7 @@ private object MapboxInstancesCounter {
 /**
  * The special Mapbox configuration that enables the Asset Tracking Profile (a.k.a. the Cycling Profile).
  */
-private const val ASSET_TRACKING_PROFILE_CONFIGURATION: String = """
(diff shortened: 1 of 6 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
```diff
@@ -189,6 +194,7 @@ internal class DefaultMapbox(
     predictionsEnabled: Boolean,
     private val rawHistoryCallback: ((String) -> Unit)?,
     constantLocationEngineResolution: Resolution?,
+    vehicleProfile: VehicleProfile,
```
paddybyers, line 197: Will we not support both simultaneously? Or is that just setting `BICYCLE` ?

KacperKluka: As far as I understand that special cycling profile can either be enabled or disabled. If it's enabled then the SDK works best for bicycles (using shortcuts and cycling paths). When it's disabled then the SDK works best for cars (and other vehicles using regular roads). So I don't think we can enable both simultaneously :thinking:

davyskiba (approved):
LGTM

PR #731, opened 2022-08-05
Allow to choose the vehicle profile in the example app

KacperKluka:
I added a preference to the setting screen that enable to choose the vehicle profile. Choosing the `Bicycle` profile will enable the cycling profile.
Screenshot from 2022-08-05 12-13-30

### DIFF
--- a/publishing-example-app/src/main/res/values/untranslatable_strings.xml
+++ b/publishing-example-app/src/main/res/values/untranslatable_strings.xml
@@ -14,5 +14,6 @@
   <string name="preferences_constant_resolution_accuracy_key" translatable="false">constant_resolution_accuracy</string>
   <string name="preferences_constant_resolution_desired_interval_key" translatable="false">constant_resolution_desired_interval</string>
   <string name="preferences_constant_resolution_minimum_displacement_key" translatable="false">constant_resolution_minimum_displacement</string>
+  <string name="preferences_vehicle_profile_key" translatable="false">vehicle_profile</string>
 </resources>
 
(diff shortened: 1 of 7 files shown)


publishing-example-app/src/main/res/values/untranslatable_strings.xml
```diff
@@ -14,5 +14,6 @@
   <string name="preferences_constant_resolution_accuracy_key" translatable="false">constant_resolution_accuracy</string>
   <string name="preferences_constant_resolution_desired_interval_key" translatable="false">constant_resolution_desired_interval</string>
   <string name="preferences_constant_resolution_minimum_displacement_key" translatable="false">constant_resolution_minimum_displacement</string>
+  <string name="preferences_vehicle_profile_key" translatable="false">vehicle_profile</string>
```
davyskiba, line 17: Just curious, why keep those keys in strings instead of constants in the code (as accessing those requires context)?

davyskiba (approved):
All good

publishing-example-app/src/main/res/values/untranslatable_strings.xml
KacperKluka: Because then we can use this key in both `xml` and code files, so there's only one source of truth :wink: If we were to use constants in the code there is no easy way (that I'm aware of) to access them from the preferences `xml` file :thinking:

KacperKluka: I have to resolve this conversation to land this PR :wink: Let's continue the discussion here if you'll have some further questions :bow:

PR #745, opened 2022-08-26
Add error handling in the token auth callbacks

KacperKluka:
Added a way to indicate errors in the token auth callbacks together with some docs to explain how it works.

As I was not sure which codes to use for errors I've used the generic 100 000 that's the start of the AAT error code range. Let me know if that's okay.


paddybyers:
I think it's important that the token callback can provide error information which can eventually be visible to the client, so the client knows the reason for failure.  But here you're discarding information in the exception and replacing it with a different message, code, etc.

Can we do something like:

```
/**
 * This exception should be thrown when there is s an error during the token auth flow.
 * When it is thrown the SDK will not retry the auth flow automatically.
 */
class TokenAuthForbiddenException() : TokenAuthException()

private fun TokenAuthException.toAblyException(): AblyException =
    when (this) {
        is TokenAuthForbiddenException ->
            AblyException.fromErrorInfo(ErrorInfo(this.message, 403, 100_000))
        is TokenException ->
            AblyException.fromErrorInfo(ErrorInfo(this.message, 401, 100_000))
    }

```

I'm not so sure we want a `TokenAuthRetryException` because:
- retrying is the default behaviour and I'm not sure we need to provide an additional option with the same behaviour;
- this exception doesn't force a retry to happen, but it allows the operation to be retriable in the future, and I think the name here implies the wrong thing.

KacperKluka:
@paddybyers I've renamed the exceptions https://github.com/ably/ably-asset-tracking-android/pull/745/commits/f0623e9e8383ee8ab5a11cad22d8fc20f834c40c and added a required message parameter to their constructors https://github.com/ably/ably-asset-tracking-android/pull/745/commits/48339f8fd8c3e57abb84a6dae3637edf4c6ca07f.

> The other thing that bothers me, even with the code above, is that the actual exception thrown doesn't get propagated back to the original caller. What if the app has its own hierarchy of exceptions, error codes, etc, and would like to be able to propagate one of those exceptions back to the caller?

I doubt that we can do anything about that, since the Ably SDK is expecting to receive an `AblyException` :thinking: So it looks like we have to map the original exception to that Ably-specific one, right?

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
```diff
@@ -208,8 +208,8 @@ fun io.ably.lib.types.PresenceMessage.getPresenceData(gson: Gson): PresenceData?
  */
 private fun TokenAuthException.toAblyException(): AblyException =
     when (this) {
-        is TokenAuthNoRetryException ->
-            AblyException.fromErrorInfo(ErrorInfo("Auth flow failed and should not be retried", 403, 100_000))
-        is TokenAuthRetryException ->
-            AblyException.fromErrorInfo(ErrorInfo("Auth flow failed but will be retried", 100_000))
+        is TokenAuthForbiddenException ->
```
paddybyers, line 211: on reflection, I'd also be happy with `TokenAuthNonRetriableException`

paddybyers:
> I doubt that we can do anything about that, since the Ably SDK is expecting to receive an AblyException 🤔 So it looks like we have to map the original exception to that Ably-specific one, right?

Well ably-java will (or should) wrap any `Throwable`. I think the explicit `throws AblyException` in the ably-java token callback interface was a mistake. We should (and already do) catch any throwable, but I think we should unconditionally wrap it with the 80019 exception and pass that back.

So that should mean the AAT application ought to be able to throw any of its own exception type, and have that wrapped and passed back to the initiating caller.

So if we're going to support that, the only question then is how that exception would signal that the operation is non-retriable. What's the pattern for that? Have their exception type additionally implement a `NonRetriableTokenError` interface?

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/745/commits/293ab71b5ffc83c521818a6c760fa70b063d9cf7

KacperKluka:
> So if we're going to support that, the only question then is how that exception would signal that the operation is non-retriable. What's the pattern for that? Have their exception type additionally implement a NonRetriableTokenError interface?

That seems like a viable solution :thinking: If their exception will implement this `NonRetriable` interface then all we have to do is 
```java
try {
    requestToken();
} catch (Exception exception) {
    if (exception instanceof NonRetriable){
        // stop automatic retries
    } else {
        // normal error handling
    }
}
```

paddybyers:
I would be ok with this. I suspect @QuintinWillison will also have an opinion

QuintinWillison (changes requested):
For these specific new `ErrorInfo` instances you're generating I think they should have discrete, new error codes which should be documented [centrally](https://github.com/ably/ably-asset-tracking-common/tree/main/specification#error-codes).

core-sdk/src/main/java/com/ably/tracking/Exceptions.kt
```diff
@@ -4,3 +4,20 @@ class BuilderConfigurationIncompleteException :
...
+/**
+ * The base class for all token auth exceptions.
+ */
+sealed class TokenAuthException(message: String) : Exception(message)
+
+/**
+ * This exception should be thrown when there's an error during the token auth flow.
+ * When it is thrown the SDK will retry the auth flow automatically after some time.
+ */
+class NoTokenException(message: String) : TokenAuthException(message)
```
ikbalkaya, line 17: Description says 'there is an error' but name is `NoTokenException`. Can this be renamed to something like `TokenFlowError` or similar?

ikbalkaya (changes requested):
A name change request

KacperKluka:
Blocked until https://github.com/ably/ably-java/issues/829 is resolved and a new `ably-java` version is released.

KacperKluka:
@QuintinWillison here is a PR that adds new error codes https://github.com/ably/ably-asset-tracking-common/pull/61 which I'm using in code here https://github.com/ably/ably-asset-tracking-android/pull/745/commits/0c710223a21fe21e80193c95f476a3425e4ce469 😉

core-sdk/src/main/java/com/ably/tracking/Exceptions.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/745/commits/a2a0f837c33ae9d8c5aaebdc916962d73b9511de

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
QuintinWillison: Resolving this conversation on @paddybyers' behalf.

QuintinWillison (approved):

PR #749, opened 2022-08-31
Fix the waitForChannelReconnection() method implementation

KacperKluka:
Before adding a state change listener we should first check if channel is already attached. That's because the listener won't notify us about the most recent state change that happened before it was added. So we need to check this by ourselves.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -751,6 +751,9 @@ constructor(
         try {
             withTimeout(timeoutInMilliseconds) {
                 suspendCancellableCoroutine<Unit> { continuation ->
+                    if (channel.state == ChannelState.attached) {
+                        continuation.resume(Unit)
+                    }
                     channel.on { channelStateChange ->
                         if (channelStateChange.current == ChannelState.attached) {
                             continuation.resume(Unit)


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -751,6 +751,9 @@ constructor(
         try {
             withTimeout(timeoutInMilliseconds) {
                 suspendCancellableCoroutine<Unit> { continuation ->
+                    if (channel.state == ChannelState.attached) {
+                        continuation.resume(Unit)
+                    }
```
ikbalkaya, line 756: I think it might be worth to avoid code duplication.

ikbalkaya (approved):
I think it might be worth to avoid code duplication but I don't think it's a blocker.

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: Could you provide a suggestion of what you mean? I'd like to remove code duplication as well but I'm not sure what you're talking about :wink:

davyskiba (approved):

common/src/main/java/com/ably/tracking/common/Ably.kt
ikbalkaya: Sorry. I meant that this code block is repeated in two places (here and below)
```
 if (channelStateChange.current == ChannelState.attached) {
                            continuation.resume(Unit)
                        }
```

KacperKluka: Good idea, I've removed the code duplication https://github.com/ably/ably-asset-tracking-android/pull/749/commits/3c1a4f62898321f954e6b01d76562ddfca933531
and then moved the check before the suspended block as it does not have to be inside it https://github.com/ably/ably-asset-tracking-android/pull/749/commits/9e31cd7a26e8781e36b472e87bf0010d1deebc64

ikbalkaya (approved):

PR #753, opened 2022-09-05
Send Mapbox's keyPoints as skipped locations

KacperKluka:
This is a draft of an experiment. We want to check whether treating the Mapbox's `keyPoints` (or AAT's `intermediateLocations`) as `skippedLocations` improves the UX of AAT.

From my short tests the biggest problem is that the `keyPoints` have no accuracy value which results in an odd "blinking" animation of the accuracy circle on the subscriber side.

Another issue was that when I tried to change my position before and after a crossroad (so that the straight line connecting those two locations goes over buildings) the `keyPoints` were not created. This may be because I was using an emulator and mocked locations but this won't be an issue in a real life where a rider is constantly travelling with some speed and bearing. On second thought it might also be connected to the AAT cycling profile that allows for taking shortcuts.

This video shows the "blinking" accuracy animation (don't worry about the map marker jumping, that's not a fault of `keyPoints`)

https://user-images.githubusercontent.com/62378170/188450467-8a20e6ce-75d7-4951-9ce6-984dbd46560f.mp4

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -292,9 +292,11 @@ constructor(
                 properties.lastSentEnhancedLocations[trackableId],
                 properties.resolutions[trackableId]
             ) -> {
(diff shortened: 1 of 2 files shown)


AndyTWF:
Closing for now

PR #758, opened 2022-09-15
Draft: wrapping calls in runBocking lifted up from Mapbox.kt

### LINKED ISSUE #680: Check if the usage of runBlocking in the Mapbox wrapper is safe
The `MapboxNavigation` class has to be run from the UI thread. That's why we've used `runBlocking` calls with the main dispatcher to switch the thread to the main one in the `Mapbox` wrapper. However, as pointed out [here](https://github.com/ably/ably-asset-tracking-android/pull/678#discussion_r889938867), it might not be safe. We should check if there's a better solution to this.

A possible solution suggested by @ikbalkaya :
As we need the `runBlocking` only to switch the thread we could replace it with `withContext()` calls instead. This would however require us to change the Mapbox wrapper methods to `suspend` ones.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1965) by [Unito](https://www.unito.io)

davyskiba:
`runBlocking` using the main dispatcher replaced with `withContext`. Affected methods are now `suspended`, and calls to them are wrapped in `runBlocking` without a specified dispatcher.

Extracted anonymous `LocationUpdatesObserver` implementation in `CorePublisher` to internal class for readability.

Resolves #680

### DIFF
--- a/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/MapboxTests.kt
+++ b/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/MapboxTests.kt
@@ -10,6 +10,7 @@ import com.ably.tracking.connection.ConnectionConfiguration
 import com.ably.tracking.test.android.common.NOTIFICATION_CHANNEL_ID
 import com.ably.tracking.test.android.common.createNotificationChannel
 import com.google.gson.Gson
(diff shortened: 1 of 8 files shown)


davyskiba:
Closing this, as no fix is needed

PR #762, opened 2022-09-23
Release 1.4.0-rc.1

KacperKluka:
Upgrade Mapbox to 2.8.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Change log
 
-## [1.4.0](https://github.com/ably/ably-asset-tracking-android/tree/1.4.0)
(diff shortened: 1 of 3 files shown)


QuintinWillison (changes requested):
Noticing that we are doing a `minor` version bump from `1.3` to `1.4`, then I think we need to explain to users what new functionality they are getting with this... in other words, what are the benefits of moving from MapBox `2.7` to `2.8`? This should be explained in our changelog entry.

Also, because we're moving from a GA MapBox at `2.7` to an RC MapBox at `2.8` then I think we should also release as an `RC` also. So, rather than `1.4.0` (General Availability) this should be `1.4.0-rc.1` (Release Candidate) per [our guidance](https://github.com/ably/engineering/blob/main/sdk/product-lifecycle.md#release-candidate).

QuintinWillison (approved):

PR #764, opened 2022-09-26
wrapInResultCallback utility function added to CallbackFunctionModels.kt

davyskiba:
After investigating, I've realized that moving `suspendCoroutine` to `WorkerQueue` or even to `CoreSubscriber` would introduce too much complexity.
Few workers have callbacks and would block coroutine, and we would need to distinguish them. Also, we are chaining some of the workers using the original callback on the whole chain.

In this PR I have introduced a utility function to share boilerplate code between blocking calls to enqueue

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/CallbackFunctionModels.kt
+++ b/common/src/main/java/com/ably/tracking/common/CallbackFunctionModels.kt
@@ -1,5 +1,9 @@
 package com.ably.tracking.common
 
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
 /**
  * Defines a function type, to be implemented in Kotlin code utilising the Ably Asset Tracking SDKs, allowing that code
  * to handle events using that function as a callback (typically using a lamda expression).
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/ably/tracking/common/CallbackFunctionModels.kt
```diff
@@ -11,3 +15,12 @@ typealias CallbackFunction<T> = (T) -> Unit
  * to handle an event indicating the result of an asynchronous operation.
  */
 typealias ResultCallbackFunction<T> = CallbackFunction<Result<T>>
+
+fun <T> Continuation<T>.wrapInResultCallback(): ResultCallbackFunction<T> =
```
KacperKluka, line 19: It would be great to have a documentation that explains what this helper function does 😉

KacperKluka (changes requested):
Looks great, just one additional request 😉

common/src/main/java/com/ably/tracking/common/CallbackFunctionModels.kt
davyskiba: cd096a735205975f0c2cdb7b8899217d85114944

KacperKluka (approved):
Looks good 👍 Could you please also create an issue to use this new utility function in the `CorePublisher`? 😉

PR #767, opened 2022-09-30
DefaultPublisher.kt inline ResultCallbackFunctions replaced with util…

### LINKED ISSUE #765: Replace callbacks in DefaultPublisher
Replace functional callbacks in `DefaultPublisher` with utility function `Continuation<T>.wrapInResultCallback()` defined in common/src/main/java/com/ably/tracking/common/CallbackFunctionModels.kt

davyskiba:
…ity function
Avoiding duplication by using the utility function. Resolves #765


KacperKluka (approved):
LGTM 👍

PR #773, opened 2022-10-06
Start the Ably connection when the first trackable is added and stop it when the last trackable is removed

KacperKluka:
Turned off the `autoConnect` option and now starting and stopping of the Ably instance is handled manually. When the first trackable is added (and the first channel is created) the Ably connection is started. When the last trackable is removed (i.e. the last channel is disconnected) the Ably connection is stopped. 
To make sure that those operations are synced and we don't end up in an unexpected state (i.e. having connected channels and closed Ably connection) I've used a mutex approach that checks and modifies the amount of channels in a secure and thread-safe way.

I used the opportunity to change the default `connect()` [a9e93fd](https://github.com/ably/ably-asset-tracking-android/pull/773/commits/a9e93fdc44e24b8224f3e171d3898fa3f0cf5c69) and `disconnect()` [8729e2b](https://github.com/ably/ably-asset-tracking-android/pull/773/commits/8729e2b348a1a4ce93b99241979ac4f5683b4144) methods to suspending ones and change a function to an extension function [aadd1d7](https://github.com/ably/ably-asset-tracking-android/pull/773/commits/aadd1d7b31ac8bb5aab009527998ad16b38073c1).

The main changes (starting and stopping) are included in this commit [3b104fb](https://github.com/ably/ably-asset-tracking-android/pull/773/commits/3b104fb87f9ac3f722f2d872160659b06fea34f5)

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -748,14 +748,22 @@ constructor(
 
     /**
      * Closes [AblyRealtime] and waits until it's either closed or failed.
+     * If the connection is already closed it returns immediately.
+     * If the connection is already failed it throws a [ConnectionException].
+     *
      * @throws ConnectionException if the [AblyRealtime] state changes to [ConnectionState.failed].
      */
     private suspend fun AblyRealtime.closeConnection() {
+        if (connection.state.isClosed()) {
+            return
+        } else if (connection.state.isFailed()) {
+            throw connection.reason.toTrackingException()
+        }
         suspendCancellableCoroutine<Unit> { continuation ->
             connection.on {
-                if (it.current == ConnectionState.closed) {
+                if (it.current.isClosed()) {
                     continuation.resume(Unit)
-                } else if (it.current == ConnectionState.failed) {
+                } else if (it.current.isFailed()) {
                     continuation.resumeWithException(it.reason.toTrackingException())
                 }
             }
@@ -828,6 +836,8 @@ constructor(
     private fun ChannelState.isConnected(): Boolean = this == ChannelState.attached
 
     private fun ConnectionState.isConnected(): Boolean = this == ConnectionState.connected
+
+    private fun ConnectionState.isClosed(): Boolean = this == ConnectionState.closed
 
     private fun ConnectionState.isFailed(): Boolean = this == ConnectionState.failed
 


QuintinWillison (changes requested):
Observing the fact that you're having to 'link' your `channels`, `connectingChannelNames` and `channelsMutex` fields together with commentary, I'm left wondering whether it might be better to pull all three of them out into their own class. I imagine:

- them all being `private` fields of that new class
- them having names `channels`, `connectingChannelNames` and `mutex`
- the class having an API specific to the needs of the `Ably` class - e.g.:
  - add a channel to those connecting, taking a channel name and returning whether it's the first channel or not
  - remove a channel from those connecting, taking a channel name and returning whether it was the last channel or not

This will encourage you to only lock the mutex for the minimum amount of time required.

As, for example, I see issues with code readability/maintainability and perhaps durability right now:

- the `stopAblyIfNoChannelsArePresent()` method has to be called with the mutex acquired, but that's not obvious looking at that method's definition in isolation
- you're calling `ably.connectSuspending()` when the mutex is still acquired, which I don't think is necessary

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -825,4 +885,33 @@ constructor(
...
+    private suspend fun AblyRealtime.connectSuspending(timeoutInMilliseconds: Long = 10_000L) {
+        if (connection.state.isConnected()) {
+            return
+        }
+        try {
+            withTimeout(timeoutInMilliseconds) {
+                suspendCancellableCoroutine<Unit> { continuation ->
+                    connection.on { channelStateChange ->
+                        when {
+                            channelStateChange.current.isConnected() -> continuation.resume(Unit)
```
davyskiba, line 916: Shouldn't we handle an `else` case here?

davyskiba (changes requested):
Agree with Quintin, extracting channel lists and mutex sounds great

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: Checked available statuses, and `else` does not sound like a good idea. However are we sure that those are the only two possible outcome states?

KacperKluka:
Closing as we've decided to implement this using our worker queue architecture https://github.com/ably/ably-asset-tracking-android/pull/778

PR #777, opened 2022-10-11
Move channels and their mutex to a separate class

KacperKluka:
I've tried to implement the request from [this comment](https://github.com/ably/ably-asset-tracking-android/pull/773#pullrequestreview-1135798681) from https://github.com/ably/ably-asset-tracking-android/pull/773 and wanted to create a quick draft PR to validate whether I'm following the correct approach.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -38,8 +38,6 @@ import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.supervisorScope
 import kotlinx.coroutines.suspendCancellableCoroutine
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeout
 
 /**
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/ably/tracking/common/AblyChannelsGuard.kt
```diff
@@ -0,0 +1,66 @@
...
+     */
+    private val channelsMutex = Mutex()
+
+    fun getChannel(trackableId: String): Channel? = channels[trackableId]
+
+    fun getChannelTrackableIds(): Set<String> = channels.keys
+
+    fun isChannelCreated(trackableId: String): Boolean = channels.containsKey(trackableId)
+
+    suspend fun startConnectingChannel(trackableId: String, onConnectingFirstChannel: suspend () -> Unit) {
```
davyskiba, line 29: This name confused me a bit, I was expecting to see a channel connection inside this. Maybe we could rename it to something like `addConnectingChannel`?

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -413,20 +384,22 @@ constructor(
     }
 
     override suspend fun disconnect(trackableId: String, presenceData: PresenceData): Result<Unit> {
-        return if (channels.contains(trackableId)) {
-            val channelToRemove = channels[trackableId]!!
+        return if (channelsGuard.isChannelCreated(trackableId)) {
+            val channelToRemove = channelsGuard.getChannel(trackableId)!!
```
davyskiba, line 388: This force unwrap could be avoided with something like
`  val channelToRemove = channelsGuard.getChannel(trackableId) ?: return Result.success(Unit)`

davyskiba:
I like this approach :)

KacperKluka:
Closing as we won't use the mutex based approach

PR #778, opened 2022-10-13
Start and stop the Ably connection manually depending on the number of added trackables

KacperKluka:
Another approach to the start/stop Ably connection issue, this time using the worker queue architecture to achieve this. Requirements:

- The connection should be stopped when the publisher is created (hence the `autoConnect = false` is used in the `Ably` wrapper)
- When the first trackable is added the connection should be started
- When the last trackable is removed the connection should be stopped

I've made sure that no trackables are being added when the Ably connection is starting or stopping by delaying the add worker if a connection change is happening at the moment.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -210,6 +210,16 @@ interface Ably {
      * @throws ConnectionException if something goes wrong.
      */
     suspend fun close(presenceData: PresenceData)
+
+    /**
+     * Starts the Ably connection. Must be called before any other method.
+     */
+    suspend fun startConnection(): Result<Unit>
+
+    /**
+     * Stops the Ably connection.
+     */
+    suspend fun stopConnection(): Result<Unit>
 }
 
 private const val CHANNEL_NAME_PREFIX = "tracking:"
@@ -707,23 +717,31 @@ constructor(
                 launch { disconnect(trackableId, presenceData) }
             }
         }
-        closeConnection()
+        stopConnection()
     }
 
     /**
      * Closes [AblyRealtime] and waits until it's either closed or failed.
+     * If the connection is already closed it returns immediately.
+     * If the connection is already failed it throws a [ConnectionException].
+     *
      * @throws ConnectionException if the [AblyRealtime] state changes to [ConnectionState.failed].
      */
-    private suspend fun closeConnection() {
+    private suspend fun AblyRealtime.closeSuspending() {
+        if (connection.state.isClosed()) {
+            return
+        } else if (connection.state.isFailed()) {
+            throw connection.reason.toTrackingException()
+        }
         suspendCancellableCoroutine<Unit> { continuation ->
-            ably.connection.on {
-                if (it.current == ConnectionState.closed) {
+            connection.on {
+                if (it.current.isClosed()) {
                     continuation.resume(Unit)
-                } else if (it.current == ConnectionState.failed) {
+                } else if (it.current.isFailed()) {
                     continuation.resumeWithException(it.reason.toTrackingException())
                 }
             }
-            ably.close()
+            close()
         }
     }
 
@@ -791,6 +809,12 @@ constructor(
 
     private fun ChannelState.isConnected(): Boolean = this == ChannelState.attached
 
+    private fun ConnectionState.isConnected(): Boolean = this == ConnectionState.connected
+
+    private fun ConnectionState.isClosed(): Boolean = this == ConnectionState.closed
+
+    private fun ConnectionState.isFailed(): Boolean = this == ConnectionState.failed
+
     private fun ConnectionException.isConnectionResumeException(): Boolean =
         errorInformation.let { it.message == "Connection resume failed" && it.code == 50000 && it.statusCode == 500 }
 
@@ -845,4 +869,54 @@ constructor(
         state == ChannelState.detached || state == ChannelState.failed
 
     private fun createMalformedMessageErrorInfo(): ErrorInfo = ErrorInfo("Received a malformed message", 100_001, 400)
+
+    override suspend fun startConnection(): Result<Unit> {
+        return try {
+            ably.connectSuspending()
+            Result.success(Unit)
+        } catch (connectionException: ConnectionException) {
+            Result.failure(connectionException)
+        }
+    }
+
+    override suspend fun stopConnection(): Result<Unit> {
+        return try {
+            ably.closeSuspending()
+            Result.success(Unit)
+        } catch (connectionException: ConnectionException) {
+            Result.failure(connectionException)
+        }
+    }
+
+    /**
+     * A suspending version of the [AblyRealtime.connect] method. It will begin connecting and wait until it's connected.
+     * If the connection enters the "failed" state it will throw a [ConnectionException].
+     * If the operation doesn't complete in [timeoutInMilliseconds] it will throw a [ConnectionException].
+     * If the instance is already connected it will finish immediately.
+     * If the connection is already failed it throws a [ConnectionException].
+     *
+     * @throws ConnectionException if something goes wrong.
+     */
+    private suspend fun AblyRealtime.connectSuspending(timeoutInMilliseconds: Long = 10_000L) {
+        if (connection.state.isConnected()) {
+            return
+        } else if (connection.state.isFailed()) {
+            throw connection.reason.toTrackingException()
+        }
+        try {
+            withTimeout(timeoutInMilliseconds) {
+                suspendCancellableCoroutine<Unit> { continuation ->
+                    connection.on { channelStateChange ->
+                        when {
+                            channelStateChange.current.isConnected() -> continuation.resume(Unit)
+                            channelStateChange.current.isFailed() -> continuation.resumeWithException(channelStateChange.reason.toTrackingException())
+                        }
+                    }
+                    connect()
+                }
+            }
+        } catch (exception: TimeoutCancellationException) {
+            throw ConnectionException(ErrorInformation("Timeout was thrown when waiting for Ably to connect"))
+        }
+    }
 }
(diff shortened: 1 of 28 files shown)


KacperKluka:
Converting to draft as I've forgotten to make changes in the Subscriber SDK

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
```diff
@@ -31,11 +37,30 @@ internal class AddTrackableWorker(
...
+                        asyncWork = {
+                            // delay work until Ably connection manipulation ends
+                            delay(WORK_DELAY_IN_MILLISECONDS)
+                            AddTrackableWorkResult.WorkDelayed(trackable, callbackFunction, presenceUpdateListener, channelStateChangeListener)
+                        }
+                    )
+                }
+                val isAddingTheFirstTrackable = properties.hasNoTrackablesAddingOrAdded
+                if (isAddingTheFirstTrackable) {
+                    properties.isConnectingToAbly = true
```
QuintinWillison, line 51: I find this property mutation curious and I'm not sure why it would be here. Can you explain? Or perhaps add some commentary to the code? It's just, given what's above it, it just smells wrong if you know what I mean. 🤔

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
```diff
@@ -9,11 +9,17 @@ import com.ably.tracking.publisher.PublisherProperties
...
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.StateFlow
 
 internal typealias AddTrackableResult = StateFlow<TrackableState>
 internal typealias AddTrackableCallbackFunction = ResultCallbackFunction<AddTrackableResult>
 
+/**
+ * How long should we wait before re-queueing the work if Ably connection manipulation is in progress.
+ */
+private const val WORK_DELAY_IN_MILLISECONDS = 200L
```
QuintinWillison, line 21: This value is clearly a little arbitrary. Could it be worth expanding the commentary here to explain:

- why this value was chosen
- what "connection manipulation" means
- what happens if we wait, the work is re-queued, and then we find that connection manipulation is still ongoing... I _think_ we'll keep retrying, but it's not 100% clear to me from just reading the code that this is the case (i.e. we'll keep retrying at a rate of approximately 5Hz)

QuintinWillison (changes requested):
I have been having a bit of a think about the growing collection of `Boolean` fields in `PublisherProperties`. You already had `isStopped` but you're now adding `isConnectingToAbly` and `isStoppingAbly`. I think they're clearly pretty interrelated, so I wonder whether it might be better to perhaps introduce an enum so that you can only have one field which can ever be in one of a few expected states? The code is a little hard to reason about at the moment - for example, I worry about a state where `isConnectingToAbly` and `isStoppingAbly` both end up `true` at the same time. The compiler cannot stop this happening. The only protection is that the mutations of the properties instance are consistent across all parts of the codebase that have access to it, which is a maintainability risk. 🤔

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
KacperKluka: When we have no trackables the connection to Ably is closed, therefore, when we add the first trackable we have to start the connection. Previously this property was set to `true` because we wanted to let other workers know that we're currently connecting to Ably so they should wait until that's complete before attempting any Ably-related work (like adding another trackable). Now, instead of this boolean property we have a state value which hopefully will be more meaningful https://github.com/ably/ably-asset-tracking-android/pull/778/commits/878781a0d402e1e4fd0cc1ae55553fb88284b181 😉

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
KacperKluka: > why this value was chosen

I have to admit I just randomly chosen that value 😅 Now I wonder if repeating the operation 5 times per second isn't too often 🤔 It all depends on the actual time required to start the connection which depends on things like network connection. WDYT? Maybe it would be better to increase the delay to 500ms or even 1s? 🤔 

> what "connection manipulation" means

It means starting or stopping the connection and now I see that I've made this vague without a good reason 🤦‍♂️ I'll ammend the comment and make it explicit 👍 https://github.com/ably/ably-asset-tracking-android/pull/778/commits/1f2d78bd2edb85c6d24f2c7198be057ff197624f

> what happens if we wait, the work is re-queued, and then we find that connection manipulation is still ongoing... I think we'll keep retrying, but it's not 100% clear to me from just reading the code that this is the case (i.e. we'll keep retrying at a rate of approximately 5Hz)

You're right, I've added this info to the docs 🙇 https://github.com/ably/ably-asset-tracking-android/pull/778/commits/1f2d78bd2edb85c6d24f2c7198be057ff197624f

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableFailedWorker.kt
```diff
@@ -1,20 +1,37 @@
...
     private val trackable: Trackable,
     private val callbackFunction: AddTrackableCallbackFunction,
-    val exception: Exception
+    val exception: Exception,
+    private val isConnectedToAbly: Boolean,
+    private val ably: Ably,
 ) : Worker {
     override fun doWork(properties: PublisherProperties): SyncAsyncResult {
+        if (properties.state == PublisherState.CONNECTING) {
+            properties.state = if (isConnectedToAbly) PublisherState.CONNECTED else PublisherState.IDLE
```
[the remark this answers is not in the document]
KacperKluka: We have to do it here because changing the `properties` have to be made from the main queue thread 😉

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
```diff
@@ -31,11 +39,30 @@ internal class AddTrackableWorker(
                 SyncAsyncResult(AddTrackableWorkResult.AlreadyIn(properties.trackableStateFlows[trackable.id]!!, callbackFunction))
             }
             else -> {
+                if (properties.state == PublisherState.CONNECTING || properties.state == PublisherState.DISCONNECTING) {
```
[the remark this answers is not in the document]
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/778/commits/34574992c9cc5230c9a5670c7e55ef91d26f7086

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
```diff
@@ -31,11 +39,30 @@ internal class AddTrackableWorker(
...
+                val isAddingTheFirstTrackable = properties.hasNoTrackablesAddingOrAdded
+                if (isAddingTheFirstTrackable) {
+                    properties.state = PublisherState.CONNECTING
+                }
                 properties.duplicateTrackableGuard.startAddingTrackable(trackable)
                 val presenceData = properties.presenceData.copy()
 
                 SyncAsyncResult(
                     asyncWork = {
+                        if (isAddingTheFirstTrackable) {
```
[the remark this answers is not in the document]
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/778/commits/aeb7d4bed50a2543806e271773d059a4e69614ea

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -707,23 +718,31 @@ constructor(
...
             }
         }
-        closeConnection()
+        stopConnection()
     }
 
     /**
      * Closes [AblyRealtime] and waits until it's either closed or failed.
+     * If the connection is already closed it returns immediately.
+     * If the connection is already failed it throws a [ConnectionException].
```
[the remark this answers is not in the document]
KacperKluka: Good point 👍  https://github.com/ably/ably-asset-tracking-android/pull/778/commits/b1a9c108208632d212e54d5fa1f1c6a5322d6a56

publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
```diff
@@ -752,6 +751,15 @@ constructor(
         override val trackableRemovalGuard: TrackableRemovalGuard = DefaultTrackableRemovalGuard()
             get() = if (isDisposed) throw PublisherPropertiesDisposedException() else field
         override val areRawLocationsEnabled: Boolean = areRawLocationsEnabled ?: false
+        override var state: PublisherState = PublisherState.IDLE
+            set(value) {
+                // Once we stop publisher it should never change its state
+                if (field != PublisherState.STOPPED) {
```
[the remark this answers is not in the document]
KacperKluka: That's a good idea 👍 https://github.com/ably/ably-asset-tracking-android/pull/778/commits/df1289d36eb8d92efe9871ff339fff2d0d28784d

davyskiba (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
QuintinWillison: @KacperKluka Perhaps you can add an issue to look at this again later, with the intention of improving the algorithm (if possible). Feel free to resolve this conversation on my behalf once you've created that issue. Thanks.

QuintinWillison (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
KacperKluka: Here's the issue 😉 https://github.com/ably/ably-asset-tracking-android/issues/783

PR #780, opened 2022-10-17
Add information about SLF4J warning logs

KacperKluka:
Added info about a known warning logs from SLF4J that are a normal thing if there's no concrete logger implementation provided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,6 +380,17 @@ val mapboxNavigation = MapboxNavigationProvider.retrieve()
 Because there is only one `MapboxNavigation` instance, both your app and AAT will use the same object. This means that there can be possible conflicts in usage that can lead to unexpected behaviour.
 Therefore, we do not advise using AAT in applications that already use Mapbox Navigation SDK.
 
+### SLF4J warning logs
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -380,6 +380,17 @@ val mapboxNavigation = MapboxNavigationProvider.retrieve()
 Because there is only one `MapboxNavigation` instance, both your app and AAT will use the same object. This means that there can be possible conflicts in usage that can lead to unexpected behaviour.
 Therefore, we do not advise using AAT in applications that already use Mapbox Navigation SDK.
 
+### SLF4J warning logs
+
+AAT has a transitive dependency on the `SLF4J` library and it does not provide any default logger implementation. Because of that you can encounter below warning logs:
```
QuintinWillison, line 385:
```suggestion
AAT has a transitive dependency on the [SLF4J](https://www.slf4j.org/) library but we do not provide a default logger implementation. Because of that you can encounter below warning logs if you have not explicitly provided an implementation:
```

QuintinWillison (approved):
Looks good to me, just one minor tweak for clarity...

PR #782, opened 2022-10-18
Feature/error proof wrapper

davyskiba:
This PR takes care of further error proofing Ably wrapper:

1. dropped our own channels list to avoid  potential multithreading issues in favor of using a channels list held by Ably-java 2d1bd5f48a35c011232d14c5f2e187f512dcaf86
2. reordered calls inside `failChannel` so that the listeners have a chance to get notified and deleted removing the failed channel because we can reuse it and reconnect to it 83ffe9612a4788d94f7a6208e9f3288424fa4fc9
3. ~~altered the `disconnect` method so it can no longer fail - kept `leavePresence` and moved all other clean-up calls outside to the `finally` block as they cannot fail. If I understand ably-java doc correctly, we should be using `ably.channels.release` instead of detaching the channel manually, and currently, this method eats all exceptions.~~ 91f122249fab0db874f32c595766e05837e0cbfe


common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -339,7 +338,6 @@ constructor(
                             channel.attachSuspending()
                         }
                         enterChannelPresence(channel, presenceData)
-                        channels[trackableId] = channel
```
KacperKluka, line 342: Previously we only considered a channel to be active and added it to our channels list when it succeeded in `enterChannelPresence()` operation. Now it looks like after the `channels.get()` is called (from a few lines above) the channel will be considered active and available for other operations (such as subscribing to it). This will happen also when the enter presence operation fails as in the `catch` blocks we only call the callback and we do not release the channel that was created by the `channels.get()` call, am I right? 🤔

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -374,35 +372,28 @@ constructor(
...
     override fun disconnect(trackableId: String, presenceData: PresenceData, callback: (Result<Unit>) -> Unit) {
         scope.launch {
-            if (channels.contains(trackableId)) {
-                val channelToRemove = channels[trackableId]!!
+            val channelToRemove = getChannelIfExists(trackableId)
+            if (channelToRemove != null) {
                 try {
                     retryChannelOperationIfConnectionResumeFails(channelToRemove) {
-                        disconnectChannel(it, presenceData)
+                        leavePresence(channelToRemove, presenceData)
```
KacperKluka, line 379: What will happen if this fails and throws an exception? Why do you consider a channel that has not left its presence as successfully closed? 🤔 Will Ably leave the presence of a channel automatically when the channel is released? 🤔

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: Right! I've missed this, should I add `channels.remove` to the catch block? 🤔

KacperKluka: I think we should remove the channel when connecting to it fails, so the `remove` or even `release` would be a right choice here 🤔

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: If it fails, we log the exception and retry once. I'll probably add a catch here too 🤔 

I've assumed that Ably has to have some sort of auto-leave mechanism for cases when the connection is cut off, but haven't confirmed it yet - will try to confirm it.
If I understand correctly, leaving presence is actually notifying the channel that we are leaving by sending a message. I thought that if sending a message fails twice, we're already in some partially disconnected/failed state, and I'm not sure if we should keep the channel open 🤔

davyskiba: Just checked on the demo apps using the dev console - publisher presence was removed in roughly 40 seconds after entering airplane mode on the device

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: 739dfefcbb1ab0c4c2f8dbdec49b004654771a38

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: After discussion and deeper testing reverted this change in f8da7f1b6cffcb4a7497b460dada70fd2f4b8b2e

KacperKluka: I think you've meant https://github.com/ably/ably-asset-tracking-android/pull/782/commits/a70b6f35be5f4a0119e3fb05975bb17866b9d33c and I'm resolving this conversation 😉

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -406,14 +404,15 @@ constructor(
         leavePresence(channel, presenceData)
         channel.unsubscribe()
         channel.presence.unsubscribe()
-        detachFromChannel(channel)
+        ably.channels.release(channel.name)
     }
 
     private suspend fun failChannel(channel: Channel, presenceData: PresenceData, errorInfo: ErrorInfo) {
+        // Once a channel enters failed state we can reattach it again, so we're not removing the reference
```
KacperKluka, line 411: Will this reattach happen? If yes then when? I'm asking because the idea was that when we receive a malformed location update message then the transport (channel) should be failed and stop operating 🤔

KacperKluka (commented):
Looks good 👍 Just one question 😉

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: During connect, we attempt to reattach a failed channel

KacperKluka: But the `connect()` method starts with this
```kotlin
if (!ably.channels.containsKey(trackableId)) {
  // start connection
}
```

so if I'm right since the failed channel is not removed from the channels, then this won't be executed, right? Therefore, the channel will never reattach? 😅

davyskiba: Right 😅 I think we could add checking channel state to this condition? 🤔

KacperKluka: We could but this makes me wonder, is that worth the effort? 😅 I mean why don't we simply remove (release) the channel when it fails? 🤔

davyskiba: 6272de5e94df24841f36bf1d1ab621778db59a59

KacperKluka (approved):
LGTM 👍

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -343,32 +342,20 @@ constructor(
...
-                val modesList = mutableListOf(ChannelMode.presence, ChannelMode.presence_subscribe)
-                if (willPublish) {
-                    modesList.add(ChannelMode.publish)
-                }
-                if (willSubscribe) {
-                    modesList.add(ChannelMode.subscribe)
-                }
-                modes = modesList.toTypedArray()
-            }
+        if (!ably.channels.containsKey(trackableId)) {
```
KacperKluka, line 345: I think you should either use `trackableId.toChannelName()` or switch to the `getChannelIfExists()` method 🤔

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -742,10 +724,19 @@ constructor(
...
+        } else {
+            null
+        }
+    }
+
     override suspend fun close(presenceData: PresenceData) {
         // launches closing of all channels in parallel but waits for all channels to be closed
         supervisorScope {
-            channels.keys.forEach { trackableId ->
+            ably.channels.keySet().forEach { trackableId ->
```
KacperKluka, line 739: I believe that now the `keySet()` will return channel names instead of trackable IDs 🤔 So the variable name should be changed to `channelName` and we should check if it still can be used later in the `disconnect()` method below 🤔

integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
```diff
@@ -232,13 +233,16 @@ class PublisherAndSubscriberTests {
...
+        // then
+        subscriberFailedExpectation.assertFulfilled()
+        assertThat(subscriber.trackableStates.value)
+            .isInstanceOf(TrackableState.Failed::class.java)
+
         // cleanup
+        // moved to the end of the test because currently the subscriber on stop is transitioning the trackable to Offline state, reported here https://github.com/ably/ably-asset-tracking-android/issues/802
         runBlocking {
             subscriber.stop()
         }
```
KacperKluka, line 245: I wonder whether the "cleanup" code will be called if one of the assertions from the "then" section fail? I guess we should always stop the subscriber, even if the assertions fail. So maybe let's try a `try/catch/finally` approach to achieve this here? 🤔

KacperKluka (changes requested):
Almost there 😅

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -422,14 +424,16 @@ constructor(
         leavePresence(channel, presenceData)
         channel.unsubscribe()
         channel.presence.unsubscribe()
-        detachFromChannel(channel)
+        ably.channels.release(channel.name)
```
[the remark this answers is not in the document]
davyskiba: We could do that as a part of the cleanup, but aren't we relying on them to move the trackable to an offline state?

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: 99a6d64ca78b6e10db3c0ca41f1db7ec271b6d5d

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: awesome you caught this <3 294406b4185e63ec760d692c2b005d2f9f14ba41

integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
davyskiba: I think this would be getting too complicated 🤔 
reordered it this way 3f883ca8516033bb9164a7e9c362d598c44f26aa

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: aand  here a8cdb3ed1697e688c53019d74a4a1cb76800efc6

common/src/main/java/com/ably/tracking/common/Ably.kt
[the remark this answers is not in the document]
KacperKluka: Okay, so why are we calling the `channel.off()` in the [other case](https://github.com/ably/ably-asset-tracking-android/pull/782/files#diff-baf8644244d013d061c0a0d0bc74b8f8156783cc77f7796198b11172cccb9d38R442) when we are failing the channel? 🤔

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -418,18 +418,29 @@ constructor(
         }
     }
 
-    private suspend fun disconnectChannel(channel: Channel, presenceData: PresenceData) {
-        leavePresence(channel, presenceData)
-        channel.unsubscribe()
-        channel.presence.unsubscribe()
-        detachFromChannel(channel)
+    private suspend fun safeDisconnectChannel(channelToRemove: Channel, presenceData: PresenceData) =
```
KacperKluka, line 421: Why is this called `safe`? What's unsafe with the regular `disconnectChannel`? 🤔

KacperKluka (changes requested):

common/src/main/java/com/ably/tracking/common/Ably.kt
davyskiba: This one will never throw an exception

common/src/main/java/com/ably/tracking/common/Ably.kt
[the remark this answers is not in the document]
KacperKluka: Right, sorry, I've got everything mixed up 🙈 If we add `channel.off()` here we won't receive the "channel went offline" event because it's triggered by the `ably.channels.release()` call 👍 So everything is fine here 👌

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: Do you think we could change the name that would explain this method's behaviour in more detail? 🤔 Something like `disconnectChannelIgnoringErrors()` or perhaps `tryDisconnectChannel()` or something else? 🤔

davyskiba: renamed to `tryDisconnectChannel` https://github.com/ably/ably-asset-tracking-android/pull/782/commits/cade10d817ea3988bc8863c716bdfeacf4116e9d

Issue #781, opened 2022-10-17
Refactor Publisher EventQueue to match Subscriber


davyskiba:
When introducing the event queue to subscribing SDK, we simplified it a bit by dropping the `ResultHandler` approach.
Let's refactor EventQueue inside the publishing SDK to match this approach.

PR #786, opened 2022-10-24
WorkerQueue.kt moved to commons package

davyskiba:
A preparatory work to move WorkerQueue to the commons module for reuse in the publisher in #781


common/src/main/java/com/ably/tracking/common/workerqueue/Properties.kt
```diff
@@ -0,0 +1,8 @@
+package com.ably.tracking.common.workerqueue
+
+/**
+ * An abstract base class, containing properties that affect [WorkerQueue] behavior.
+ */
+abstract class Properties {
```
KacperKluka, line 6: Why not an interface? 🤔

common/src/main/java/com/ably/tracking/common/workerqueue/Worker.kt
```diff
@@ -1,12 +1,14 @@
...
-import com.ably.tracking.subscriber.Properties
 
 /**
  * A [Worker] interface represents workers which execute work inside [WorkerQueue].
+ * Params:
+ * P - the type of properties used by this worker as both input and output
+ * S - the type of specification used to post another worker back to the queue
  */
-internal interface Worker {
+interface Worker<P : Properties, S> {
```
KacperKluka, line 11: WDYT about using full words for the type parameters? I read an article that suggested using full names because one-letter abbreviations introduce unnecessary cognitive load while you're analyzing code 😉 Instead of using `P` and `S` and having this docs annotation about what they represent, we could use `PropertiesType` and `WorkerSpecificationType` which should make it self explanatory both here and below in places where you use it 🤔 
```suggestion
interface Worker<PropertiesType : Properties, WorkerSpecificationType> {
```

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerFactory.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+interface WorkerFactory<P : Properties, S> {
+
+    /**
+     * Creates an appropriate [Worker] from the passed specification [S].
+     *
+     * @param params The parameters that indicate which [Worker] implementation should be created.
+     * @return New [Worker] instance.
+     */
+    fun createWorker(params: S): Worker<P, S>
```
KacperKluka, line 17: Shouldn't we call the parameter `specification` or `workerSpecification` instead of `params`? 🤔

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerQueue.kt
```diff
@@ -0,0 +1,55 @@
...
+ * A worker queue is responsible for enqueueing [Worker]s and executing them.
+ * Params:
+ * P - the type of properties used by this worker as both input and output
+ * S - the type of specification used to post worker back to the queue
+ */
+class WorkerQueue<P : Properties, S>(
+    private var properties: P,
+    private val scope: CoroutineScope,
+    private val workerFactory: WorkerFactory<P, S>,
+    private val copyProperties: P.() -> P,
```
KacperKluka, line 17: That's a bit odd 🤔 Can't we simply add a `copy` method to the `Properties` interface instead of requiring an extension function here? 🤔

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerQueue.kt
```diff
@@ -0,0 +1,55 @@
...
+    private val getStoppedException: () -> Exception,
```
KacperKluka, line 18: Can't it just be an `Exception` instead of a function that returns an exception? 🤔

KacperKluka (changes requested):
A few suggestions 😉

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerQueue.kt
davyskiba: If we pass an exception here, it will be the exact same instance used for every worker.
I'm wondering if this can lead to issues when handling them?

KacperKluka: Hmmm true, let's stay on the safe side then 😅

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerQueue.kt
KacperKluka: In this case, that's definitely better 😅 Could you perhaps add a comment here about why we're using this approach?

common/src/main/java/com/ably/tracking/common/workerqueue/Properties.kt
davyskiba: 7932f445f57e63e98f60f7a9f1a724ea2a4fabb4

common/src/main/java/com/ably/tracking/common/workerqueue/Worker.kt
davyskiba: This doesn't feel natural, but I see no reason not to do it 😮 
01a23c27660695394f7e30f8660fec637d86992f

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerQueue.kt
davyskiba: 4713c57ffc5c2803640a5dca031068b13892a862

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerFactory.kt
davyskiba: 4713c57ffc5c2803640a5dca031068b13892a862

common/src/main/java/com/ably/tracking/common/workerqueue/Worker.kt
KacperKluka: Actually, I didn't mean to rename the `Properties` to something different in order to change this type declaration 😅 I wanted to change the `P` to `PropertiesType` to make it explicit that this is a "generic type" and it extends "Properties" 😉 So basically to take the original name and add the `Type` suffix.

davyskiba: ed24ab40047589268666b59f404bbb32077c81f2

KacperKluka: I think you've missed [one place](https://github.com/ably/ably-asset-tracking-android/pull/786/files#diff-295145610258ad576bb05bd11957d365028c45d65d43407467d6e9dfa88503e4R42-R43) 😉

davyskiba: 61f05aa451463edbd94461b45ce710de03beb0ed

Issue #173, opened 2021-02-09
Annotate Java facade listener interfaces as functional


QuintinWillison:
See https://www.baeldung.com/java-8-lambda-expressions-tips#functionalInterface for reasons behind this being a good idea / best practice.

KacperKluka:
@QuintinWillison I've checked and the `@FunctionalInterface` annotation only works in `.java` files. If it's used in a `.kt` file it has no effect at all and you can declare multiple methods in the annotated interface without any errors. 
So if we'd really like to use that annotation then we would have to refactor those interfaces from Kotlin to Java. WDYT?

QuintinWillison:
Thanks for checking. That is interesting. Refactoring these interfaces to Java is probably a very good idea. I can't imagine there are any Kotlin features that the facade projects (for [publishing](https://github.com/ably/ably-asset-tracking-android/tree/main/publishing-sdk-java) and [subscribing](https://github.com/ably/ably-asset-tracking-android/tree/main/subscribing-sdk-java)) are using that would stop us doing this, but that would clearly fall out when it comes to carrying out this task.

QuintinWillison:
Relates to https://github.com/ably/ably-asset-tracking-android/issues/174, in that we could present the Java facade interfaces as a Java project and then implement them in Kotlin via a second project.

i.e., ideally, we would keep the existing _Kotlin_ projects `subscribing-sdk-java` and `publishing-sdk-java` as facade implementations, and add new _Java_ projects `subscribing-sdk-java-api` and `publishing-sdk-java-api` which would be pure interface declarations, leveraging `@FunctionalInterface` annotation where appropriate.

PR #787, opened 2022-10-24
Add token based auth to the Java API

KacperKluka:
I've created an `AuthenticationFacade` static class that is responsible for creating all authentication options in the Java API. I've moved the `basic` auth to it and added `tokenRequest` and `jwt` auths that instead of coroutines use the `CompletableFuture`s.


davyskiba (approved):

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -39,7 +44,11 @@ public void beforeEach() {
     @Test
     public void subscriberBuilderUsageExample() {
         Subscriber.Builder nativeBuilder = Subscriber.subscribers()
-            .connection(new ConnectionConfiguration(Authentication.basic("CLIENT_ID", "API_KEY"), null))
+            .connection(new ConnectionConfiguration(AuthenticationFacade.basic("CLIENT_ID", "API_KEY"), null))
```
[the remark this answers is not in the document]
KacperKluka: Those tests are just a way to show how the Java API can be used and not necessarily to test anything 😅 Therefore, I'd just leave it here as this should be easier for the users to see all options in one place 😉 WDYT?

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -78,4 +87,58 @@ public void subscriberFacadeUsageExample() {
             // handle interruption exception
         }
     }
+
+    private Authentication createTokenRequestAuthentication() {
+        return AuthenticationFacade.tokenRequest(tokenParams -> CompletableFuture.supplyAsync(() -> {
+            // get the token from you auth servers
+            return new TokenRequest() {
```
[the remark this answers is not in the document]
KacperKluka: Good idea 👍 https://github.com/ably/ably-asset-tracking-android/pull/787/commits/ad9329d723cfd9d09cd3d07c5e4bf6362504e354

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -78,4 +83,18 @@ public void subscriberFacadeUsageExample() {
...
+    private Authentication createTokenRequestAuthentication() {
+        return AuthenticationFacade.tokenRequest(tokenParams -> CompletableFuture.supplyAsync(() -> {
+            // get the token from you auth servers
+            return new EmptyTokenRequest();
+        }));
+    }
+
+    private Authentication createJwtAuthentication() {
+        return AuthenticationFacade.jwt(tokenParams -> CompletableFuture.supplyAsync(() -> {
+            // get the token from you auth servers
```
QuintinWillison, line 96:
```suggestion
            // get the token from your auth servers
```

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -78,4 +83,18 @@ public void subscriberFacadeUsageExample() {
...
+            return "created-jwt";
```
QuintinWillison, line 97: This could perhaps be clearer to indicate that it's a placeholder - even the dashes, at the moment, make it look more 'official'. Perhaps:

```suggestion
            return "DUMMY PLACEHOLDER: RETURN JWT FROM AUTH SERVER HERE";
```

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -78,4 +83,18 @@ public void subscriberFacadeUsageExample() {
             // handle interruption exception
         }
     }
+
+    private Authentication createTokenRequestAuthentication() {
+        return AuthenticationFacade.tokenRequest(tokenParams -> CompletableFuture.supplyAsync(() -> {
+            // get the token from you auth servers
```
QuintinWillison, line 89:
```suggestion
            // get the token from your auth servers
```

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/EmptyTokenRequest.java
```diff
@@ -0,0 +1,48 @@
+package com.ably.tracking.example.javasubscriber;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.ably.tracking.connection.TokenRequest;
+
+public class EmptyTokenRequest implements TokenRequest {
```
QuintinWillison, line 8: Perhaps some commentary on this class to explain that the values returned from the implemented getters are intentionally placeholder dummies, in order to demo how to implement this interface.

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
```diff
@@ -51,7 +51,7 @@ sealed class Authentication(
          * @param apiKey Ably key string as obtained from the dashboard.
          * @param clientId ID of the client
          */
-        @JvmStatic
+        @JvmSynthetic
```
QuintinWillison, line 54: Does this mean we're presenting a breaking change for those using this SDK's core interfaces from Java?

Or might we classify this as a bug fix? (i.e. based on the argument that this method should never have been visible to Java users) 🤔

core-sdk-java/src/main/java/com/ably/tracking/java/AuthenticationFacade.kt
```diff
@@ -0,0 +1,62 @@
+package com.ably.tracking.java
+
+import com.ably.tracking.TokenAuthException
+import com.ably.tracking.connection.Authentication
+import com.ably.tracking.connection.TokenParams
+import com.ably.tracking.connection.TokenRequest
+import kotlinx.coroutines.future.await
+import java.util.concurrent.CompletableFuture
+
+interface TokenRequestFunction {
```
QuintinWillison, line 10: I'm not sure I like the `Function` suffix on this interface and `JwtFunction` below.

I wonder if `Creator` or `CreationHandler` would be more classically Java idiomatic.

(while I recognise that this is a functional interface, we're not [annotating it with `FunctionalInterface`](https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html) - a [known issue](https://github.com/ably/ably-asset-tracking-android/issues/173) in this codebase... so mixing in 'function' into interface names feels a bit strange)

QuintinWillison (changes requested):

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/787/commits/ccf94c19bd2fbb41201098788c608b5ba2411b57

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/EmptyTokenRequest.java
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/787/commits/ccf94c19bd2fbb41201098788c608b5ba2411b57

core-sdk-java/src/main/java/com/ably/tracking/java/AuthenticationFacade.kt
KacperKluka: I didn't like the `Function` as well but no better name came to my mind 😅 I like the `Creator` one, it's much better 🙇 https://github.com/ably/ably-asset-tracking-android/pull/787/commits/aea31d514b5aab64e25169f7e16515652083f0a4

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
KacperKluka: Yes, that's a breaking change 🤔 Technically we don't need to do this because the `basic` auth can safely work without a facade, but I thought that it would be better if all auth options in Java are created from the same `AuthenticationFacade` object 🤔 So it's not a "bug", this is more of a consistency thing. WDYT we should do with it then? 🤔

QuintinWillison: @mikelee638 - while this is technically a breaking change (therefore should be held off until the next `major` version bump), at the same time we're presenting in this pull request an enhancement that expands the scope of auth support... I'm also not aware of any customers who've been in contact that are using this from Java. So, perhaps we _can_ release this in a `minor` version bump. What do you think?

mikelee638: @QuintinWillison Yes, seems fine to me.

QuintinWillison: Thanks. I've added the `enhancement` label to both this PR and the linked issue.

PR #791, opened 2022-10-25
Add the location animator to the Java API

KacperKluka:
I've created a facade for Java users that should allow them to use the `LocationAnimator` with all its features. In the process I've added a new module `ui-sdk-java` which will be published as an artefact.


subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -103,6 +108,30 @@ public void subscriberFacadeUsageExample() {
...
+    public void locationAnimatorFacadeUsageExample() {
+        LocationAnimator locationAnimator = new CoreLocationAnimator();
+        LocationAnimatorFacade locationAnimatorFacade = new CoreLocationAnimatorFacade(locationAnimator);
+
+        AtomicLong locationUpdateInterval = new AtomicLong(0L);
+        subscriberFacade.addNextLocationUpdateIntervalListener(locationUpdateIntervalInMilliseconds -> {
+            locationUpdateInterval.set(locationUpdateIntervalInMilliseconds);
+        });
+        subscriberFacade.addLocationListener(locationUpdate -> {
+            locationAnimatorFacade.animateLocationUpdate(locationUpdate, locationUpdateInterval.get());
```
davyskiba, line 121: What should we do with that in a real-world case?

davyskiba (approved):

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
KacperKluka: What do you mean? This more or less shows how to use it in an app 😉

ui-sdk-java/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
QuintinWillison, line 1: I think you can remove this file entirely, as this folder is already covered [here](https://github.com/ably/ably-asset-tracking-android/blob/8016b0fb97be04332d0c4d39404933a944d444fc/.gitignore#L6) in the root Git ignore file.

ui-sdk-java/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+    id 'kotlin-android'
+    id 'org.jlleitschuh.gradle.ktlint'
+    id 'org.jetbrains.dokka'
+}
+
+ext {
+    PUBLISH_POM_DESCRIPTION = "Ably Asset Tracking UI extensions module for Java."
+}
+
+dependencies {
```
QuintinWillison, line 12: I noticed [this](https://github.com/ably/ably-asset-tracking-android/blob/8016b0fb97be04332d0c4d39404933a944d444fc/subscribing-sdk-java/build.gradle#L13) in the Gradle file for the subscribing module/project. Is that not needed here?

QuintinWillison (changes requested):
Perhaps this PR should also include an addition to the root readme? (i.e. [here](https://github.com/ably/ably-asset-tracking-android#java-users))

ui-sdk-java/build.gradle
KacperKluka: The dependency you mentioned is used when we want to convert coroutines to `CompletableFuture`s 😉 Here we only convert `Flow`s to listeners which does not require this library 👍

ui-sdk-java/.gitignore
KacperKluka: You're right 👍 https://github.com/ably/ably-asset-tracking-android/pull/791/commits/09a8a03ff0434c27d104f4556b1020fad9ec6e54

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
davyskiba: Sorry, looks like I've misclicked the line 😞 
I've meant to ask what exactly would "handle" mean in the following lines. Maybe we could be make more specific comment as an example?
```
// handle new map marker position
// handle new camera position
```

QuintinWillison (approved):

PR #800, opened 2022-11-09
Safeguard continuation from being called multiple times

KacperKluka:
When using the `channel.on` or `connection.on` listeners with continuations we noticed that the `continuation.resume()` could be called multiple times as those listeners are not being cleared. I tried to remove the listeners with the `off()` method but it's not possible to do because we want to remove the listener from its inside but to do that we need a reference to that listener, which is not accessible from the inside. Therefore, I've used the `continuation.isActive` method which is `true` only when the continuation was not resumed.

The problem seems to happen because we're not removing the channel state listener and it's active until the channel gets removed. Therefore if while tracking the network connection will be lost for some time and the channel would go to the `suspended` state and then when the connection is back it will go to the `attached` once again, the channel state listener will be invoked once more and the `continuation.resume()` will be called for the second time.

Due to the above, I think it will be hard to test, as the `ably-android` is not easily mockable and I'm not sure if we can turn off the internet connection from an integration test.


QuintinWillison (approved):

davyskiba:
The code looks good, but I wonder if removing the event listener instead of safeguarding on continuation state wouldn't be a better approach. @KacperKluka wdyt?

KacperKluka:
Thanks @davyskiba for your suggestion 🙇 I was able to deregister the listener from inside it and now the usage of `continuation.isActive` is not needed. I was worried about what would happen when the operation is canceled before the listener gets removed but according to [docs](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/resume.html) when we try to resume a canceled continuation it will simply call its `onCancelation` callback 😉 I assume that if we don't specify this callback then by default we get a no-op one, however I've not found anything about it in the docs, as well as there's no information that you shouldn't call the continuation resume method more than once, which is a bit disappointing 😒

QuintinWillison (approved):
Nice work, @KacperKluka and @davyskiba.
I've enjoyed watching this solution evolve. 💪 🚀

davyskiba (approved):
Looks good :)

We could hide anonymous object implementation behind a function, but I'm not sure if that's a good idea 🤔 The call site would look nicer, but on the other hand, it would be a lambda with a switched context, which is unobvious behavior. I'll leave the decision to you, here's an example
```
    /**
     * Waits for the [channel] to change to the [ChannelState.attached] state.
     * If the [channel] state already is the [ChannelState.attached] state it does not wait and returns immediately.
     * If this doesn't happen during the next [timeoutInMilliseconds] milliseconds, then an exception is thrown.
     */
    private suspend fun waitForChannelReconnection(channel: Channel, timeoutInMilliseconds: Long = 10_000L) {
        if (channel.state.isConnected()) {
            return
        }
        try {
            withTimeout(timeoutInMilliseconds) {
                suspendCancellableCoroutine<Unit> { continuation ->
                    channel.on(getChannelStateListener { channelStateChange ->
                        if (channelStateChange.current.isConnected()) {
                            channel.off(this)
                            continuation.resume(Unit)
                        }
                    })
                }
            }
        } catch (exception: TimeoutCancellationException) {
            throw ConnectionException(ErrorInformation("Timeout was thrown when waiting for channel to attach"))
        }
    }

    private fun getChannelStateListener(onEvent: ChannelStateListener.(ChannelStateListener.ChannelStateChange) -> Unit) =
        object : ChannelStateListener {
            override fun onChannelStateChanged(channelStateChange: ChannelStateListener.ChannelStateChange) {
                onEvent(channelStateChange)
            }
        }
```

ikbalkaya (approved):
LGTM

PR #806, opened 2022-11-16
Allow to specify the remainPresentFor Ably transport parameter

KacperKluka:
Added the `remainPresentFor` parameter to the Ably configuration according to the [docs](https://ably.com/docs/realtime/presence#unstable-connections).

### DIFF
--- a/core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
+++ b/core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
@@ -15,6 +15,11 @@ data class ConnectionConfiguration(
      * Allows a non-default Ably environment to be used such as 'sandbox'.
      */
     val environment: String? = null,
+    /**
+     * Specifies for how long should the SDK remain present in a channel when the connection is gone.
+     * The minimum value can be 1 second (1000 milliseconds). If not set the default value will be used.
+     */
+    val remainPresentForMilliseconds: Long? = null,
 )
 
 typealias TokenRequestCallback = suspend (TokenParams) -> TokenRequest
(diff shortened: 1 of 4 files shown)


core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
```diff
@@ -15,6 +15,11 @@ data class ConnectionConfiguration(
      * Allows a non-default Ably environment to be used such as 'sandbox'.
      */
     val environment: String? = null,
+    /**
+     * Specifies for how long should the SDK remain present in a channel when the connection is gone.
+     * The minimum value can be 1 second (1000 milliseconds). If not set the default value will be used.
```
QuintinWillison, line 20: The last sentence, in its current form, is probably just adding confusion - I'm left asking the question "What is the default value?". It would be useful to make it clearer where that default is defined, probably with a link to [the documentation](https://ably.com/docs/realtime/presence#unstable-connections).

Also, given the fact that this SDK does nothing to enforce the 1000ms minimum, it's perhaps not appropriate to explicitly document that in the API as this is a service-side minimum... perhaps better left explained by the canonical documentation link I mentioned in my opening paragraph to this comment. 🤔

QuintinWillison (changes requested):

davyskiba (approved):
Code looks good :)

core-sdk/src/main/java/com/ably/tracking/connection/ConnectionConfigurationModels.kt
KacperKluka: Good idea 👍 https://github.com/ably/ably-asset-tracking-android/pull/806/commits/d9c298660e7742120a563ae14c95b5fc948af9b2

QuintinWillison (approved):

PR #813, opened 2022-11-17
Add more logs to the SDKs

KacperKluka:
I've added a bunch of logs to the SDKs:
- Publisher lifecycle logs
- Subscriber lifecycle logs
- Error logging for the Ably wrapper

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
@@ -74,11 +74,11 @@ internal data class PublisherBuilder(
 
     @RequiresPermission(anyOf = [ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION])
     override fun start(): Publisher {
-        logHandler?.v("$TAG Creating a publisher instance")
         if (isMissingRequiredFields()) {
             logHandler?.v("$TAG Creating a publisher instance failed due to missing required fields")
             throw BuilderConfigurationIncompleteException()
         }
+        logHandler?.v("$TAG Creating a publisher instance")
         // All below fields are required and above code checks if they are nulls, so using !! should be safe from NPE
         return DefaultPublisher(
             DefaultAbly(connectionConfiguration!!, logHandler),
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
@@ -7,6 +7,7 @@ import androidx.annotation.RequiresPermission
 import com.ably.tracking.BuilderConfigurationIncompleteException
 import com.ably.tracking.Resolution
 import com.ably.tracking.common.DefaultAbly
+import com.ably.tracking.common.logging.v
 import com.ably.tracking.connection.ConnectionConfiguration
 import com.ably.tracking.logging.LogHandler
 
@@ -71,7 +72,9 @@ internal data class PublisherBuilder(
 
     @RequiresPermission(anyOf = [ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION])
     override fun start(): Publisher {
+        logHandler?.v("Creating a publisher instance")
         if (isMissingRequiredFields()) {
+            logHandler?.v("Creating a publisher instance failed due to missing required fields")
             throw BuilderConfigurationIncompleteException()
         }
         // All below fields are required and above code checks if they are nulls, so using !! should be safe from NPE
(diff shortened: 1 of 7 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
```diff
@@ -71,7 +74,9 @@ internal data class PublisherBuilder(
 
     @RequiresPermission(anyOf = [ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION])
     override fun start(): Publisher {
+        logHandler?.v("$TAG Creating a publisher instance")
```
QuintinWillison, line 77: Looking at this message in context of required fields being missing then I think that might cause confusion when debugging. Perhaps consider moving this line to after the required fields check.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/SubscriberBuilder.kt
```diff
@@ -26,14 +29,17 @@ internal data class SubscriberBuilder(
         this.copy(logHandler = logHandler)
 
     override suspend fun start(): Subscriber {
+        logHandler?.v("$TAG Creating a subscriber instance")
```
QuintinWillison, line 32: Looking at this message in context of required fields being missing then I think that might cause confusion when debugging. Perhaps consider moving this line to after the required fields check.

publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
```diff
@@ -71,7 +74,9 @@ internal data class PublisherBuilder(
...
         if (isMissingRequiredFields()) {
+            logHandler?.v("$TAG Creating a publisher instance failed due to missing required fields")
```
QuintinWillison, line 79: I generally avoid log-before-throw as more often than not it ends up with a situation where code above this also logs the exception and then when debugging it's difficult to ascertain how many exceptions were actually thrown - whether caught or uncaught.

Therefore, I think you should remove this log statement entirely. It an app developer has managed to call `start()` but is unaware that it's thrown them back an exception then I think they have larger problems afoot. 😬

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/SubscriberBuilder.kt
```diff
@@ -26,14 +29,17 @@ internal data class SubscriberBuilder(
...
         if (isMissingRequiredFields()) {
+            logHandler?.v("$TAG Creating a subscriber instance failed due to missing required fields")
```
QuintinWillison, line 34: Similar to my comment on publisher `start()`, I would suggest removing this log line completely. The app should be catching the exception.

QuintinWillison (changes requested):
Furthermore, on the subject of logging failure/exception cases, I'm a bit worried that the `onError` implementations (`wrapInResultCallback`) that you've provided will also end up with confusing log output when the application itself also logs exceptions. There's also no way for the app developer to opt out of this behaviour, so my gut feeling is that this presents over-logging.

publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
QuintinWillison: I'm removing this objection, given I've now [created an issue](https://github.com/ably/ably-asset-tracking-android/pull/813#ref-issue-1454514942) to look at this across all of our SDKs.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/SubscriberBuilder.kt
QuintinWillison: I'm removing this objection, given I've now [created an issue](https://github.com/ably/ably-asset-tracking-android/pull/813#ref-issue-1454514942) to look at this across all of our SDKs.

publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt
KacperKluka: [0c143bd](https://github.com/ably/ably-asset-tracking-android/pull/813/commits/0c143bd404aa48d678d6b27fd29b8a82d8b31675)

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/SubscriberBuilder.kt
KacperKluka: [0c143bd](https://github.com/ably/ably-asset-tracking-android/pull/813/commits/0c143bd404aa48d678d6b27fd29b8a82d8b31675)

QuintinWillison (approved):

Issue #808, opened 2022-11-17
Fix crash when publisher tries to modify its internal state after being stopped


KacperKluka:
We've seen a fatal exception occur when using the publisher SDK. It looks like the publisher is stopped but then it itself tries to access its own state which leads to the `PublisherPropertiesDisposedException` being thrown. This should not happen as after stopping the publisher should perform no operations on its own.

```
Fatal Exception: w3.i0: Cannot access the publisher properties after it's disposed.
       at com.ably.tracking.publisher.DefaultCorePublisher$Properties.getTrackableStateFlows(CorePublisher.kt:700)
       at com.ably.tracking.publisher.DefaultCorePublisher$updateTrackableState$1.invokeSuspend(CorePublisher.kt:475)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:923)
```

ikbalkaya:
Just a question on this, Is there  a mechanism that prevents use of disposed properties? If not I suppose that might be a race condition issue where trackableStateflows still emits updates after it was stopped. I will investigate the potential of this happening

KacperKluka:
The mechanism you mention is the error throwing that happens 😉 When you try to access disposed properties they will throw the `PublisherPropertiesDisposedException`

PR #820, opened 2022-11-23
Check publisher state before emitting state on flow

ikbalkaya:
This is a workaround fix for https://github.com/ably/ably-asset-tracking-android/issues/808

After some investigation and a fruitful conversation with @KacperKluka, it turns out that `runBlocking` in `StopWorker` will block the processing of some coroutines which have started but not finished yet. This causes a race condition on accessing `trackableStateFlows`.  In this case when trackable state emission block is entered and `StopWorker` has started working, it is possible for `trackableStateFlows` gets in an inconsistent state before in the time it was accessed.

I was able to reproduce this adding a `delay()` inside the coroutine block and saw that crash happens when I add, remove a trackable and then stop the publisher. I was also able to see that this was not happening after I added the code that does fix.

This is a workaround fix that checks the state of publisher before accessing `trackableStateFlows` to protect the clients from crashing in such an edge case.

Closes  https://github.com/ably/ably-asset-tracking-android/issues/808

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -472,7 +472,11 @@ constructor(
         }
         if (newTrackableState != properties.trackableStates[trackableId]) {
             properties.trackableStates[trackableId] = newTrackableState
-            scope.launch { properties.trackableStateFlows[trackableId]?.emit(newTrackableState) }
+            scope.launch {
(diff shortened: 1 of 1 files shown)


QuintinWillison (approved):
I'll admit I don't have enough knowledge of the codebase to know for sure whether it is ok to read `properties.state` from within `scope.launch`, but given `properties.trackableStateFlows` was already being read then I assume it's ok. Looks good to me!

KacperKluka (approved):
LGTM 👍

PR #821, opened 2022-11-24
Timber debug log tree planted in both example apps

### LINKED ISSUE #818: Enable logging in the example apps by default
Right now when you use the example apps you have to manually enable the [Timber library](https://github.com/JakeWharton/timber). I think it should be enabled by default and we should add the `Timber.plant()` calls to both publisher and subscriber example apps to enable logs.

davyskiba:
In this PR, I've added the code required for Timber logging to work. We are using calls to Timber across the app for logging, but they do not do anything when no tree is planted.

Resolves #818

### DIFF
--- a/subscribing-example-app/src/main/java/com/ably/tracking/example/subscriber/SubscribingExampleApplication.kt
+++ b/subscribing-example-app/src/main/java/com/ably/tracking/example/subscriber/SubscribingExampleApplication.kt
@@ -1,0 +1,12 @@
+package com.ably.tracking.example.subscriber
+
+import androidx.multidex.MultiDexApplication
+import timber.log.Timber
+import timber.log.Timber.DebugTree
+
+class SubscribingExampleApplication : MultiDexApplication() {
+    override fun onCreate() {
+        super.onCreate()
+        Timber.plant(DebugTree())
+    }
+}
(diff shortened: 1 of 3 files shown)


subscribing-example-app/src/main/java/com/ably/tracking/example/subscriber/SubscribingExampleApplication.kt
```diff
@@ -0,0 +1,12 @@
+package com.ably.tracking.example.subscriber
+
+import androidx.multidex.MultiDexApplication
+import timber.log.Timber
+import timber.log.Timber.DebugTree
+
+class SubscribingExampleApplication : MultiDexApplication() {
```
KacperKluka, line 7: Do we need to use a `MultiDexApplication`? 🤔

KacperKluka (commented):
Looks good, just one question 😉

subscribing-example-app/src/main/java/com/ably/tracking/example/subscriber/SubscribingExampleApplication.kt
KacperKluka: This seems to be fixed in https://github.com/ably/ably-asset-tracking-android/pull/821/commits/cc86d0f8239af3e269d97591bf21cd1122b93594

KacperKluka (approved):
LGTM 👍

ikbalkaya (approved):
>  they do not do anything when no tree is planted.

I love that 😍

PR #825, opened 2022-12-01
Un-nest nested interfaces

ikbalkaya:
I have un-nested and renamed all interfaces that were nested under `ResolutionPolicy` that were causing an issue when Java users were adding a publisher.

For detail and more context please visit this [internal Slack thread](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1665130953392569)

Note to reviewers 
*  I quickly renamed interfaces that made sense at the time but please let me know if you think of better names 
* I haven't thoroughly checked the impact of this change on users but our tests seems to pass as a result of the change. I just needed to change a snippet in example Java code. But considering those interfaces were public, I wonder what other impact might we have on users apart from they would have to update the name of interfaces (package shouldn't change)


publishing-sdk/src/main/java/com/ably/tracking/publisher/HooksSubscriberSetListener.kt
```diff
@@ -0,0 +1,22 @@
+package com.ably.tracking.publisher
+
+/**
+ * A handler of events relating to the addition or removal of remote [Subscriber]s to a [Publisher] instance.
+ */
+interface HooksSubscriberSetListener {
```
KacperKluka, line 6: All the other interfaces that were nested in the `ResolutionPolicy` have now names starting with `ResolutionPolicy`, however, this one doesn't. Maybe for consistency, this should be called `ResolutionPolicySubscriberSetListener`? 🤔

publishing-sdk/src/main/java/com/ably/tracking/publisher/HooksTrackableSetListener.kt
```diff
@@ -0,0 +1,35 @@
+package com.ably.tracking.publisher
+
+/**
+ * A handler of events relating to the addition, removal and activation of [Trackable] objects for a
+ * [Publisher] instance.
+ */
+interface HooksTrackableSetListener {
```
KacperKluka, line 7: Like above, maybe for consistency, this should be called `ResolutionPolicyTrackableSetListener`? 🤔

KacperKluka (changes requested):
Overall it looks good, just a few naming suggestions 😉 

I wonder, are there some other nested interfaces in AAT API that we should be worried about? 🤔

davyskiba (approved):

ikbalkaya:
Closing this as this is an issue with Android studio and not an issue with this repo - Filed a bug report https://issuetracker.google.com/issues/261386526

PR #829, opened 2022-12-06
Set all location updates type to ACTUAL

KacperKluka:
We have disabled prediction in the Mapbox SDK so we will never get a `PREDICTED` location update, therefore all location updates should have the `ACTUAL` type set.

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt
@@ -192,14 +192,12 @@ constructor(
 
             override fun onEnhancedLocationChanged(enhancedLocation: Location, intermediateLocations: List<Location>) {
                 logHandler?.v("$TAG Enhanced location received: $enhancedLocation")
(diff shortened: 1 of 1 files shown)


davyskiba (approved):

QuintinWillison (changes requested):
I might have expected the `PREDICTED` value to disappear or be marked deprecated or something as part of this atomic change. 🤔

KacperKluka:
@QuintinWillison that's a very good [suggestion](https://github.com/ably/ably-asset-tracking-android/pull/829#pullrequestreview-1208249698) 🙇 I think that removing the `PREDICTED` value will be a breaking change. And if we were to make a breaking change regarding this I would prefer to [consider removing the `type` field entirely](https://github.com/ably/ably-asset-tracking-common/issues/69) (as it can only be `ACTUAL`) 🤔 Therefore I will only deprecate this value for now 👍 https://github.com/ably/ably-asset-tracking-android/pull/829/commits/4e496a050b4da64a3c27ad436a9254ed5627f012

QuintinWillison (approved):

PR #850, opened 2022-12-14
Rid R8 minify missing class warning

QuintinWillison:
As seen [in this run](https://github.com/ably/ably-asset-tracking-android/actions/runs/3693590989/jobs/6253753819#step:5:616):

```
WARNING:Missing classes detected while running R8. Please add the missing classes or apply additional keep rules that are generated in /home/runner/work/ably-asset-tracking-android/ably-asset-tracking-android/integration-testing-app/build/outputs/mapping/debug/missing_rules.txt.

WARNING:R8: Missing class com.google.auto.value.AutoValue$Builder (referenced from: com.mapbox.api.directions.v5.models.Admin$Builder and 38 other contexts)
Missing class com.google.firebase.messaging.FirebaseMessaging (referenced from: void io.ably.lib.push.ActivationContext.getRegistrationToken(io.ably.lib.types.Callback))
```

I've used the lines suggested in the `missing_rules.txt` generated on my local machine after a `./gradlew check assemble`.


davyskiba (approved):

integration-testing-app/proguard-rules.pro
```diff
@@ -23,10 +23,12 @@
 # To suppress Gradle build warning, we conservatively keep those classes and then supress the warnings
 -keep class com.google.auto.value.** { *; }
 -dontwarn com.google.auto.value.AutoValue
+-dontwarn com.google.auto.value.AutoValue$Builder
 
 -keep class com.google.firebase.installations.** { *; }
 -dontwarn com.google.firebase.installations.FirebaseInstallations
 -dontwarn com.google.firebase.installations.InstallationTokenResult
+-dontwarn com.google.firebase.messaging.FirebaseMessaging
```
ikbalkaya, line 31: Shouldn't this have its -keep counterpart at top 🤔 
`-keep class com.google.firebase.messaging.** { *; }
`

ikbalkaya (changes requested):
I think we might need to keep before suppress the warning in one of the changes

integration-testing-app/proguard-rules.pro
QuintinWillison: Looking at the logs from the assemble run for this PR there aren't any instances of `R8` warnings. I would rather not add this if it's not required, as all checks are passing and this is only for the integration testing app. Or did you spot something I'm missing? I included everything that the tools suggested I include (`missing_rules.txt`).

ikbalkaya: Of course, Other classes above this has this pattern of `keep` and `dontwarn` I thought this one would need the same

ikbalkaya (approved):

PR #851, opened 2022-12-14
Replace deprecated GitHub actions

QuintinWillison:
We've been seeing warnings like this in CI summaries for a while:

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-java@v2, actions/upload-artifact@v2

We've also been seeing warnings like this:

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Which I suspect are a side effect of the Node.js-12-based actions using the older `@actions/core` dependency. We'll see for sure once this pull request has run in CI and presented a summary. 🤞

### DIFF
--- a/.github/workflows/assemble.yml
+++ b/.github/workflows/assemble.yml
@@ -71,7 +71,7 @@ jobs:
           name: publishing-example-app-debug
           path: publishing-example-app/build/outputs/apk/debug/publishing-example-app-debug.apk
           if-no-files-found: error
- 
+
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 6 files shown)


davyskiba (approved):

ikbalkaya (approved):
LGTM

PR #855, opened 2022-12-16
Present Publisher as offline to Subscribers on loss of connectivity on the Subscriber's side

### LINKED ISSUE #833: Subscriber, upon losing connectivity, continues to show the Publisher as online
The connectivity loss is being correctly reflected in the trackable state, which does become 'offline', however the publisher presence remains indicated as 'online'.

The fix for this bug _might_ be in the underlying [core `ably-android` SDK](https://github.com/ably/ably-java/), with the working theory being that it is not sending a "presence leave" event for presence members in this scenario, therefore leaving the disconnected AAT Subscriber thinking that the publisher is online (based on presence events only). We need to look at our features spec and see if the client library SDK behaviour is defined in relation to whether SDKs should emit "leave" events for presence members when the Ably connection has been lost. We can then decide what to do next.

Plenty of conversation on this issue in [this Slack thread (internal)](https://ably-real-time.slack.com/archives/C01QU436CSZ/p1666278420602919).

QuintinWillison:
Please note that this pull request is in _Draft_ state as this work is incomplete. It's been pushed up so I can benefit from CI testing.

Fixes #833.


subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -143,37 +111,169 @@ private class DefaultCoreSubscriber(
 
     override fun subscribeForEnhancedEvents(presenceData: PresenceData) {
         ably.subscribeForEnhancedEvents(trackableId, presenceData) {
-            scope.launch { _enhancedLocations.emit(it) }
+            eventFlows.emitEnhanced(it)
```
ikbalkaya, line 114: I wonder what impact will this have now being called outside of scope 🤔

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -143,37 +111,169 @@ private class DefaultCoreSubscriber(
...
-            scope.launch { _nextLocationUpdateIntervals.emit(publisherResolution.desiredInterval) }
+            eventFlows.emitRaw(it)
         }
     }
 
     override fun notifyAssetIsOffline() {
-        scope.launch { _trackableStates.emit(TrackableState.Offline()) }
+        // TODO what is this method achieving, why is it not in normal flow?
+        // Perhaps related to: https://github.com/ably/ably-asset-tracking-android/issues/802
+        eventFlows.emit(TrackableState.Offline())
```
ikbalkaya, line 127: And this.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -143,37 +111,169 @@ private class DefaultCoreSubscriber(
...
     var presenceData: PresenceData,
+    private val eventFlows: EventFlows,
+
     override var isStopped: Boolean = false,
-    var isPublisherOnline: Boolean = false,
-    var trackableState: TrackableState = TrackableState.Offline(),
-    var lastConnectionStateChange: ConnectionStateChange =
+
+    private var presentPublisherMemberKeys: MutableSet<String> = HashSet(),
+    private var lastEmittedIsPublisherVisible: Boolean? = null,
```
ikbalkaya, line 138: I think it would be more readable for variable to start with 'is' (eg. `isLastEmittedPublisherVisible`)

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/ChangeResolutionWorkerTest.kt
```diff
@@ -1,41 +0,0 @@
...
-import com.ably.tracking.test.common.mockUpdatePresenceDataSuccess
-import io.mockk.mockk
-import io.mockk.verify
-import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.test.runBlockingTest
-import org.junit.Assert
-import org.junit.Test
-
-@ExperimentalCoroutinesApi
-internal class ChangeResolutionWorkerTest {
```
ikbalkaya, line 17: I would be interested to understand why this was removed. Isn't this checking `ChangeResolutionWorker` doing its expected work 🤔

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/DisconnectWorkerTest.kt
```diff
@@ -1,41 +0,0 @@
-package com.ably.tracking.subscriber.workerqueue.workers
```
ikbalkaya, line 1: And this

ikbalkaya (changes requested):
As this PR contains mostly refactoring code, I wasn't able to confirm where exactly the change happened 

I assume that's `isPublisherVisible` but wasn't sure - could you please provide an explanation     to PR description so this is more clear?

Also about tests, as far as I can see they do test the individual workers, I think they are still important as they protect the correct functioning of individual workers.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
QuintinWillison: The `scope` is now encapsulated within `eventFlows`.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
QuintinWillison: It was constructed to be read in two parts, as in "last emitted" _value of_ "isPublisherVisible". I would rather keep it like it is as it aligns with naming of siblings `lastEmittedTrackableState `, `lastConnectionStateChange` and `lastChannelConnectionStateChange`.

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/ChangeResolutionWorkerTest.kt
QuintinWillison: I explained in the opening commentary. These tests were a barrier to refactoring and I didn't feel they were adding any value to the maintainability of the codebase going forwards. I'm happy to hear any counter arguments on this, of course.

ikbalkaya: I would argue that tests verify that the refactored code performs correctly (how else we could verify?) and I understand when you refactor code, you need to refactor tests too .However I wouldn't see this as a barrier to refactor but a barrier that protects functioning. In terms of adding values, I think these add quite a lot of value as they test the correctness of the workers.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
ikbalkaya: Ah ok I see. But I still find 'is' being in the middle making the variable a bit unreadable. Can this not be something like `lastEmittedPublisherVisibleValue`?

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/SubscribeForPresenceMessagesWorkerTest.kt
```diff
@@ -1,125 +0,0 @@
...
-import io.mockk.CapturingSlot
-import io.mockk.mockk
-import io.mockk.slot
-import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.test.runBlockingTest
-import org.junit.Assert
-import org.junit.Test
-
-@ExperimentalCoroutinesApi
-internal class SubscribeForPresenceMessagesWorkerTest {
```
davyskiba, line 26: I think this test might be worth keeping, `SubscribeForPresenceMessagesWorker` has a few conditional branches inside

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/StartConnectionWorkerTest.kt
```diff
@@ -1,118 +0,0 @@
...
-import io.mockk.mockk
-import io.mockk.runs
-import io.mockk.verify
-import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.test.runBlockingTest
-import org.junit.Assert
-import org.junit.Test
-
-@ExperimentalCoroutinesApi
-internal class StartConnectionWorkerTest {
```
davyskiba, line 26: I think this test might be worth keeping, StartConnectionWorker has a few conditional branches inside

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/workerqueue/workers/StartConnectionWorker.kt
```diff
@@ -18,7 +18,7 @@ internal class StartConnectionWorker(
         doAsyncWork: (suspend () -> Unit) -> Unit,
         postWork: (WorkerSpecification) -> Unit
     ): SubscriberProperties {
-        subscriberInteractor.updateTrackableState(properties)
```
davyskiba, line 21: Could you remove `subscriberInteractor` passed in the constructor, as it is now unused?
Shouldn't CI check flag that? 🤔

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -143,37 +111,169 @@ private class DefaultCoreSubscriber(
...
+            }
+        }
+    }
+
+    fun emitEventsIfRequired() {
+        val isAPublisherPresent = (presentPublisherMemberKeys.isNotEmpty())
+
+        val trackableState = when (lastConnectionStateChange.state) {
+            ConnectionState.ONLINE -> {
+                when (lastChannelConnectionStateChange.state) {
```
davyskiba, line 184: Could handling for the `ConnectionState.ONLINE` case be extracted to a separate method to flatten the nesting?

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -143,37 +111,169 @@ private class DefaultCoreSubscriber(
...
+            lastEmittedTrackableState = trackableState
+            eventFlows.emit(trackableState)
+        }
+
+        // It is possible for presentPublisherMemberKeys to not be empty, even when we have no connectivity from our side,
+        // because we've had presence entry events without subsequent leave events.
+        // Therefore, from the perspective of a user consuming events from publisherPresenceStateFlow, what matters
+        // is what we're computing for isPublisherVisible (not the simple isAPublisherPresent).
+        val isPublisherVisible = (isAPublisherPresent && lastConnectionStateChange.state == ConnectionState.ONLINE)
+        if (null == lastEmittedIsPublisherVisible || lastEmittedIsPublisherVisible!! != isPublisherVisible) {
```
davyskiba, line 204: This condition could be reduced to `lastEmittedIsPublisherVisible != isPublisherVisible`, as `isPublisherVisible` cannot be `null`

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/ChangeResolutionWorkerTest.kt
davyskiba: I agree that testing if workers pass parameters properly does not add much value, but I think some of the removed tests were checking some conditional logic flow and are worth keeping

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -143,37 +111,169 @@ private class DefaultCoreSubscriber(
...
+        } else {
+            // PRESENT, ENTER or UDPATE
+            presentPublisherMemberKeys.add(presenceMessage.memberKey)
+            presenceMessage.data.resolution?.let { publisherResolution ->
+                pendingPublisherResolutions.add(publisherResolution)
+            }
+        }
+    }
+
+    fun emitEventsIfRequired() {
```
davyskiba, line 179: At first I was a bit confused why all of those events are aggregated under a single function, then I realised that all "state" values are actually linked together. Maybe we could rename this function to emphasize that only "state" related events are considered (no location updates)?

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
```diff
@@ -143,37 +111,169 @@ private class DefaultCoreSubscriber(
...
+        lastConnectionStateChange = stateChange
+        emitEventsIfRequired()
+    }
+
+    fun updateForChannelConnectionStateChangeAndThenEmitEventsIfRequired(stateChange: ConnectionStateChange) {
+        lastChannelConnectionStateChange = stateChange
+        emitEventsIfRequired()
+    }
+
+    fun updateForPresenceMessage(presenceMessage: PresenceMessage) {
```
davyskiba, line 161: In both cases, right after calling this method, the caller should call `emitEventsIfRequired.` Maybe we could integrate it inside to prevent mistakes?
Like in the case of `updateForConnectionStateChangeAndThenEmitEventsIfRequired`

Ideally, I would make `emitEventsIfRequired` private, but this would require creating a proxy method like `emitInitialEvents` for usage in `StartConnectionWorker` wdyt?

davyskiba (changes requested):

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
davyskiba: Right, and if `lastEmittedIsPublisherVisible` is null, it is not going to be equal to `isPublisherVisible,` which cannot be null

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/workerqueue/workers/StartConnectionWorker.kt
QuintinWillison: Thanks, good spot! I'm going to fix this. In respect of why it wasn't flagged in CI, I've created https://github.com/ably/ably-asset-tracking-android/issues/858.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
QuintinWillison: It could, however I don't think "flatten the nesting" needs to be a goal in and of itself, especially when this block of logic is only used in this one place. Perhaps a little subjective, but in my opinion the nesting is not too deep here and the code is readable enough. Therefore, my preference is to leave it as is.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
QuintinWillison: I've gone with a slightly different name to what you suggest, more in keeping with my other field names in this class: https://github.com/ably/ably-asset-tracking-android/pull/855/commits/afe49dda29967b3dd4bcad7e83323a2f32927d17

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/workerqueue/workers/StartConnectionWorker.kt
QuintinWillison: https://github.com/ably/ably-asset-tracking-android/pull/855/commits/b0cd85ec514052e415e34137338737274b1c7347

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
QuintinWillison: Renamed in https://github.com/ably/ably-asset-tracking-android/pull/855/commits/8ea1a1b10ac75ecffa83ee6452871ca0ea529e64.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
QuintinWillison: Integrated inside as suggested in https://github.com/ably/ably-asset-tracking-android/pull/855/commits/078d33da7f9c58dfdd09499b9790f1dc2380d787.

In respect of making that method private, I do agree, however perhaps that's a refactor for another day as it's not doing any major harm being present visible in current (internal) form.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/CoreSubscriber.kt
QuintinWillison: Got you and thanks for persevering in explaining to me! 😆 I was being dim. Addressed in https://github.com/ably/ably-asset-tracking-android/pull/855/commits/da21c9b3a7b685ffe10fbc5c6045253e0389efb8.

QuintinWillison:
Recognising the push back from both @ikbalkaya and @davyskiba on my removal of tests, I'm going to move this PR back to _Draft_ state while I look at those again.

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/ChangeResolutionWorkerTest.kt
QuintinWillison: Tests reinstated in https://github.com/ably/ably-asset-tracking-android/pull/855/commits/fca078904b8233ba3b0e62cb96de112c41c5ed88.

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/DisconnectWorkerTest.kt
QuintinWillison: Tests reinstated in https://github.com/ably/ably-asset-tracking-android/pull/855/commits/fca078904b8233ba3b0e62cb96de112c41c5ed88.

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/SubscribeForPresenceMessagesWorkerTest.kt
QuintinWillison: Tests reinstated in https://github.com/ably/ably-asset-tracking-android/pull/855/commits/fca078904b8233ba3b0e62cb96de112c41c5ed88.

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/workerqueue/workers/StartConnectionWorkerTest.kt
QuintinWillison: Tests reinstated in https://github.com/ably/ably-asset-tracking-android/pull/855/commits/fca078904b8233ba3b0e62cb96de112c41c5ed88.

davyskiba (approved):

PR #856, opened 2022-12-19
Refactor Publisher EventQueue to match Subscriber

### LINKED ISSUE #781: Refactor Publisher EventQueue to match Subscriber

davyskiba:
In this PR I'm converting the existing worker queue to the approach introduced in the Subscriber queue. The main differences are protecting properties object thanks to copying it before passing to a worker and inlining all results and result handlers into worker object, which makes it easier to follow. Additionally, I've extracted `PublisherInteractor` interface from `CorePublisher` interface, to separate functions used by workers from functions used by `DefaultPublisher.`

resolves #781


QuintinWillison (commented):
When I was working on #855 the biggest challenge for me was understanding the role of the internal `SubscriberInteractor` interface. It wasn't documented and I could not understand what it was gaining the codebase. I ended up removing the the `update` family of methods from it, replacing calls to the interactor with direct calls to the subscriber properties, with no obvious disadvantage to the codebase that I could see.

I notice that we now have a `PublisherInteractor` interface added in this pull request, which also has an `update` family of methods. This is now at odds with the work I did on the subscriber side, so I think we should consider how to proceed as it doesn't make sense for these "interactor siblings" to have different scopes of responsibility. I recognise that we worked on these two pieces in parallel to one another, however I think we need to agree how to converge in order to achieve adequate symmetry.

davyskiba:
@QuintinWillison Interactor interface was introduced to separate methods exposed to the `DefaultPublisher` via `CorePublisher` interface. The idea was to extract methods used by workers to the `PublisherInteractor` interface.  Thanks for pointing that out. I'll add a clarifying comment to those interfaces.

I didn't take a closer look at those methods to limit the refactor scope as much as possible. I think it would be an excellent next step to reorganize the `update` family to match what you did on the subscriber side

QuintinWillison (approved):
I've not looked at every line of change in detail, however on a relatively brief read through this looks good to me. Thanks - quite an undertaking but really good to conform the codebase like this. 💪 🚀

PR #864, opened 2022-12-27
location: sanitize locations from mapbox

mschristensen:
This commit ensures that new Location objects that have an invalid lat/lng/altitude component throw an IllegalArgumentException. Additionally, a sanitize() method is made available on the Location object that will gracefully convert any invalid values to a sensible default.


mschristensen:
@KacperKluka this is a WIP, but wanted some early feedback:

Am I on the right lines here with how to do this in an idiomatic Kotlin way? Or is there a better approach?

I think one potential problem with the require pattern could be that other instantiations of a Location with non-finite values for lat/lng/altitude could now throw (eek). So, perhaps we could return a special value, like a `data class ValidationError(val message: String)`, from the `sanitize()` method?

Still to add: similar logic for enhanced location updates and some tests.

mschristensen:
@KacperKluka I have just updated with a new approach that doesn't rely on throwing an exception. Any feedback appreciated :)

core-sdk/src/main/java/com/ably/tracking/LocationUpdateModels.kt
```diff
@@ -117,6 +119,60 @@ data class Location(
...
+            if (speed.isFinite()) speed else 0.0f,
+            time,
+        )
+    }
+
+    /**
+     * Utility function that returns a new, valid Location with no non-finite members.
+     * If the validation errors list is empty, the returned Location is non-null and valid; otherwise it is null.
+     */
+    fun validate(): Pair<Location?, ArrayList<LocationValidationError>> {
```
KacperKluka, line 142: While it works, using the `Pair` is a bit hacky 😉 I'd suggest using `Result<Location>` instead and then instead of `LocationValidationError` create an exception that holds a list of String messages. Also, I'm wondering whether we really need to return a new `Location` that is an exact copy of the current location from this method 🤔 Perhaps we could return the same location given that we are not updating its fields?

After applying all suggestions it would look something like this:
```kotlin
class LocationValidationException(val errors: List<String>): Exception("Failed to validate a location object")

data class Location(
    // ...
) {
    fun validate(): Result<Location> {
        // ...
        return if (errs.isNotEmpty()) {
            Result.failure(LocationValidationException(errs))
        } else {
            Result.success(this)
        }
    }
}
```

core-sdk/src/main/java/com/ably/tracking/LocationUpdateModels.kt
```diff
@@ -117,6 +119,60 @@ data class Location(
...
+        val errs: ArrayList<LocationValidationError> = ArrayList<LocationValidationError>()
```
KacperKluka, line 143: In Kotlin we tend to use lists instead of arrays, so I'd change this to a `MutableList` 😉 Also I think the less abbreviations the better so I'd change the name to `errors` or even `validationErrorMessages` 🤔 
```suggestion
        val errors: MutableList<LocationValidationError> = mutableListOf()
```

publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt
```diff
@@ -356,7 +356,12 @@ internal class DefaultMapbox(
         object : LocationObserver {
             override fun onNewRawLocation(rawLocation: android.location.Location) {
                 logHandler?.v("$TAG Raw location received from Mapbox: $rawLocation")
-                locationUpdatesObserver.onRawLocationChanged(rawLocation.toAssetTracking())
+                val (atLocation, errs) = rawLocation.toAssetTracking().sanitize().validate()
```
KacperKluka, line 359: I think that the `sanitize()` and `validate()` steps are always wanted when we convert a `Location` to an AAT location, so I'd put those two methods inside the `toAssetTracking()` mapping function, WDYT? 🤔

KacperKluka (commented):
Nicely done, I like the fact that you managed to get rid of the `require` calls in the constructor 👏 A few suggestions from me 😉

jaley:
Given that there's little time for back and forth validation of this fix with real users, would it possible to reproduce the downstream crash in a unit test, then confirm that this PR fixes it?

Had a look around at `CorePublisherLocationUpdatesPublishingTest` seems like a relevant place for a "santizes location data" or similar unit test, there's already a mock of the Mapbox dependency there to inject invalid data.

WDYT, @KacperKluka & @mschristensen ?

mschristensen:
@KacperKluka  A couple of checks are not passing, but these appear to be flakes unrelated to this work, afaict (they differ each time they're run)?

publishing-sdk/src/test/java/com/ably/tracking/publisher/MapboxTest.kt
```diff
@@ -0,0 +1,529 @@
...
+
+@SuppressLint("MissingPermission")
+class MapboxTest {
+    private val mockLocationUpdatesObserver = mockk<LocationUpdatesObserver>()
+    private val mapboxLocationObserverProvider =
+        MapboxLocationObserverProvider(null, "MapboxTest")
+
+    @Test
+    fun `Should forward valid raw location update`() {
+        val mapboxLocationObserver =
```
KacperKluka, line 20: Since those tests are quite long it would be great to have the [given/when/then](https://en.wikipedia.org/wiki/Given-When-Then) structure on them, however, that's not a critical problem, so feel free to ignore this suggestion 😉

publishing-sdk/src/test/java/com/ably/tracking/publisher/MapboxTest.kt
```diff
@@ -0,0 +1,529 @@
...
+                    1.0f,
+                    1.0f,
+                    1
+                )
+            )
+        }
+        verify {
+            mockLocationUpdatesObserver.onRawLocationChanged(
+                withArg {
+                    Assert.assertTrue(it.latitude == 1.0)
```
KacperKluka, line 39: Generally, it is better to use `Assert.assertEquals(1.0, it.latitude)` because in case of a failed assertion we receive more info about what went wrong. However, this doesn't affect the tests, so feel free to ignore this suggestion and leave the code as it is 😉

KacperKluka (changes requested):
Almost there 👍 A few more suggestions 😉

core-sdk/src/main/java/com/ably/tracking/LocationUpdateModels.kt
```diff
@@ -117,6 +119,55 @@ data class Location(
...
+            altitude,
+            if (accuracy.isFinite()) accuracy else 0.0f,
+            if (bearing.isFinite()) bearing else 0.0f,
+            if (speed.isFinite()) speed else 0.0f,
+            time,
+        )
+    }
+
+    /**
+     * Utility function that returns a successful result with the Location iff. it is valid.
```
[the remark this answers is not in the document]
mschristensen: Ah, I meant [if and only if](https://en.wikipedia.org/wiki/If_and_only_if#:~:text=In%20logic%20and%20related%20fields,true%20or%20both%20are%20false.), but will update for clarity!

publishing-sdk/src/test/java/com/ably/tracking/publisher/MapboxTest.kt
mschristensen: I'll skip this as I'm just making these changes in a spare ten minutes I have, but feel free to make them if you would prefer them to be there :)

publishing-sdk/src/test/java/com/ably/tracking/publisher/MapboxTest.kt
mschristensen: I'll skip this as I'm just making these changes in a spare ten minutes I have, but feel free to make them if you would prefer them to be there :)

KacperKluka (approved):
LGTM 👍

davyskiba (approved):

Issue #862, opened 2022-12-23
Sending presence leave event times out whilst still connected


paddybyers:
When a delivery has completed, the publisher sends a presence leave event on the tracking channel and then detaches from the channel.

This is typically visible in the logs as “`zo0.b: leave();`” followed by “`io.ably.lib.transport.b: onMessage()`” with the returned ACK, for example:

```
18-12-22 19:59:07.702: zo0.b: leave(); channel = tracking:bo:22606154
18-12-22 19:59:07.702: zo0.b: update(); channel = tracking:bo:22606154
18-12-22 19:59:07.703: io.ably.lib.transport.WebSocketTransport: send(); action = presence
18-12-22 19:59:07.703: io.ably.lib.transport.WebSocketTransport: send(): presence: {"action":14,"channel":"tracking:bo:22606154","count":0,"flags":0,"msgSerial":339,"presence":[{"data":"{\"rawLocations\":false,\"type\":\"PUBLISHER\"}","action":3}],"timestamp":0}
18-12-22 19:59:07.923: io.ably.lib.transport.b: onMessage() ( ... ): ack: {"action":1,"count":1,"flags":0,"msgSerial":339,"timestamp":0}
```

From log redacted_08147607dc6202d1cb65ef57cdc42d53_error_session_639FA682023000012C9EE2B941CCBE13_DNE_1_v2.log however, the leave event was sent but no ACK was received:

```
18-12-22 19:48:58.570: zo0.b: leave(); channel = tracking:bo:22607777
18-12-22 19:48:58.571: zo0.b: update(); channel = tracking:bo:22607777
18-12-22 19:48:58.572: io.ably.lib.transport.WebSocketTransport: send(); action = presence
18-12-22 19:48:58.589: io.ably.lib.transport.WebSocketTransport: send(): presence: {"action":14,"channel":"tracking:bo:22607777","count":0,"flags":0,"msgSerial":1,"presence":[{"data":"{\"rawLocations\":false,\"type\":\"PUBLISHER\"}","action":3}],"timestamp":0}
[...]

Issue #859, opened 2022-12-21
Publisher apps reporting "Timeout was thrown when waiting for channel to attach"


QuintinWillison:
We have a customer who is reporting seeing [this exception](https://github.com/ably/ably-asset-tracking-android/blob/a75fa17dcba987d86e726c7f937f740cb5b6f8c5/common/src/main/java/com/ably/tracking/common/Ably.kt#L890) quite a lot via their Crashlytics reports:

```
Non-fatal Exception: s3.c: Timeout was thrown when waiting for channel to attach
       at com.ably.tracking.common.DefaultAbly.waitForChannelReconnection(Ably.kt:890)
       at com.ably.tracking.common.DefaultAbly.access$waitForChannelReconnection(Ably.kt:247)
       at com.ably.tracking.common.DefaultAbly$waitForChannelReconnection$1.invokeSuspend(Ably.kt)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:33)
       at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:102)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
       at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
       at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:664)
```

[...]

PR #866, opened 2022-12-28
Networking connectivity integration test using local proxy

jaley:
@cruickshankpg here's an empty test and a place to put the proxy implementation. Let's discuss the interface we're working to initial so we can both get started.


jaley:
@cruickshankpg - this will now run a simple integration test, which publishes a bunch of location updates for a single Trackable, using an injected `AblyRealtime` instance. This should be enough for you to put a proxy in the middle and confirm that you can break the test by interfering with traffic? Can also disable TLS as required.

android-test-common/src/main/java/com/ably/tracking/test/android/common/AblyProxy.kt
```diff
@@ -0,0 +1,199 @@
...
+                } else if (payloadLen == 127U) {
+                    dataOff += 4
+                    payloadLen = bigEndianConversion(buff.copyOfRange(2, 10), logHandler)
+                }
+
+                val op = buff[0].toUInt().and(0x0Fu)
+                if (op<3u) {
+                    logHandler("PROXY-MSG: payload length: " + payloadLen + " data offset: " + dataOff + " buff len: " + bytesRead)
+
+                    val isMasked = buff[1].toInt().and(0x01) == 1
```
lmars, line 144: From the diagram above, it looks like MASK is the MSB, so should this be `.and(0x80) == 0x80`?

Saying this though, is there not a WebSocket library out there we can use?

jaley:
This branch now has failing tests, which I believe reproduce some of our open issues:

* `faultDuringTracking[TcpConnectionRefused]` currently fails because `presence.leave()` times out after two minutes if you call `publisher.remove()` while connectivity is interrupted. I believe this is similar to what's happening in #862, though it's important we cover this with more types of fault when those are implemented. cc: @lawrence-forooghian 
* `faultBeforeAddingTrackable[TcpConnectionRefused]` fails after 10 seconds because we throw `ConnectionException` if initial requests to track occur while offline, this is #871 , cc: @davyskiba 
* `faultDuringTracking[TcpConnectionUnresponsive]` fails because of `com.ably.tracking.ConnectionException: Timeout was thrown when waiting for channel to attach`, which looks very much like #859, though I think even with this fixed there may be more failures to come. (no assignee yet)

I've not yet implemented a Layer 7 proxy, which is needed for the rest of the high priority fault types, so I'll try to get that done next. Tomorrow is really busy for me, but I'll see what I can do.

It might be a good idea to try to get this reviewed and ready to merge before then, if anyone wants to use the tests we have so far as a reproduction of the issues above to validate fixes? (Assuming that turns them green...)

Word of warning -- these are slow tests when they're failing. Generally, the pass in 2-3 seconds each, but waiting for `ably-java` timeouts when things are going wrong takes a few minutes. We may be able to tweak some of the timeouts in `ClientOptions` to speed up testing, but sadly I think some important ones are not client-configurable, e.g. the time taken for a channel to move to suspended state.

QuintinWillison:
At https://github.com/ably/ably-asset-tracking-android/pull/866/commits/1e96b3a3d5cf19f1433ebd24e17ddd6f1def8bcd, locally from Android Studio, I am seeing:

Screenshot 2023-01-11 at 17 49 07

With the following failures:

| Test | Failure |
| ---- | ------- |
| faultDuringTracking[TcpConnectionRefused] | com.ably.tracking.ConnectionException: failed to connect to localhost/127.0.0.1 (port 13579): connect failed: ECONNREFUSED (Connection refused) |
| faultBeforeAddingTrackable[TcpConnectionRefused] | java.lang.AssertionError: Expectation 'attempt to add active Trackable while fault active' did not result in success. |
| faultDuringTracking[TcpConnectionUnresponsive] | com.ably.tracking.ConnectionException: Timeout was thrown when waiting for channel to attach |
| faultBeforeAddingTrackable[TcpConnectionUnresponsive] | java.lang.AssertionError: Expectation 'attempt to add active Trackable while fault active' did not result in success. |

@jaley Is this what you would expect at this stage with this pull request? If so, then I would like to annotate those tests as ignored for now to make this pull request green and mergeable to `main` branch.

jaley:
Nice work, @QuintinWillison! 

Yes, I can confirm that I get the exact same failures, even on the #886 branch, as I don't think any of the relevant fixes have landed in main yet. In case you want to add comments/etc as to why these tests are ignored:

* `faultDuringTracking[TcpConnectionRefused]` - this is #859 , which @ikbalkaya has linked to ably/ably-java#885
* `faultBeforeAddingTrackable[TcpConnectionRefused]` - @davyskiba has PR #883 open, which should resolve this error
* `faultDuringTracking[TcpConnectionUnresponsive]` - #859 again
* `faultBeforeAddingTrackable[TcpConnectionUnresponsive]` - also fixed by #883

(Of course, there's no guarantee that there won't be more issues masked by those...)

android-test-common/src/main/java/com/ably/tracking/test/android/common/AblyProxy.kt
QuintinWillison: I'm resolving this on @lmars' behalf, given that the code being commented on no longer exists in the changes being presented by this pull request.

AndyTWF (approved):

publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
```diff
@@ -0,0 +1,520 @@
...
+            )
+        }
+    }
+
+    /**
+     * Performs the given async (suspending) operation in a runBlocking, attaching the
+     * returned StateFlow<TrackableState> to the given receiver, then waits for expectations
+     * to be delivered (or not) before cleaning up.
+     */
+    private fun waitForStateTransition(
```
davyskiba, line 237: This way of asserting state transitions is quite complicated, and it would be great if we could simplify it a bit 🤔 
I'll see if this can be improved quickly otherwise, we can improve it after merging

davyskiba (commented):
I have a few cosmetic

publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
QuintinWillison: I think deferring fixing this might be preferable, as it appears to function correctly and we do have https://github.com/ably/ably-asset-tracking-android/pull/886 further up the stack which is reliant upon this work in its current state. I'm keen to get this pull request landed to `main` branch soon.

davyskiba (approved):

publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
QuintinWillison: I've created an issue for this, so we can defer to later and allow this PR to merge now: https://github.com/ably/ably-asset-tracking-android/issues/901

.github/workflows/check.yml
```diff
@@ -64,13 +64,6 @@ jobs:
           name: java-version-${{ matrix.java-version }}-publishing-sdk-build-reports
           path: publishing-sdk/build/reports
 
-      - uses: actions/upload-artifact@v3
-        name: Build Reports for subscribing-example-app
```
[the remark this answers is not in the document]
QuintinWillison: I've created an issue so we don't forget to look at this later: https://github.com/ably/ably-asset-tracking-android/issues/902

publishing-sdk/src/main/java/com/ably/tracking/publisher/debug/FlowLocationEngine.kt
```diff
@@ -0,0 +1,21 @@
...
+import kotlinx.coroutines.flow.onEach
+
+internal class FlowLocationEngine(
+    flow: Flow<Location>,
+    logHandler: LogHandler?,
+) : BaseLocationEngine(logHandler) {
+    init {
+        flow.onEach {
+            onLocationEngineResult(LocationEngineResult.create(it.toAndroid()))
+        }.launchIn(CoroutineScope(Dispatchers.IO))
```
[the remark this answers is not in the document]
QuintinWillison: I've created an issue so we don't forget to improve this later: https://github.com/ably/ably-asset-tracking-android/issues/903

PR #872, opened 2022-12-30
Handle unexpected exceptions in the worker queue

KacperKluka:
Added a few `try/catch`es to make sure no unexpected exception fails the `WorkerQueue`. Instead, when such exception happens, the `Worker` will be notified about it via an appropriate method so it can react to the exception. Right now, the default reaction is to call the callback with a failure with the unexpected exception or to do nothing if there is no callback to call.

In the future it would be a good idea to add specific behaviour to workers that would rollback any changes they made to the `Properties` when this unexpected exception is thrown. Otherwise, we may end up in an unexpected SDK state from which it won't be able to recover. An example might be setting the state to `CONNECTING` and then never switching it back due to an unexpected error.


common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
```diff
@@ -0,0 +1,136 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.runBlocking
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+
+typealias TestWorkerSpecificationType = Unit
+
+class WorkerQueueTest {
+    private lateinit var workerQueue: WorkerQueue<Properties, TestWorkerSpecificationType>
```
davyskiba, line 22: Can this be instantiated here instead of being `var`?
Would be great to drop `setup` and `cleanup` methods if possible

common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    private fun mockWorkerQueueStopped() {
+        every { properties.isStopped } returns true
+    }
+
+    private fun mockWorkerQueueStarted() {
+        every { properties.isStopped } returns false
+    }
+
+    private fun mockAllWorkers(worker: Worker<Properties, TestWorkerSpecificationType>) {
+        every { workerFactory.createWorker(Unit) } returns worker
```
davyskiba, line 124: Can this get inlined into the mock definition? I think this configuration is the same for all tests 🤔

davyskiba:
I think rolling back changes to properties is not necessary - each worker is modifying its copy of properties, which the queue will not hold if an exception is thrown. Unfortunately, the current `PublisherProperties.copy()` implementation is doing only shallow copy, to harden it further, we should implement copying of non-primitive fields like `DuplicateTrackableGuard`

common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/872/commits/7e18bb70f0937dc496253ec61db08f1bb1a3b3aa

common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/872/commits/c910c9171269c3ef4e369eb4f9356cd276800c02

KacperKluka:
> I think rolling back changes to properties is not necessary - each worker is modifying its copy of properties, which the queue will not hold if an exception is thrown.

True but it gets complicated when a worker's synchronous work updates the properties and expects that the asynchronous work will post a new worker that will change the properties as well. For example, the `AddTrackableWorker` can change the SDK state to `CONNECTING` and then delay all other `AddTrackableWorker`s until the first one's async work completes with either a success or a failure. If the async work then fails with an unexpected error, the SDK will get stuck in this `CONNECTING` state forever 😅 So we do need to consider some edge cases and write proper rollbacks for them 🤔

davyskiba (approved):

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerQueue.kt
```diff
@@ -52,6 +59,10 @@ class WorkerQueue<PropertiesType : Properties, WorkerSpecificationType>(
                     try {
                         asyncWork()
                     } catch (unexpectedException: Exception) {
+                        logHandler?.e(
+                            "Unexpected error thrown from the asynchronous work of ${worker.javaClass.simpleName}",
```
davyskiba, line 63: Maybe change to "Unexpected exception (...)"?

davyskiba (approved):

common/src/main/java/com/ably/tracking/common/workerqueue/WorkerQueue.kt
KacperKluka: https://github.com/ably/ably-asset-tracking-android/pull/872/commits/040e0fa3d73c22932570d1932dd45d3a73d10e34

davyskiba (approved):

ikbalkaya (approved):
We had a sync call about the impact of this change in terms of state consistency if a worker fails. @KacperKluka  pointed out that this change will not guarantee the consistency but will reduce the likelihood of it getting broken which is an improvement. Provided he will also create another PR that will improve this change a bit more I am happy with the changes. 

I think we can improve APIs in the future but this is not the most pressing issue at the moment.

Issue #873, opened 2023-01-03
Investigate `ConnectionException` thrown by `Publisher.stop` and whether we can remove it


lawrence-forooghian:
It’s not clear for callers of `Publisher.stop` what it means if it [throws `ConnectionException`](https://github.com/ably/ably-asset-tracking-android/blob/d957d6c03d157f5cfdbb8e93708a903bd98e35a2/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt#L124), or how they should respond to it. For example, do they need to try calling it again, or is the publisher now in some terminal broken state?

Investigate whether it's possible to remove this exception, like we removed `TimeoutCancellationException` in #874. If not, describe the implications of this exception being thrown.

jaley:
@lawrence-forooghian, With this change and #871 , will there be anything remaining in the Publisher API that we think should still through `ConnectionException`? Feels like we're converging on just making this API fully async where error handling always requires callers to monitor `StateFlows`. I guess for `stop()` there isn't an obvious thing to monitor, but I agree that it's confusing for the caller if this operation appears to fail.

lawrence-forooghian:
After the changes you've mentioned, I think the only part of the publisher API that can throw a `ConnectionException` is `Publisher.Builder.start`. I’ve created https://github.com/ably/ably-asset-tracking-android/issues/876 to look into this.

paddybyers:
I don't think there's any value in throwing an exception in this case. If the `leave()` fails or times out, then the publisher will still go on to clean up the attachment and (assuming this is the last trackable), the connection. It will only fail to do that tidyup if the connection itself is broken, in which case it's all moot.

KacperKluka:
For the record:
While we declared that the `ably.close()` method throws a `ConnectionException` in reality it would never do that. That's because some time ago (namely as a side-effect of https://github.com/ably/ably-asset-tracking-android/issues/170) we changed the method that stops the connection in a way that it would never throw an exception but rather wrap it in a `Result`.

Nevertheless, the `stop()` method could still hang on the `ably.close()` indefinitely due to us waiting for the connection state to change to either `closed` or `failed`, so we should add a timeout that prevents the SDK to freeze while waiting for it to happen.

PR #874, opened 2023-01-03
Remove `TimeoutCancellationException` from `Publisher.stop`

### LINKED ISSUE #862: Sending presence leave event times out whilst still connected

lawrence-forooghian:
## Background

It’s not clear for callers of `Publisher.stop` what it means if it throws `TimeoutCancellationException`, or how they should respond to it. For example, do they need to try calling it again, or is the publisher now in some terminal broken state?

(The same questions also apply if the publisher throws `ConnectionException`; I’ve created #873 for addressing this.)

## What’s changed

This comes from the change suggested by @lmars in https://github.com/ably/ably-asset-tracking-android/issues/862#issuecomment-1366083580. 

We remove the operation-wide timeout parameter, and instead expose a parameter to give the caller control over only the presence leave operation.

We explain what will happen if this timeout elapses and how this affects the publisher’s online status from the point of view of subscribers.

## API considerations

From 944d88461cb36318032136adcf3b085c8e4da3d9:

> This is technically a breaking change to the publisher SDK API, both in terms of semantics and parameter name. I’ll start a discussion on the pull request to decide how we want to address this, and then update this commit message.

My change to `UPDATING.md` anticipates this just being a minor version bump. Do we think this warrants a major version bump?

## Reviewing

Please review commit-by-commit.

## Issues closed

Closes #862.

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -1,5 +1,23 @@
 # Upgrade / Migration Guide
 
+## Version 1.5.1 to 1.6.0
+
+### Removed user-specified timeout from `Publisher.stop`
+
+It is no longer possible to specify a timeout for the `Publisher.stop` operation. The `timeoutInMilliseconds` parameter has been deprecated and will be ignored.
+
+If you do not wish to wait indefinitely for this operation to complete, we recommend using [Kotlin’s `withTimeout`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout.html) to implement a timeout yourself. For example:
+
+```kotlin
+try {
+    withTimeout(timeMillis = 20_000) {
+        publisher.stop()
+    }
+} catch (e: TimeoutCancellationException) {
+    // Choose how to you want to respond to the timeout, by example for calling publisher.stop() again to retry
+}
+```
+
 ## Version 1.3.0 to 1.4.0
 
 ### Token based auth configuration
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -18,7 +18,6 @@ import com.ably.tracking.connection.ConnectionConfiguration
 import com.ably.tracking.logging.LogHandler
 import com.google.gson.Gson
 import io.ably.lib.realtime.AblyRealtime
-import io.ably.lib.realtime.Channel
 import io.ably.lib.realtime.ChannelState
 import io.ably.lib.realtime.ChannelStateListener
 import io.ably.lib.realtime.CompletionListener
@@ -317,7 +316,7 @@ constructor(
      */
     private suspend fun enterChannelPresence(channel: AblySdkRealtime.Channel, presenceData: PresenceData) {
         try {
-            channel.enterPresenceSuspending(presenceData)
+            enterPresenceSuspending(channel, presenceData)
         } catch (connectionException: ConnectionException) {
             if (connectionException.errorInformation.code == AUTH_TOKEN_CAPABILITY_ERROR_CODE) {
                 val renewAuthResult = renewAuthSuspending()
@@ -326,8 +325,8 @@ constructor(
                     logHandler?.w("$TAG Failed to renew auth while entering the presence of channel ${channel.name}", it.toTrackingException())
                     throw it.toTrackingException()
                 }
-                channel.attachSuspending()
-                channel.enterPresenceSuspending(presenceData)
+                attachSuspending(channel)
+                enterPresenceSuspending(channel, presenceData)
             } else {
                 logHandler?.w("$TAG Failed to enter the presence of channel ${channel.name}", connectionException)
                 throw connectionException
@@ -367,7 +366,7 @@ constructor(
                 scope.launch {
                     try {
                         if (channel.isDetachedOrFailed()) {
-                            channel.attachSuspending()
+                            attachSuspending(channel)
                         }
                         enterChannelPresence(channel, presenceData)
                         callback(Result.success(Unit))
@@ -800,30 +799,30 @@ constructor(
     }
 
     /**
-     * Closes [AblySdkRealtime] and waits until it's either closed or failed.
+     * Closes [ably] and waits until it's either closed or failed.
      * If the connection is already closed it returns immediately.
      * If the connection is already failed it returns immediately as closing a failed connection should be a no-op
      * according to the Ably features spec (https://sdk.ably.com/builds/ably/specification/main/features/#state-conditions-and-operations).
      *
      * @throws ConnectionException if the [AblySdkRealtime] state changes to [ConnectionState.failed].
      */
-    private suspend fun AblySdkRealtime.closeSuspending() {
-        if (connection.state.isClosed() || connection.state.isFailed()) {
+    private suspend fun closeSuspending() {
+        if (ably.connection.state.isClosed() || ably.connection.state.isFailed()) {
             return
         }
         suspendCancellableCoroutine<Unit> { continuation ->
-            connection.on(object : ConnectionStateListener {
+            ably.connection.on(object : ConnectionStateListener {
                 override fun onConnectionStateChanged(connectionStateChange: ConnectionStateListener.ConnectionStateChange) {
                     if (connectionStateChange.current.isClosed()) {
-                        connection.off(this)
+                        ably.connection.off(this)
                         continuation.resume(Unit)
                     } else if (connectionStateChange.current.isFailed()) {
-                        connection.off(this)
+                        ably.connection.off(this)
                         continuation.resumeWithException(connectionStateChange.reason.toTrackingException())
                     }
                 }
             })
-            close()
+            ably.close()
         }
     }
 
@@ -906,13 +905,13 @@ constructor(
         errorInformation.let { it.message == "Connection resume failed" && it.code == 50000 && it.statusCode == 500 }
 
     /**
-     * Enter the presence of the [Channel] and waits for this operation to complete.
+     * Enter the presence of [channel] and waits for this operation to complete.
      * If something goes wrong then it throws a [ConnectionException].
      */
-    private suspend fun AblySdkRealtime.Channel.enterPresenceSuspending(presenceData: PresenceData) {
+    private suspend fun enterPresenceSuspending(channel: AblySdkRealtime.Channel, presenceData: PresenceData) {
         suspendCancellableCoroutine<Unit> { continuation ->
             try {
-                presence.enter(
+                channel.presence.enter(
                     gson.toJson(presenceData.toMessage()),
                     object : CompletionListener {
                         override fun onSuccess() {
@@ -921,42 +920,42 @@ constructor(
 
                         override fun onError(reason: ErrorInfo?) {
                             val trackingException = reason?.toTrackingException()
-                                ?: ConnectionException(ErrorInformation("Unknown error when entering presence $name"))
-                            logHandler?.w("$TAG Failed to suspend enter presence for channel $name", trackingException)
+                                ?: ConnectionException(ErrorInformation("Unknown error when entering presence $channel.name"))
+                            logHandler?.w("$TAG Failed to suspend enter presence for channel $channel.name", trackingException)
                             continuation.resumeWithException(trackingException)
                         }
                     }
                 )
             } catch (ablyException: AblyException) {
                 val trackingException = ablyException.errorInfo.toTrackingException()
-                logHandler?.w("$TAG Failed to suspend enter presence for channel $name", trackingException)
+                logHandler?.w("$TAG Failed to suspend enter presence for channel $channel.name", trackingException)
                 continuation.resumeWithException(trackingException)
             }
         }
     }
 
     /**
-     * Attaches the [AblySdkRealtime.Channel] and waits for this operation to complete.
+     * Attaches [channel] and waits for this operation to complete.
      * If something goes wrong then it throws a [ConnectionException].
      */
-    private suspend fun AblySdkRealtime.Channel.attachSuspending() {
+    private suspend fun attachSuspending(channel: AblySdkRealtime.Channel) {
         suspendCancellableCoroutine<Unit> { continuation ->
             try {
-                attach(object : CompletionListener {
+                channel.attach(object : CompletionListener {
                     override fun onSuccess() {
                         continuation.resume(Unit)
                     }
 
                     override fun onError(reason: ErrorInfo?) {
                         val trackingException = reason?.toTrackingException()
-                            ?: ConnectionException(ErrorInformation("Unknown error when attaching channel $name"))
-                        logHandler?.w("$TAG Failed to suspend attach to channel $name", trackingException)
+                            ?: ConnectionException(ErrorInformation("Unknown error when attaching channel $channel.name"))
+                        logHandler?.w("$TAG Failed to suspend attach to channel $channel.name", trackingException)
                         continuation.resumeWithException(trackingException)
                     }
                 })
             } catch (ablyException: AblyException) {
                 val trackingException = ablyException.errorInfo.toTrackingException()
-                logHandler?.w("$TAG Failed to suspend attach to channel $name", trackingException)
+                logHandler?.w("$TAG Failed to suspend attach to channel $channel.name", trackingException)
                 continuation.resumeWithException(trackingException)
             }
         }
@@ -969,7 +968,7 @@ constructor(
 
     override suspend fun startConnection(): Result<Unit> {
         return try {
-            ably.connectSuspending()
+            connectSuspending()
             Result.success(Unit)
         } catch (connectionException: ConnectionException) {
             logHandler?.w("$TAG Failed to start Ably connection", connectionException)
@@ -979,7 +978,7 @@ constructor(
 
     override suspend fun stopConnection(): Result<Unit> {
         return try {
-            ably.closeSuspending()
+            closeSuspending()
             Result.success(Unit)
         } catch (connectionException: ConnectionException) {
             logHandler?.w("$TAG Failed to stop Ably connection", connectionException)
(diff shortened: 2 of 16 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
```diff
@@ -119,13 +118,12 @@ interface Publisher {
      * Stops this publisher from publishing locations. Once a publisher has been stopped, it cannot be restarted.
      * Please note that calling this method will remove the notification provided by [Builder.backgroundTrackingNotificationProvider].
      *
-     * @param timeoutInMilliseconds After this duration the stopping procedure will be canceled. Default value is 30 seconds.
+     * @param presenceLeaveTimeoutInMilliseconds After this duration, the attempt to leave the Ably channel presence will be cancelled. If this timeout elapses, the remainder of the stopping procedure will still be performed, and subscribers will see the publisher go offline within 75 seconds. Default value is 30 seconds.
```
KacperKluka, line 121: The same thing about those `75 seconds` as before 😉

KacperKluka (commented):
Overall looks good, just one thing to clarify 😉

davyskiba (approved):

UPDATING.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade / Migration Guide
 
+## Version 1.5.1 to 1.6.0
+
+### Changed timeout behaviour of `Publisher.stop`
+
+It is no longer possible to specify a timeout for the entire `Publisher.stop` operation. The `timeoutInMilliseconds` parameter has been replaced by a `presenceLeaveTimeoutInMilliseconds` parameter, which allows you to specify a timeout for part of this operation.
+
+The `Publisher.stop` method now no longer throws `TimeoutCancellationException`, and guarantees that subscribers will see the publisher go offline within 75 seconds of this method returning.
```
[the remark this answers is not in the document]
lawrence-forooghian: Good point! I've updated the documentation, and also updated it to make it consistent with the other (to talk about how long it may take for subscribers to see the publisher go offline starting from the moment the method is called instead of the moment it returns).

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -224,10 +224,15 @@ interface Ably {
...
...t will still call [AblySdkRealtime.close] on the underlying Realtime object. As described by [RTN12](https://sdk.ably.com/builds/ably/specification/main/features/#RTN12), this means that whatever state the Realtime object is currently in, it will no longer attempt to connect to Ably. If the call to [AblySdkRealtime.close] is successful, then the client will have immediately left presence (see [the presence documentation](https://ably.com/docs/realtime/presence?lang=javascript#unstable-connections) which says "If a client calls `leave()` or `close()` they immediately send a `leave` event."). If there are network issues that prevent the call to [AblySdkRealtime.close] from succeeding, then the server will after around 60 seconds consider the client to have disconnected and 15 seconds after that will synthesize a presence leave event (see mention of “abrupt disconnect“ on the aforementioned documentation). The data of this presence leave event will be the same as the last presence data published for that member – this behaviour is not officially documented but is confirmed in [this GitHub comment](https://github.com/ably/ably-asset-tracking-android/issues/862#issuecomment-1366958096).
```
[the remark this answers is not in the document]
lawrence-forooghian: Updated, thanks!

publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
lawrence-forooghian: Good point! I've updated the documentation, and also updated it to make it consistent with the rest (to talk about how long it may take for subscribers to see the publisher go offline starting from the moment the method is called instead of the moment it returns).

KacperKluka (approved):
LGTM 👍

lawrence-forooghian:
@KacperKluka @davyskiba did you have any thoughts about what kind of version bump this necessitates (perhaps Kacper you have an idea of how we've handled this sort of thing in the past)? (See "API considerations" in the description.)

KacperKluka:
@lawrence-forooghian IMO this particular API situation is debatable as you only change a parameter name, so this only affects users that use the named parameters in functions. Due to this I'd be fine with either bumping the minor number or bumping the major one 😉

paddybyers (approved):
I've look at your reasoning in the comments and I'm happy to approve that approach. I haven't reviewed the code itself or the tests.

QuintinWillison:
@lawrence-forooghian @KacperKluka I've had a look at the proposed change to the `stop` method (being the rename of the `Long` argument from `timeoutInMilliseconds` to `presenceLeaveTimeoutInMilliseconds`).

I don't believe it's correct to summarise the impact of this change as "only affects users that use the named parameters in functions": If a user doesn't use named parameters when calling this method but does pass a value to override this argument's default value then that value will now be interpreted in a different way, resulting in a change in SDK behaviour, regardless of the name.

Therefore, I think the change as currently presented is a breaking / incompatible change (`major`) whichever way we look at it.

A couple of alternative routes spring to mind that might make this easier to explain to customers when we do a release (migration guide, etc..):

1. Change the argument type entirely, perhaps something like a `StopOptions` data class that has a `presenceLeaveTimeoutInMilliseconds` member.
2. Use deprecation and allow the timeout value to be optionally injected elsewhere:
    - For the current `stop(Long = 30_000L)` method, leave the argument in place, remove the default value and deprecate it -- it becomes `stop(Long)` (deprecated)
    - add a new overload for `stop()` that takes no arguments
    - add a builder method to allow users to inject this new timeout value before they even start the publisher -- I'm pretty sure they'll know this up front, right?

Option 1 is definitely a breaking change (`major`), but option 2 could be a new feature / enhancement (`minor`).
My preference would be option 2.

jaley:
@lawrence-forooghian (or maybe @paddybyers or @lmars) - does the user really *need* control over this presence leave timeout? It looks a bit odd to me, as Ably presence functionality isn't really exposed to the user elsewhere in the API, it's more of implementation strategy for delivering the required experience, AFAICT? If there's no strong reason to expose the presence leave timeout, would it not be better to just do part of @QuintinWillison's suggestion (2) above:

* Add a new `stop()` overload that takes no parameters
* Deprecate `stop(Long)` and just document that the timeout parameter is now ignored and we won't throw that exception.

This isn't source-breaking, and the only behaviour change is that we'll no-longer throw an exception, which has already been agreed as an improvement from the user perspective?

paddybyers:
>  does the user really need control over this presence leave timeout?

I don't think so.

> would it not be better to just do part of @QuintinWillison's suggestion (2) above:

- Add a new stop() overload that takes no parameters
- Deprecate stop(Long) and just document that the timeout parameter is now ignored and we won't throw that exception.

I'm ok with that

lawrence-forooghian:
Before I go ahead with deprecating the ability to control the duration of the publisher `stop()` method call, I'd like to be sure that we're not removing something the client is depending on. @KacperKluka do you remember what the original motivation was for adding that timeout parameter? Was it based on some user need?

KacperKluka:
> do you remember what the original motivation was for adding that timeout parameter

I don't remember this exactly but I think we simply wanted to avoid freezing the SDK as the `stop` method is going to block the internal worker queue in order to correctly perform the stop operation 🤔 As far as I remember it wasn't a request from a customer but rather our idea 😉

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/workerqueue/workers/StopConnectionWorker.kt
```diff
@@ -22,7 +22,7 @@ internal class StopConnectionWorker(
         // We're using [runBlocking] on purpose as we want to block the whole subscriber when it's stopping.
         runBlocking {
             try {
-                ably.close(properties.presenceData)
+                ably.close(properties.presenceData, presenceLeaveTimeoutInMilliseconds = null)
```
davyskiba, line 25: What is the reasoning behind treating subscriber stop differently?

jaley:
> deprecating the ability to control the duration of the publisher stop() method call, I'd like to be sure that we're not removing something the client is depending on.

This is a fair point. Another option then would actually be to leave the function signature as it is, honour the timeout but not throw an exception if it expires, simply give up and stop blocking the caller. Was that option considered? (Sorry, I may have missed earlier discussion)

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/workerqueue/workers/StopConnectionWorker.kt
lawrence-forooghian: I don't know – this operation previously had no timeout and so that's the behaviour I’ve preserved.

lawrence-forooghian:
If we believe that having any sort of timeout on `stop` is a mistake, should we just remove the argument from that method, increase the major version and call it a day? Any other kind of change (ignoring the parameter, not throwing an exception etc) is all, technically, a breaking change to the API.

lawrence-forooghian:
As agreed in our call this morning, I’ve removed the timeout logic and removed the corresponding parameter on `stop()`. I've written a bit about this decision in the commit message.

I’ve updated the `UPDATING.md` to explain how users can replace the timeout by using Kotlin’s `withTimeout`.

@KacperKluka @davyskiba I've added a couple of tests to give me confidence that if a user calls `stop()` from within a `withTimeout` block then the `TimeoutCancellationException` does indeed reach the user. I think that they probably come from my poor knowledge of Kotlin coroutines, so if you think them unnecessary please let me know and I'll get rid of them.

publishing-sdk/src/test/java/com/ably/tracking/publisher/DefaultPublisherTest.kt
```diff
@@ -221,4 +225,35 @@ class DefaultPublisherTest {
...
+        }
+
+        // Then...
+        coVerify(exactly = 1) {
+            // ...ably.close() is called...
+            ably.close(any())
+        }
+
+        // ...and the withTimeout block throws a TimeoutCancellationException.
+        Assert.assertTrue(caughtTimeoutCancellationException)
```
davyskiba, line 255: Please use [Expectations](https://github.com/ably/ably-asset-tracking-android/blob/main/android-test-common/src/main/java/com/ably/tracking/test/android/common/Expectations.kt) instead of assertions here

lawrence-forooghian: Happy to change it, but would you mind explaining why? It seems like the main functionality of the `Expectation` class is to provide the ability to wait for some asynchronous operation to complete. But we're already waiting for the coroutine call to complete by using `runBlocking`.

davyskiba: Right await is not needed here, I would use it here mostly for a more verbose messages when the test fails

lawrence-forooghian: Does that mean we can stick to using assert, or is there some other reason to switch to using expectations?

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/StopWorker.kt
```diff
@@ -26,19 +24,16 @@ internal class StopWorker(
...
+                if (properties.isTracking) {
+                    publisherInteractor.stopLocationUpdates(properties)
                 }
+                publisherInteractor.closeMapbox()
+                ably.close(properties.presenceData)
+                properties.dispose()
+                callbackFunction(Result.success(Unit))
             } catch (exception: ConnectionException) {
                 callbackFunction(Result.failure(exception))
             } catch (exception: TimeoutCancellationException) {
```
KacperKluka, line 39: Since the `withTimeout` block is removed I think we no longer need to catch the `TimeoutCancellationException` 🤔

KacperKluka (changes requested):
Looks good, just a few small things to improve 😉

publishing-sdk/src/main/java/com/ably/tracking/publisher/DefaultPublisher.kt
```diff
@@ -112,6 +111,14 @@ constructor(
         }
     }
 
+    @Deprecated(
+        "The timeoutInMilliseconds parameter is now ignored and should not be used.",
+        replaceWith = ReplaceWith("stop()")
+    )
+    override suspend fun stop(timeoutInMilliseconds: Long) {
+        stop()
+    }
```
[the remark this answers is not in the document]
lawrence-forooghian: I added it because Android Studio was giving me this warning:

image

Should I ignore the warning, or keep it as it currently is, or suppress the warning in some other way?

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/StopWorker.kt
lawrence-forooghian: Good point, removed.

publishing-sdk/src/main/java/com/ably/tracking/publisher/DefaultPublisher.kt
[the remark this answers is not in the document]
lawrence-forooghian: (I just tried removing it, and it actually seems to be a build error, not just a warning - try running `./gradlew check` to reproduce.)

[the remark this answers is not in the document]
KacperKluka: Oh sorry, this must be my mistake, I was recently deprecating some things in AAT but they were regular methods, not an interface-implementation pairs 😅 It looks like in this situation the `@Deprecated` annotation needs to be added in both places 👍 Sorry for the trouble 😉

KacperKluka (approved):
Good job! LGTM 👍

publishing-sdk/src/test/java/com/ably/tracking/publisher/DefaultPublisherTest.kt
davyskiba: Yes, we can :)
I prefer to use something with a more informative error message here, like Expectation or Truth fluent assertion, but it is not worth keeping this pr open longer just for that

PR #877, opened 2023-01-04
Add dedicated unexpected error handling to certain workers

KacperKluka:
I checked all publisher and subscriber workers and tried to find paths in which an unexpected exception might break the SDK state (e.g. make it stay in the `CONNECTING` state indefinitely). Then I applied the simplest fixes to those paths that make sure the SDK is working properly. 

This is the next iteration of the unexpected exceptions handling that added a generic, default handling for all workers https://github.com/ably/ably-asset-tracking-android/pull/872.


ikbalkaya (approved):
I haven't checked the impact of these changes deeply but they all seems sensible improvements to me.

davyskiba (approved):

Issue #869, opened 2022-12-28
Write unit tests for `DefaultAbly`


lawrence-forooghian:
Placeholder; will fill in.

lawrence-forooghian:
After #914 is merged, the remaining work for this issue is the following:

- black-box tests for:
    - `subscribeForRawEvents`
    - `subscribeForEnhancedEvents`
    - `sendEnhancedLocation`
    - `getCurrentPresence`
    - `subscribeForPresenceMessages`
    - `subscribeForChannelStateChange`
    - `subscribeForAblyStateChange`
    - `connect`’s other parameters (`useRewind`, `willPublish`, `willSubscribe`)
    - `close` (we already have some tests for this method, but want a full set of tests consistent with the other tests added for this issue)
- augment the black-box tests with any further information found by looking at the code of `DefaultAbly` (off the top of my head, I think I remember some auth-related retry behaviour)

jaley:
Removed `important` tag as this doesn't need to block the next release

PR #878, opened 2023-01-04
Write black-box tests for `DefaultAbly.connect`

lawrence-forooghian:
These document the current behaviour of this method.

I wrote them having only seen the `Ably` interface (which `DefaultAbly` implements), and knowing the dependencies that `DefaultAbly` has.

I intentionally did not look at the code whilst writing these tests because I wanted to limit myself to writing test cases based on scenarios that I could imagine might occur, instead of ones based on logic that I know to exist.

Once I’ve written tests like this for the whole class, then I’d like to look at its implementation and add any further test cases for logic that wasn’t clear from the public interface.

This is part of #869.

### DIFF
--- a/common/src/test/java/com/ably/tracking/common/DefaultAblyTests.kt
+++ b/common/src/test/java/com/ably/tracking/common/DefaultAblyTests.kt
@@ -1,7 +1,8 @@
 package com.ably.tracking.common
 
+import com.ably.tracking.ErrorInformation
 import com.ably.tracking.common.helper.DefaultAblyTestEnvironment
-import io.mockk.verify
+import com.ably.tracking.common.helper.DefaultAblyTestScenarios
 import io.mockk.verifyOrder
 import io.ably.lib.realtime.ChannelState
 import io.ably.lib.realtime.ConnectionState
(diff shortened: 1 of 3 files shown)


common/src/test/java/com/ably/tracking/common/DefaultAblyTests.kt
```diff
@@ -12,24 +14,402 @@ import org.junit.Test
 import kotlin.system.measureTimeMillis
 
 class DefaultAblyTests {
-    // This is just an example test to check that the AblySdkRealtime mocks are working correctly. We need to add a full set of unit tests for DefaultAbly; see https://github.com/ably/ably-asset-tracking-android/issues/869
+    /*
+    Observations from writing black-box tests for `connect`:
+
+    - The following statement from the docstring "If a channel for the given [trackableId] exists then it just calls [callback] with success." doesn’t seem to match what I’m seeing in the tests. If for a channel to "exist" it means to say that the Channels.containsKey or Channels.entrySet methods already indicate an entry for that channel, the tests show no method calls that would suggest this check is being made.
```
KacperKluka, line 20: > If for a channel to "exist" it means to say that the Channels.containsKey or Channels.entrySet methods already indicate an entry for that channel, the tests show no method calls that would suggest this check is being made.

I'm not sure if that statement is true, because in all tests the first thing you verify is a call to the [containsKey()](https://github.com/ably/ably-asset-tracking-android/pull/878/files?diff=unified&w=0#diff-f043da3aa98e2e50bdee73be5bfe3c49b361200160b4c8767b4b65a965210b94R171) 🤔 So the SDK does check the channel existence before doing anything else.

common/src/test/java/com/ably/tracking/common/DefaultAblyTests.kt
```diff
@@ -12,24 +14,402 @@ import org.junit.Test
...
+
+    - When given a channel in certain states, it seems to fetch the channel’s state more than once. I have not tested what happens if a different state is returned on the second call.
```
KacperKluka, line 22: That's a good catch 👍 The reason behind this is that we check whether the channel is in one of two states:
```
channel.state == ChannelState.detached || channel.state == ChannelState.failed
```
So if it is in the `detached` state then the second `state` won't be executed. Otherwise, there will always be 2 calls to the `state` getter. Do you think that this is an issue that we fetch the state twice instead of storing the initial value in some variable and using the variable for comparison?

Not sure if you wanted to hear the answer as you mentioned that you will be looking into the implementation later 😅

KacperKluka (commented):
I have to admit that I had mixed feelings about the `DefaultAblyTestScenarios` object, but after familiarizing myself with it the reasoning about particular test cases is much easier with it 👍 Though, it still might be difficult at first to understand what's going on in those tests 😉 

Overall the tests look good, just a few comments from my side

lawrence-forooghian:
> I have to admit that I had mixed feelings about the `DefaultAblyTestScenarios` object, but after familiarizing myself with it the reasoning about particular test cases is much easier with it 👍 Though, it still might be difficult at first to understand what's going on in those tests 😉

I agree with you 100%. I find both reading and writing tests exceptionally difficult _in general_! And I'm constantly flipping back and forth between "the test cases should be very explicit" (which then leads to super long test cases which completely obscure what you're actually testing with a load of boilerplate - see the ably-cocoa tests) and "the test cases should be concise" (where you end up with these test scenario methods which have a load of code and require you to read them).

I've aimed to ease some of the issues with the latter approach here by writing lots of prose - the test scenario method has a human-readable Given / When / Then description of exactly what it's testing, as does every single test case. So people looking at the `DefaultAblyTests` file don't need to understand the scenario code if they don't want to.

common/src/test/java/com/ably/tracking/common/DefaultAblyTests.kt
lawrence-forooghian: Ah, that's interesting. Maybe we could simplify that. But that's for later 😄

common/src/test/java/com/ably/tracking/common/DefaultAblyTests.kt
lawrence-forooghian: Oh my, how embarrassing! Yep, you're right.

I think that what I probably actually should be saying in this note is that even if `containsKey` returns `true`, it doesn't just call the callback with success, it still enters presence etc.

I should probably add a test cases for the case where `containsKey` returns `false`, though. I think it would only be the `INITIALIZED` state where this would be necessary, right?

KacperKluka: > I think that what I probably actually should be saying in this note is that even if containsKey returns true, it doesn't just call the callback with success, it still enters presence etc.

What makes you think that? 🤔 AFAIK you are [mocking](https://github.com/ably/ably-asset-tracking-android/blob/3ea18d8b50e8f42cb139d929f88b8eaaf43522e7/common/src/test/java/com/ably/tracking/common/helper/DefaultAblyTestEnvironment.kt#L332) the `containsKey` to always return `false` and you never override this in your other tests 😅 So the tested behaviour is correct, when there is no trackable we start connecting it 😉 

If you want you can add a test case when the `containsKey` returns `true` and check whether in this situation there is no other method being called 👍

lawrence-forooghian: Haha, really not having a good brain day today. You're right 👍 – will change accordingly.

lawrence-forooghian: OK, I've added a test case for when `containsKey` returns `true` (and updated the tests to reflect the fact that a different overload of `Channels.get` is used in this case).

KacperKluka (approved):
LGTM 👍

common/src/test/java/com/ably/tracking/common/helper/DefaultAblyTestEnvironment.kt
```diff
@@ -28,7 +32,7 @@ class DefaultAblyTestEnvironment(
     val objectUnderTest: DefaultAbly,
     /**
      * The [AblySdkRealtime] mock created by [create].
-     * Its [AblySdkRealtime.connection] property returns [connectionMock], and its [AblySdkRealtime.channels] property returns [channelsMock].
+     * Its [AblySdkRealtime.connection] property returns [connectionMock], and its [AblySdkRealtime.channels] property returns [channelsMock]. Calls to these property getters are ignored by MockK’s [confirmVerified] method (by use of [excludeRecords]).
```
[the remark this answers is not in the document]
lawrence-forooghian: No, there shouldn’t be any reason to directly invoke the constructor. Good idea, I’ve made this change in #874 because that's where this constructor was introduced.

common/src/test/java/com/ably/tracking/common/helper/DefaultAblyTestScenarios.kt
```diff
@@ -0,0 +1,248 @@
...
+            }
+        }
+
+        /**
+         * Provides test scenarios for [DefaultAbly.connect].
+         */
+        companion object {
+            /**
+             * Implements the following parameterised test case for [DefaultAbly.connect]:
+             *
```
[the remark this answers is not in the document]
lawrence-forooghian: It's intentionally that way, for two reasons:

1. to explicitly document what the test case is doing without needing to read the code
2. to make it easy to write prose documentation for each of the test methods in `DefaultAblyTests`, by taking this docstring and deleting / replacing placeholders where appropriate.

common/src/test/java/com/ably/tracking/common/helper/DefaultAblyTestScenarios.kt
```diff
@@ -0,0 +1,248 @@
...
+                     * ...[and] which, when told to attach, fails to do so with error ${givenConfig.channelAttachBehaviour.errorInfo}...
+                     * }
+                     */
+                    is CompletionListenerMockBehaviour.Failure -> {
+                        configuredChannel.mockFailedAttach(channelAttachBehaviour.errorInfo)
+                    }
+                }
+
+                // When...
+                val result = runBlocking {
```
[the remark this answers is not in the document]
lawrence-forooghian: 👍 I’ve made that change

davyskiba (approved):

lawrence-forooghian:
I'm going to convert this to a draft since I think I might make some changes to it as part of my continued work on #878.

lawrence-forooghian:
@davyskiba @KacperKluka I've [pushed](https://github.com/ably/ably-asset-tracking-android/compare/1a2b413715821c051dbdeea832b97d70d9b50284..cc92d617061ab438422dba79abb1baca48c32751) some changes (just the ones to `DefaultAblyTestScenarios.kt`; the other changes are from `main`). It generalises some of the stuff that this PR adds, so that I can use it in subsequent tests. There are no changes in terms of content.

lawrence-forooghian:
Also [pushed](https://github.com/ably/ably-asset-tracking-android/compare/cc92d617061ab438422dba79abb1baca48c32751..95124aed36cff23e55deaa498a276816e3c9f9a7) a minor change, simplifying syntax used in docstring comments for `if` / `when`.

PR #886, opened 2023-01-08
Layer 7 Proxy for Network Connectivity Tests

jaley:
This adds a Layer 7 Proxy based on Ktor. A few things to note:

* Added client and websockets server ktor dependencies as test dependencies in `android.test-common`
* Kotlin Coroutines dependency bump required for ktor to work (1.4.2 to 1.6.2) -- is this safe?
* Pass-through interceptor implementation provided, which does nothing, and null Layer 7 fault, shows normal functionality working fine with proxy just passing all frames through.

This should be ready for others to build on now, implementing more of the planned fault types.


integration-testing-app/build.gradle
```diff
@@ -11,8 +11,10 @@ android {
 
     buildTypes {
         debug {
-            minifyEnabled true
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            minifyEnabled false
```
KacperKluka, line 14: Why is the Proguard tool disabled now? 🤔 We enabled it in this module on purpose because we wanted to make sure that it doesn't break the SDK when customers use it in their apps with Proguard enabled. We even have `ObfuscationTest` in this module.

KacperKluka (changes requested):
Looks good but I have one important question 😉

integration-testing-app/build.gradle
QuintinWillison: Thanks for picking this up and you're correct, though I had forgotten the details! 😁 We have to balance risks at the moment. I disabled it because of the complexities introduced by all the new dependencies required to test with. I was going around in circles trying to get them all to play together and prevent class loader issues when running. I shall create an issue in the backlog to re-enable this in future, but right now I think obfuscation concerns have to go on the back burner. 🤷

KacperKluka (approved):

integration-testing-app/build.gradle
QuintinWillison: https://github.com/ably/ably-asset-tracking-android/issues/926

QuintinWillison:
I've triggered fresh CI job runs by [merging in `main` branch](https://github.com/ably/ably-asset-tracking-android/pull/886/commits/d9e7b64565ae481b21244f01ad082448b0c8f4a6), as it seems (just) one of the previous emulation runs had hung:

Screenshot 2023-01-19 at 10 54 25

https://github.com/ably/ably-asset-tracking-android/actions/runs/3957205010/jobs/6777284020

PR #887, opened 2023-01-09
coroutines and coroutines-test dependency updated to 1.6.4

davyskiba:
Updating test coroutines dependency to use `runTest` for better handling of async calls in the test code, as we are currently adding lots of them.
`runTest` can be used to provide async context for the whole tests instead of local `runBlocking` blocks.
The biggest gain from using `runTest` is virtual time manipulation - a test with 10 seconds delay won't be executing 10 seconds

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 buildscript {
     ext.kotlin_version = '1.6.21'
+    ext.kotlin_coroutines_version = '1.6.4'
     ext.copyright_message = (new File('COPYRIGHT')).readLines().get(0)
 
     // The version of Ably's client library for Java that we're using.
@@ -242,11 +243,12 @@ subprojects {
                 // Projects that use our SDK projects will crash when dispatching to the Main thread on
                 // the Android runtime unless we include this dependency. See:
                 // https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-main.html
-                implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.2'
-                testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.4.2"
+                implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlin_coroutines_version"
+                testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlin_coroutines_version"
                 testImplementation 'io.mockk:mockk:1.12.2'
             } else {
                 testImplementation 'org.mockito:mockito-core:3.6.28'
+                testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test-jvm:$kotlin_coroutines_version"
             }
 
             if (evaluatedSubProjectIsAnApp) {
--- a/subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
+++ b/subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
@@ -39,6 +39,9 @@ public class UsageExamples {
     Subscriber.Builder nativeSubscriberBuilder;
     SubscriberFacade.Builder subscriberFacadeBuilder;
 
+    // We need to inject a different coroutine dispatcher in place of the Main dispatcher,
+    // because DefaultSubscriberFacade is directly accessing it.
+    // Otherwise, tests fail with RuntimeException: Stub!
     @Rule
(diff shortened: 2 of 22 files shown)


ikbalkaya (approved):

build.gradle
```diff
@@ -242,11 +243,12 @@ subprojects {
...
                 // the Android runtime unless we include this dependency. See:
                 // https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-main.html
-                implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.2'
-                testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.4.2"
+                implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlin_coroutines_version"
+                testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlin_coroutines_version"
                 testImplementation 'io.mockk:mockk:1.12.2'
             } else {
                 testImplementation 'org.mockito:mockito-core:3.6.28'
+                testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test-jvm:$kotlin_coroutines_version"
```
KacperKluka, line 251: Do we need this in the Java tests? 🤔 I think that in the Java tests we should use the Java Facade API that doesn't use anything from the coroutines package

test-common/build.gradle
```diff
@@ -8,6 +8,8 @@ dependencies {
     implementation project(':core-sdk')
     implementation project(':common')
     implementation 'io.mockk:mockk:1.10.2'
+    implementation 'junit:junit:4.13.2'
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlin_coroutines_version"
```
KacperKluka, line 12: Just to be sure, this is added here because the common test code is placed in the "production" folders from this module, right? 🤔 So the `testImplementation` with those dependencies in the root `build.gradle` file are not enough, right?

subscribing-java-testing/src/test/java/com/ably/tracking/example/javasubscriber/UsageExamples.java
```diff
@@ -16,24 +16,32 @@
...
+
+@ExperimentalCoroutinesApi
 public class UsageExamples {
     SubscriberFacade subscriberFacade;
     Subscriber nativeSubscriber;
     Subscriber.Builder nativeSubscriberBuilder;
     SubscriberFacade.Builder subscriberFacadeBuilder;
 
+    @Rule
+    public ReplaceMainCoroutineDispatcherRule replaceMainCoroutineDispatcherRule = new ReplaceMainCoroutineDispatcherRule();
```
KacperKluka, line 43: Forgive me for my ignorance, but why do we need it here? 😅

test-common/src/main/java/com/ably/tracking/test/common/ReplaceMainCoroutineDispatcherRule.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.TestDispatcher
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+@ExperimentalCoroutinesApi
+class ReplaceMainCoroutineDispatcherRule(
```
KacperKluka, line 13: Perhaps we could add some docs here that would explain what's the purpose of this class and how and when we should use it? 🤔

KacperKluka (commented):
Looks good, a few questions from my side 😉

build.gradle
KacperKluka: As discussed with @davyskiba in a call, this is actually needed to make Java tests work, so I'm resolving this comment 👍

test-common/build.gradle
KacperKluka: As above, this was confirmed by @davyskiba on a call, therefore, I'm resolving this conversation 😉

test-common/src/main/java/com/ably/tracking/test/common/ReplaceMainCoroutineDispatcherRule.kt
davyskiba: 3426bf4bb0cfdcb9df171d18a8409ba664ae5fa6

KacperKluka (approved):
LGTM 👍

PR #890, opened 2023-01-10
Add timeout on verification in WorkerQueueTest

### LINKED ISSUE #888: Flakey test: `when an unexpected exception is thrown by worker's async work, the queue should call worker's unexpected async exception method`
```
com.ably.tracking.common.workerqueue.WorkerQueueTest > when an unexpected exception is thrown by worker's async work, the queue should call worker's unexpected async exception method FAILED
    java.lang.AssertionError at WorkerQueueTest.kt:101
```

Fails in https://github.com/ably/ably-asset-tracking-android/actions/runs/3882562782/jobs/6622854474#step:4:420

Possibly caused by the async code not running before the verification step is run.

AndyTWF:
The delays already present in the test weren't quite long enough to allow for the asynchronous coroutines to run. This change adds a slightly longer timeout to the verification step, to ensure that the WorkerQueue has time to run the jobs. This should hopefully fix the flakey test issue.

Fixes #888

### DIFF
--- a/common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
+++ b/common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
@@ -38,10 +38,9 @@ class WorkerQueueTest {
 
         // when
         workerQueue.enqueue(Unit)
-        waitForWorkerToBeProcessed()
 
         // then
-        verify(exactly = 1) { worker.doWork(any(), any(), any()) }
+        verify(exactly = 1, timeout = 5000) { worker.doWork(any(), any(), any()) }
     }
 
     @Test
@@ -51,10 +50,9 @@ class WorkerQueueTest {
 
         // when
         workerQueue.enqueue(Unit)
-        waitForWorkerToBeProcessed()
 
         // then
-        verify(exactly = 1) { worker.doWhenStopped(any()) }
+        verify(exactly = 1, timeout = 5000) { worker.doWhenStopped(any()) }
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


davyskiba (approved):

QuintinWillison (approved):
Feel free to merge this as is, though I might suggest it might be useful from a maintainability perspective to pull the `5000` (presumably millisecond) value out to a constant - something like `DO_WORK_TIMEOUT_MILLISECONDS`.

common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
```diff
@@ -96,10 +90,9 @@ class WorkerQueueTest {
 
         // when
         workerQueue.enqueue(Unit)
-        waitForWorkerToBeProcessed()
 
         // then
-        verify(exactly = 1) { worker.onUnexpectedError(any(), any()) }
+        verify(exactly = 1, timeout = 5000) { worker.onUnexpectedError(any(), any()) }
```
KacperKluka, line 95: This timeout value is repeated in lots of places, perhaps it would be worth extracting it to a constant? 🤔 Also it would be good to know what's the unit of the timeout (I assume milliseconds), so we could name the constant something like `ASYNC_WORK_TIMEOUT_IN_MILLISECONDS` 😉

KacperKluka (changes requested):
Looks good, just one suggestion 😉

common/src/test/java/com/ably/tracking/common/workerqueue/WorkerQueueTest.kt
QuintinWillison: I prefer @KacperKluka's constant name. 😆

AndyTWF: All done @KacperKluka :)

KacperKluka (approved):
LGTM, thanks! 👍

Issue #259, opened 2021-03-26
Flakey Test: `createAndStartPublisherAndSubscriberAndWaitUntilDataEnds` (`PublisherAndSubscriberTests`)


QuintinWillison:
Observed when running `emulate` workflow on #243.

```
com.ably.tracking.tests.PublisherAndSubscriberTests > createAndStartPublisherAndSubscriberAndWaitUntilDataEnds[test(AVD) - 7.0] FAILED 
	java.lang.AssertionError: Expectation 'track response' unfulfilled.
	at com.ably.tracking.test.common.Expectation.assertFulfilled(Expectations.kt:55)
```

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-789) by [Unito](https://www.unito.io)

KacperKluka:
This should be fixed after landing https://github.com/ably/ably-asset-tracking-android/pull/428

QuintinWillison:
[At API Level 21](https://github.com/ably/ably-asset-tracking-android/actions/runs/3716720723/jobs/6303356761#step:4:779) (https://github.com/ably/ably-asset-tracking-android/pull/855):

```
com.ably.tracking.tests.PublisherAndSubscriberTests > createAndStartPublisherAndSubscriberAndWaitUntilDataEnds[test(AVD) - 5.0.2] FAILED 
	java.lang.AssertionError: Received Subscriber location should be equal to published Publisher location (index 3) expected:<com.ably.tracking.EnhancedLocationUpdate@bc797d73> but was:<com.ably.tracking.EnhancedLocationUpdate@be4f8529>
	at org.junit.Assert.fail(Assert.java:89)
```

QuintinWillison:
[At API Level 29](https://github.com/ably/ably-asset-tracking-android/actions/runs/3730953632/jobs/6329036149#step:4:919) (https://github.com/ably/ably-asset-tracking-android/pull/855):

```
com.ably.tracking.tests.PublisherAndSubscriberTests > createAndStartPublisherAndSubscriberAndWaitUntilDataEnds[test(AVD) - 10] FAILED 
	java.lang.AssertionError: Received Subscriber location should be equal to published Publisher location (index 3) expected:<com.ably.tracking.EnhancedLocationUpdate@b407d20d> but was:<com.ably.tracking.EnhancedLocationUpdate@b63766de>
	at org.junit.Assert.fail(Assert.java:89)
```

[...]

PR #892, opened 2023-01-11
Use different locations in test data file

AndyTWF:
Prior to this change, the testing data file contained a series of the same locations (excluding timestamp). On receiving a location update, we stamp the current millisecond-precision timestamp on it.

If during tests, mapbox enhanced updates come through very quickly (<1ms), they are registered on the publisher, but may not be sent to Ably.

This may be due to resolution, or because such a quick update would make the timestamp on subsequent updates the same. As everything else is the same, this would produce the same hashcode for the location update, which is also what we use as the message id for the Ably realtime service. So the message would be rejected for reasons of idempotency.

This change therefore makes sure that each data point in the file is different, so we can never have clashing hashcodes.

Fixes https://github.com/ably/ably-asset-tracking-android/issues/259


AndyTWF:
This has now run 3 times on CI without any failures.

QuintinWillison (approved):
:shipit: 🎉

PR #893, opened 2023-01-11
Publisher and Subscriber builders start method throws exception descr…

### LINKED ISSUE #876: Investigate documented `ConnectionException` thrown by `Publisher.Builder` and how users are meant to handle it
The documentation for `Publisher.Builder` states that it can throw a `ConnectionException` "if something goes wrong during connection initialization":

https://github.com/ably/ably-asset-tracking-android/blob/355bc214818da6f83b3cc2419c778542c227981f/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt#L274

It seems like this is because `DefaultAbly` can throw a `ConnectionException` "if something goes wrong during Ably SDK initialization".

https://github.com/ably/ably-asset-tracking-android/blob/355bc214818da6f83b3cc2419c778542c227981f/common/src/main/java/com/ably/tracking/common/Ably.kt#L249

We should understand what are the circumstances under which this might happen and explain to users how they are meant to recover from this situation. Can we handle it in a way that’s consistent with https://github.com/ably/ably-asset-tracking-android/issues/871?

davyskiba:
…iption updated
The investigation of #876 concluded that the `ConnectionException` thrown by the start functions of subscriber and publisher builders could only be caused by invalid configuration. This PR updates comments on those methods to better convey it to the user.

Resolves #876

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -245,7 +245,7 @@ private const val AUTH_TOKEN_CAPABILITY_ERROR_CODE = 40160
 
 class DefaultAbly
 /**
- * @throws ConnectionException if something goes wrong during Ably SDK initialization.
+ * @throws ConnectionException If connection configuration is invalid.
  */
 constructor(
     realtimeFactory: AblySdkRealtimeFactory,
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt
@@ -280,8 +280,8 @@ interface Publisher {
          * In order to detect device's location ACCESS_COARSE_LOCATION or ACCESS_FINE_LOCATION permission must be granted.
          *
          * @return A new publisher instance.
-         * @throws com.ably.tracking.BuilderConfigurationIncompleteException If all required params aren't set
-         * @throws ConnectionException If something goes wrong during connection initialization
+         * @throws com.ably.tracking.BuilderConfigurationIncompleteException If all required params aren't set.
+         * @throws ConnectionException If connection configuration is invalid.
          */
         @RequiresPermission(anyOf = [ACCESS_COARSE_LOCATION, ACCESS_FINE_LOCATION])
         @Throws(BuilderConfigurationIncompleteException::class, ConnectionException::class)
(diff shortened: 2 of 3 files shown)


KacperKluka:
@paddybyers I'm not sure if we've made a decision but it seems like a good idea to throw an exception instead of returning the SDK instance if the connection is `failed`. However, there are 2 problems:
1. We would have to wait for the connection state change (to either connected or failed) before returning an instance, and this might take time and suspend the code that creates the instance.
2. The connection in the `Publisher` SDK is started when the first trackable is added, so we can't do it when the instance is created.

KacperKluka (approved):
LGTM 👍

jaley:
Will there be a way for the application code to know that the user's Ably token is invalid without waiting for the user to add a trackable? I'm just thinking the app will probably want to sign the user out and re-authenticate (if needed) without having to wait for them to start a new ride?

paddybyers:
> Will there be a way for the application code to know that the user's Ably token is invalid without waiting for the user to add a trackable? I'm just thinking the app will probably want to sign the user out and re-authenticate (if needed) without having to wait for them to start a new ride?

In general there's no way for the client to know that the token is valid - it handles the token more or less opaquely. There's also no way to ask Ably if the token is valid, other than to attempt an operation.

Also ... if a token is valid at the time you check, it could become invalid at any time later (eg by expiring, or being revoked).

So I think it would be friendlier to be able to check at the time you create the library, but unless you go ahead and establish a connection, I don't think you're going to get a very reliable indication.

(If the token is a JWT then the library could parse it and get the expiry time from it and do that check - but then that comes with other issues, such as whether or not the time is set accurately on the phone.)

paddybyers:
> 2\. The connection in the `Publisher` SDK is started when the first trackable is added, so we can't do it when the instance is created.

ok, I thought we were already talking about the exception when you call `track()`.

> 1. We would have to wait for the connection state change (to either connected or failed) before returning an instance, and this might take time and suspend the code that creates the instance.

I wasn't really thinking of that. The connection can transition to `failed` at any time, even after having been connected, because for example the key that the token is derived from could be revoked. So we can't eliminate the need for the app to handle a possible `failed` transition asynchronously.

So the question is just - if we happen to have a `failed` connection at the time that `track()` is called, do we throw?

KacperKluka:
@paddybyers 
> So the question is just - if we happen to have a failed connection at the time that track() is called, do we throw?

Yes, when we start the Ably connection there are 2 situations in which it will throw:
- If the connection is already in the `failed` state we throw the `ably.connection.reason` as the exception.
- After initiating the `connect()` operation we wait for the Ably connection state to change to either `connected` or `failed`. If it changes to the latter, we throw an exception with the connection state change's `reason` value as the exception.

PR #894, opened 2023-01-11
Wrap `ably-java`’s `ChannelStateChange` in our own type

lawrence-forooghian:
I wanted to write unit tests that checked how `DefaultAbly` behaves in response to `AblySdkRealtime.Connection.on` emitting various `ChannelStateChange` values. However, I discovered that `ably-java`’s `ChannelStateChange` class does not have a public constructor. (For some reason, the analogous `ConnectionStateChange` class _does_ have a public constructor; not sure if this difference is intentional.) Furthermore, we cannot use MockK to create a useful mock of it, since its `current` etc values are implemented as Java fields as opposed to properties.

I wasn’t sure whether to modify ably-java to make `ChannelStateChange` publicly constructible. I decided not to, partly because I wasn’t very keen to go through the process of another `ably-java` release, and partly because I’m not convinced that it _should_ be publicly constructible. I’m happy to be challenged on both these points, though.

Furthermore, using a type that can be mocked by MockK gives us extra benefits, like being able to verify which fields of `ConnectionStateChange` are actually being accessed in a given test scenario.

<del>I feel a little bit out of my depth, Java/Kotlin-wise, with the code that manages the mapping of `AblySdkRealtime` listeners to `ably-java` listeners, so would appreciate a close eye on that. I'm also not sure whether the approach is a bit myopic and it would be simply better to hold a strong reference and remove it appropriately when `off` is called.</del> - new approach taken now which removes this complexity.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -275,7 +275,7 @@ constructor(
                     )
                 }
             }
-            ably = ablySdkFactory.create(clientOptions)
+            ably = ablySdkFactory.createRealtime(clientOptions)
         } catch (exception: AblyException) {
             throw exception.errorInfo.toTrackingException().also {
                 logHandler?.w("$TAG Failed to create an Ably instance", it)
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -19,7 +19,6 @@ import com.ably.tracking.logging.LogHandler
 import com.google.gson.Gson
 import io.ably.lib.realtime.AblyRealtime
 import io.ably.lib.realtime.ChannelState
-import io.ably.lib.realtime.ChannelStateListener
 import io.ably.lib.realtime.CompletionListener
 import io.ably.lib.realtime.ConnectionState
 import io.ably.lib.realtime.ConnectionStateListener
@@ -243,17 +242,17 @@ private const val CHANNEL_NAME_PREFIX = "tracking:"
 private const val AGENT_HEADER_NAME = "ably-asset-tracking-android"
 private const val AUTH_TOKEN_CAPABILITY_ERROR_CODE = 40160
 
-class DefaultAbly
+class DefaultAbly<ChannelStateListenerType : AblySdkChannelStateListener>
 /**
  * @throws ConnectionException If connection configuration is invalid.
  */
 constructor(
-    ablySdkFactory: AblySdkFactory,
+    private val ablySdkFactory: AblySdkFactory<ChannelStateListenerType>,
     connectionConfiguration: ConnectionConfiguration,
     private val logHandler: LogHandler?,
 ) : Ably {
     private val gson = Gson()
-    private val ably: AblySdkRealtime
+    private val ably: AblySdkRealtime<ChannelStateListenerType>
     private val scope = CoroutineScope(Dispatchers.IO + SupervisorJob())
     private val TAG = createLoggingTag(this)
 
@@ -305,7 +304,11 @@ constructor(
                 // Initial state is launched in a fire-and-forget manner to not block this method on the listener() call
                 scope.launch { listener(ConnectionStateChange(currentChannelState, null)) }
             }
-            channel.on { listener(it.toTracking()) }
+
+            val wrappedChannelStateListener = ablySdkFactory.wrapChannelStateListener { _, stateChange ->
+                listener(stateChange.toTracking())
+            }
+            channel.on(wrappedChannelStateListener)
         }
     }
 
@@ -314,7 +317,7 @@ constructor(
      * a new auth token is requested and the operation is retried once more.
      * @throws ConnectionException if something goes wrong or the retry fails
      */
-    private suspend fun enterChannelPresence(channel: AblySdkRealtime.Channel, presenceData: PresenceData) {
+    private suspend fun enterChannelPresence(channel: AblySdkRealtime.Channel<ChannelStateListenerType>, presenceData: PresenceData) {
         try {
             enterPresenceSuspending(channel, presenceData)
         } catch (connectionException: ConnectionException) {
@@ -439,14 +442,14 @@ constructor(
         }
     }
 
-    private suspend fun tryDisconnectChannel(channelToRemove: AblySdkRealtime.Channel, presenceData: PresenceData) =
+    private suspend fun tryDisconnectChannel(channelToRemove: AblySdkRealtime.Channel<ChannelStateListenerType>, presenceData: PresenceData) =
         try {
             disconnectChannel(channelToRemove, presenceData)
         } catch (exception: Exception) {
             // no-op
         }
 
-    private suspend fun disconnectChannel(channelToRemove: AblySdkRealtime.Channel, presenceData: PresenceData) {
+    private suspend fun disconnectChannel(channelToRemove: AblySdkRealtime.Channel<ChannelStateListenerType>, presenceData: PresenceData) {
         retryChannelOperationIfConnectionResumeFails(channelToRemove) { channel ->
             leavePresence(channel, presenceData)
             channel.unsubscribe()
@@ -455,7 +458,7 @@ constructor(
         }
     }
 
-    private suspend fun failChannel(channel: AblySdkRealtime.Channel, presenceData: PresenceData, errorInfo: ErrorInfo) {
+    private suspend fun failChannel(channel: AblySdkRealtime.Channel<ChannelStateListenerType>, presenceData: PresenceData, errorInfo: ErrorInfo) {
         leavePresence(channel, presenceData)
         channel.unsubscribe()
         channel.presence.unsubscribe()
@@ -464,7 +467,7 @@ constructor(
         ably.channels.release(channel.name)
     }
 
-    private suspend fun leavePresence(channel: AblySdkRealtime.Channel, presenceData: PresenceData) {
+    private suspend fun leavePresence(channel: AblySdkRealtime.Channel<ChannelStateListenerType>, presenceData: PresenceData) {
         suspendCancellableCoroutine<Unit> { continuation ->
             try {
                 channel.presence.leave(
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt
```diff
@@ -101,17 +103,51 @@ constructor(clientOptions: ClientOptions) : AblySdkRealtime {
...
-        override fun off(listener: ChannelStateListener) {
-            channel.off(listener)
+        override fun off(listener: AblySdkRealtime.ChannelStateListener) {
+            var ablyJavaListeners: List<ChannelStateListener?> = listOf()
+            synchronized(this) {
+                ablyJavaListeners =
+                    listeners.entries.filter { it.value.get() == listener }.map { it.key }
+            }
+
+            ablyJavaListeners.mapNotNull { it }.forEach { channel.off(it) }
```
KacperKluka, line 150: We could refactor this code to be more Kotlin idiomatic but the most important thing that's missing is that we're never removing the listeners from the map. I think that after we unsubscribe a listener (by using `off()`) we should also remove it from the map 🤔 I've attached a suggestion to this comment with the refactoring and listeners removal. Feel free to apply it or create your own implementation 😁 

```suggestion
            val ablyJavaListeners: List<ChannelStateListener> = synchronized(this) {
                listeners.entries
                    .filter { it.value.get() == listener }
                    .mapNotNull { it.key }
                    .onEach { listeners.remove(it) }
            }

            ablyJavaListeners.forEach { channel.off(it) }

```

Changes from the suggestion in detail:
- changed `ablyJavaListeners` list items type to a non-nullable `ChannelStateListener` as map keys aren't nullable
- changed `ablyJavaListeners` to a `val` since the `synchronized` block can return a value
- place list manipulation operations on separate lines where there was more than 1 operation
- move the listener null check to the `synchronized` block where the list is created
- remove listeners that will be unsubscribed within an `onEach` operation in the `synchronized` block

KacperKluka (changes requested):
A few suggestions from my side, but overall looks good 👍

common/src/main/java/com/ably/tracking/common/AblySdk.kt
```diff
@@ -39,6 +39,18 @@ interface AblySdkRealtime {
...
     }
 
+    fun interface ChannelStateListener {
+        data class ChannelStateChange(
+            val event: ChannelEvent,
+            val current: ChannelState,
+            val previous: ChannelState,
+            val reason: ErrorInfo?,
+            val resumed: Boolean
+        )
```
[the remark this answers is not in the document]
lawrence-forooghian: What difficulties were you thinking of? In my usage so far, I've been able to create mocks of this data class using `mockk<AblySdkRealtime.ChannelStateListener.ChannelStateChange>()`, and to use the normal `every` and `verify` methods on this mock’s properties with the desired effects.

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -305,10 +305,21 @@ constructor(
...
-            channel.on { listener(it.toTracking()) }
+            channel.on { listener(convertChannelStateChangeToTracking(it)) }
         }
     }
 
+    /**
+     * Converts Ably Realtime channel state change events to the equivalent [ConnectionStateChange] API
+     * presented to users of the Ably Asset Tracking SDKs.
+     */
+    private fun convertChannelStateChangeToTracking(channelStateChange: AblySdkRealtime.ChannelStateListener.ChannelStateChange): ConnectionStateChange {
```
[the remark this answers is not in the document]
lawrence-forooghian: I thought that the decision I'd made here was consistent with what you'd said in https://github.com/ably/ably-asset-tracking-android/pull/868#discussion_r1058889294, about extension functions only being intended for types that we don't control?

> The point of using extension functions is to extend the functionality of objects you can't control. However, right now we have control over the `AblySDKRealtime` interface so instead of using an extension function we can simply add the `closeSuspending()` method to the interface 🤔 But this is not critical, so feel free to do it as a part of this PR or create an issue for changing it later 😉

Since we have control of the newly-created `AblySdkRealtime.ChannelStateListener.ChannelStateChange` type, I didn't think an extension was appropriate?

common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt
lawrence-forooghian: I've actually realised that there's a broader problem with both of these approaches. I think I have a different approach, will comment later.

lawrence-forooghian:
@KacperKluka I realised that there was a large problem with the approach I'd taken above. We can't use any approach which stores state inside a `DefaultAblySdkRealtime.Channel` object, since these objects are very short-lived; we [re-create them each time a channel is fetched](https://github.com/ably/ably-asset-tracking-android/blob/94f4764ab89fb8383345e39a080379207564e819/common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt#L69-L95).

I suppose that we could change it so that the channels are not re-created each time, but then we'd have to make sure that we perform other kinds of bookkeeping such as getting rid of the channel when `dispose` is called on the `Channels` instance, which I'd prefer not to do, and leave that to `ably-java`.

So I’ve taken a different approach, which is actually much simpler. We store the `ably-java` listener inside the AAT listener itself. This also means there's no mutable state, and nothing that needs to be synchronised.

The only downside is that the `AblySdk*` interfaces get a bit more complicated - we need to introduce a generic parameter to represent the listener.

common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt
lawrence-forooghian: See https://github.com/ably/ably-asset-tracking-android/pull/894#issuecomment-1381248789 for the problem and the new approach.

lawrence-forooghian:
(I'd also suggest reviewing the PR commit by commit now, since there are some renamings that will make it hard to review as one diff.)

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -243,17 +242,17 @@ private const val CHANNEL_NAME_PREFIX = "tracking:"
 private const val AGENT_HEADER_NAME = "ably-asset-tracking-android"
 private const val AUTH_TOKEN_CAPABILITY_ERROR_CODE = 40160
 
-class DefaultAbly
+class DefaultAbly<T : AblySdkChannelStateListener>
```
KacperKluka, line 245: Perhaps a bit controversial but I usually try to [avoid single-letter generics](https://www.vojtechruzicka.com/break-java-generics-naming-convention/) as they introduce unnecessary cognitive load while analyzing the code 😉 What I do is I take the class that generic should extend and append `Type` to its end (like [here](https://github.com/ably/ably-asset-tracking-android/blob/main/common/src/main/java/com/ably/tracking/common/workerqueue/Worker.kt#L11)) but I wonder if that won't be too verbose and long in this case. Therefore, what would you think about renaming the `T` to `ChannelStateListenerType`? 🤔

common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt
```diff
@@ -16,16 +16,39 @@ import io.ably.lib.types.Message
...
+    override fun createRealtime(clientOptions: ClientOptions): AblySdkRealtime<DefaultAblySdkChannelStateListener> {
         return DefaultAblySdkRealtime(clientOptions)
     }
+
+    override fun wrapChannelStateListener(underlyingListener: AblySdkFactory.UnderlyingChannelStateListener<DefaultAblySdkChannelStateListener>): DefaultAblySdkChannelStateListener {
+        return DefaultAblySdkChannelStateListener(underlyingListener)
+    }
+}
+
+class DefaultAblySdkChannelStateListener(val underlyingListener: AblySdkFactory.UnderlyingChannelStateListener<DefaultAblySdkChannelStateListener>) :
```
KacperKluka, line 31: The IDE suggests me that the `underlyingListener` could be a `private val`, so if you don't plan to access it from tests or other places we might as well make it private 😉

KacperKluka (changes requested):
Almost there 👍

common/src/main/java/com/ably/tracking/common/AblySdk.kt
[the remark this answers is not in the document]
KacperKluka: Actually, I was wrong, `mockk` technically allows you to mock `data class`es, sorry for that 😅 Still, this is not thought to be a good idea as discussed in a [mockk issue](https://github.com/mockk/mockk/issues/508) (arguments [here](https://github.com/mockk/mockk/issues/508#issuecomment-716763259) and [here](https://github.com/mockk/mockk/issues/508#issuecomment-712435784)) so we probably should either stick to using real instances of this `data class` in our tests or change it to an interface, depending on our use-case 🤔

common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt
KacperKluka: Due to the above changes, this conversation is not valid, therefore, I'll resolve it.

common/src/main/java/com/ably/tracking/common/Ably.kt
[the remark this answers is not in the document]
lawrence-forooghian: OK, I've kept the previous extension and just changed the type that it's on 🙂

common/src/main/java/com/ably/tracking/common/AblySdk.kt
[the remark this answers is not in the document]
lawrence-forooghian: I've changed it to an interface 👍

common/src/main/java/com/ably/tracking/common/Ably.kt
lawrence-forooghian: Changed it, although it's longer now it's definitely more readable!

common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt
lawrence-forooghian: Good catch! Changed it.

KacperKluka (approved):
Awesome work! 💪  LGTM, thanks 👍

common/src/main/java/com/ably/tracking/common/DefaultAblySdk.kt
lawrence-forooghian: Gonna resolve this conversation now you've approved and it's a small thing 🙂

PR #911, opened 2023-01-16
wait for publisher emissions before performing test assertion

### LINKED ISSUE #899: `createAndStartPublisherAndSubscriberAndWaitUntilDataEnds` IndexOutOfBoundsException
```
com.ably.tracking.tests.PublisherAndSubscriberTests > createAndStartPublisherAndSubscriberAndWaitUntilDataEnds[test(AVD) - 8.1.0] FAILED 
	java.lang.IndexOutOfBoundsException: Index: 5, Size: 5
	at java.util.ArrayList.get(ArrayList.java:437)
```

Failing in https://github.com/ably/ably-asset-tracking-android/actions/runs/3910193395/jobs/6683211315

AndyTWF:
Emitting events on the publishers location channel is done on the same coroutine scope as, but outside of the worker queue. It's also initiated after we send the location update to any subscriber over the realtime service.

In a very specific order of events, this could cause the test to fail as the publisher will have sent the events to the subscriber, but not emitted them onto its location channel.

This change fixes the problem by ensuring that the publisher has emitted at least as many events as received by the subscriber before allowing test assertions to made.

Fixes #899

### DIFF
--- a/integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
+++ b/integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
@@ -16,6 +16,7 @@ import java.util.UUID
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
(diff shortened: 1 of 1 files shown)


integration-testing-app/src/androidTest/java/com/ably/tracking/tests/PublisherAndSubscriberTests.kt
```diff
@@ -94,6 +95,17 @@ class PublisherAndSubscriberTests {
...
         publisherStoppedExpectation.assertFulfilled()
         subscriberStoppedExpectation.assertFulfilled()
+
+        /*
+            Wait for everything to have been emitted onto the publisher locations channel,
+            as this happens on the same coroutine scope as, but outside of, the worker queue.
+         */
+        runBlocking {
+            while (publishedLocations.size < receivedLocations.size) {
+                delay(100)
```
KacperKluka, line 105: This could become an infinite loop (perhaps not now but if we change the tests or implementation in the future it might), therefore, maybe let's wrap the `while` within a `withTimeout` to make sure the test will fail if this takes too long? 🤔

KacperKluka (approved):
LGTM 👍

PR #915, opened 2023-01-16
Upgrade build tools to make Layer 7 proxy branch work

QuintinWillison:
This pull request will hopefully merge to https://github.com/ably/ably-asset-tracking-android/pull/886 in order to fix the compilation.

I've spent several hours getting to this stage from what seemed like an innocuous, small lint error. From [this Slack message (internal)](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1673858347027739?thread_ts=1673628424.974169&cid=C01EPJENRD0) from @davyskiba I started investigating solving `NewApi` lint error for `java.lang.Iterable#forEach` based on [this Stackoverflow answer](https://stackoverflow.com/questions/68470163/error-call-requires-api-level-24-current-min-is-21-java-util-mapforeach-ne). As far as I can see, as suggested, https://github.com/ably/ably-asset-tracking-android/commit/207d5552177380d3096a972df16e34fb19f7f4f0 does solve that.

But, in the process, I had to upgrade Gradle and other bits of tooling which then has opened up other issues.


QuintinWillison:
Current state of this pull request is that I'm seeing this locally (hopefully CI checks, once done, will see the same):

```
> Task :integration-testing-app:lintDebug FAILED
/Users/quintinwillison/.gradle/caches/transforms-3/c1020ffcb6f66cd5af519dbe40cd56d2/transformed/fragment-1.4.0/jars/lint.jar: Error: Library lint checks out of date;
these checks will be skipped!

Lint found an issue registry (androidx.fragment.lint.FragmentIssueRegistry)
which was compiled against an older version of lint
than this one. This is usually fine, but not in this
case; some basic verification shows that the lint
check jar references (for example) the following API
which is no longer valid in this version of lint:
org.jetbrains.uast.kotlin.KotlinUClass: org.jetbrains.kotlin.psi.KtClassOrObject getKtClass()
(Referenced from androidx/fragment/lint/OnCreateDialogIncorrectCallbackDetector$UastHandler.class)

Therefore, this lint check library is not included
in analysis. This affects the following lint checks:
FragmentTagUsage
FragmentAddMenuProvider
FragmentBackPressedCallback
FragmentLiveDataObserve
UseRequireInsteadOfGet
UseGetLayoutInflater
DialogFragmentCallbacksDetector
UnsafeRepeatOnLifecycleDetector
DetachAndAttachSameFragment

Recompile the checks against the latest version, or if
this is a check bundled with a third-party library, see
if there is a more recent version available.

Version of Lint API this lint check is using is 11.
The Lint API version currently running is 13 (7.3+). [ObsoleteLintCustomCheck]
```

Despite https://github.com/ably/ably-asset-tracking-android/pull/915/commits/08c4f0606d95153429a3d4f14fadc1d16c58caa9.

Ever-so-slightly pulling my hair out at this point in frustration.

QuintinWillison:
Thanks to @KacperKluka for direction on https://github.com/ably/ably-asset-tracking-android/pull/915/commits/4cb02e4d18fc420336de925d7ad48e0c274210fb, which led me to improve it in https://github.com/ably/ably-asset-tracking-android/pull/915/commits/404dd869aa1f5d1e5110511c104650943607d75d.

Thanks to @AndyTWF for analysis on the Lint 11 vs 13 error `ObsoleteLintCustomCheck` which led me to override a transitive dependency version in https://github.com/ably/ably-asset-tracking-android/pull/915/commits/b4f773f01f693fb836e37bb0b1bfdc56f2436c42.

We're still not there, yet. 😅 -- currently I'm seeing the following lint errors locally when I run `./gradlew check testDebugUnitTestCoverage testReleaseUnitTestCoverage --profile -PruntimeSecrets=USE_DUMMY_EMPTY_STRING_VALUES`:

```
> Task :publishing-sdk:lintDebug FAILED
/Users/quintinwillison/code/ably/ably-asset-tracking-android/publishing-sdk/src/main/java/com/ably/tracking/publisher/Mapbox.kt:346: Error: Call requires permission which may be rejected by user: code should explicitly check to see if permission is available (with checkPermission) or explicitly handle a potential SecurityException [MissingPermission]
            NotificationManagerCompat.from(context).notify(notificationId, notificationProvider.getNotification())
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/quintinwillison/code/ably/ably-asset-tracking-android/publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/DublicateTrackableGuard.kt:53: Error: Call requires API level 24 (current min is 21): java.lang.Iterable#forEach [NewApi]
        duplicateAddCallsHandlers[trackable]?.forEach { handler -> handler(result) }
                                              ~~~~~~~
/Users/quintinwillison/code/ably/ably-asset-tracking-android/publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/TrackableRemovalGuard.kt:38: Error: Call requires API level 24 (current min is 21): java.lang.Iterable#forEach [NewApi]
        handlers?.forEach {
                  ~~~~~~~
```

Neither of them look like the sort of errors we should safely downgrade to warning/informational. I had already started trying to address the `MissingPermission` error in https://github.com/ably/ably-asset-tracking-android/pull/915/commits/0f22e74938def0ddbc1cbe1c89423978c475ddb0, so it may well be just that we also need to add something into the code also (perhaps as far as up to user-API level).

publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/DublicateTrackableGuard.kt
```diff
@@ -50,7 +50,7 @@ internal class DefaultDuplicateTrackableGuard : DuplicateTrackableGuard {
      */
     override fun finishAddingTrackable(trackable: Trackable, result: Result<AddTrackableResult>) {
         trackablesCurrentlyBeingAdded.remove(trackable)
-        duplicateAddCallsHandlers[trackable]?.forEach { handler -> handler(result) }
+        duplicateAddCallsHandlers[trackable]?.iterator()?.forEach { handler -> handler(result) }
```
QuintinWillison, line 53: @AndyTWF I wonder whether here, and for the other instance, it would be appropriate to fail early and NPE if `iterator()` returns `null` - that is, `iterator()!!` rather than `iterator()?`. 🤔

AndyTWF: Done in https://github.com/ably/ably-asset-tracking-android/pull/915/commits/b64f88ff541923af05edab847e842dc95488ce52

AndyTWF: @QuintinWillison I had to revert this as the !! operator appears to check the entire result of the expression (so if we get a NULL from `duplicateAddCallsHandlers[trackable]`, it still throws the NPE)

QuintinWillison: Thanks! 😁 FYI, @KacperKluka and @davyskiba, in case you have any further thoughts on this.

KacperKluka: If we want to have the NPE we could use `?.let {}` to get a non-null list of handlers and then use `iterator()!!` to fail early in case of a NPE 😉 It would look like this:
```suggestion
        duplicateAddCallsHandlers[trackable]?.let { it.iterator()!!.forEach { handler -> handler(result) } }
```
I'd also add a comment that this is intentional because we want to fail early on NPEs as otherwise someone might be tempted to change it to null-safe calls in the future 😅

QuintinWillison: Thanks, @KacperKluka. That looks like the right balance, and good point regarding commentary. I'll push a commit shortly for this.

QuintinWillison: Once the `let` was inserted it became clear that `iterator()` itself, in isolation, can't return `null` (oops, should have checked that API!). No need for explicit NPE injection after all. 😁 https://github.com/ably/ably-asset-tracking-android/pull/915/commits/d5f6719737c53fef046da6eb0d8f3310cd7b90b0

jaley (approved):
Thanks Quintin! This looks good to me. I have no strong opinions on all the version bumps, it looks like they all move in the right direction and mostly impact test code anyway, so happy to merge from my perspective.

AndyTWF (approved):

PR #916, opened 2023-01-16
901 simplify transition assertions

### LINKED ISSUE #901: Reduce complexity of state transition assertions in `NetworkConnectivityTests`
Noticed by @davyskiba and described in [this review comment](https://github.com/ably/ably-asset-tracking-android/pull/866#discussion_r1069361037).

davyskiba:
I have updated `waitForStateTransition` function to use a more direct approach, namely using `map` and `first` on the flow and moving this operation to the `TrackableStateReceiver.` This allowed replacing `Expectations` with blocking calls and removing now unused coroutine scope from `TestResources.`

Resolves #901

### DIFF
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -1,6 +1,9 @@
 package com.ably.tracking.test.android.common
 
 import com.ably.tracking.TrackableState
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.mapNotNull
 import kotlin.reflect.KClass
 
 /**
@@ -160,8 +163,6 @@ class TrackableStateReceiver(
     private val expectedStates: Set<KClass<out TrackableState>>,
     private val failureStates: Set<KClass<out TrackableState>>
 ) {
-    val outcome = BooleanExpectation(label)
-
     companion object {
         fun onlineWithoutFail(label: String) = TrackableStateReceiver(
             label,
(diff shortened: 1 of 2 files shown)


ikbalkaya (approved):
LGTM

jaley (approved):
Nice improvement, LGTM! Quick question for you... the wait I added [here](https://github.com/ably/ably-asset-tracking-android/pull/916/files#diff-cc8d63bc5d25adbe0a89eab3912b227b0536c5f1753f7f716982c3814fc15630L254) was actually because I had issues with coroutines continuing to run and listen to further `StateFlow` transitions throughout the test, even when we were done with that specific assertion. That cause spurious logging and some invalid failing assertions. With this change, will the coroutines listening on `StateFlow`s and making assertions in the `waitForStateTransition` call still be cleaned up by the time the function returns?

davyskiba:
@jaley Yes, [first](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/first.html) will cancel once it gets any value

jaley:
> @jaley Yes, [first](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/first.html) will cancel once it gets any value

@davyskiba - will `first` just take the first `Boolean?` returned by `receive` in the `map` transform? If so, that's a change in behaviour. We deliberately skip any of the `ignored` transitions, as those are considered harmless but not the state we're waiting to see. Should `first` actually be some kind of `takeWhile` call checking for `null`?

davyskiba:
@jaley nulls are already filtered by [mapNotNull](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null.html) and first returns `Boolean,` so the behavior should stay the same. `MapNotNull` could be replaced with `map` + `filterNotNull` to make it more visible. Do you think it is worth doing?

PR #917, opened 2023-01-17
android test orchestrator configured

davyskiba:
Added android test orchestrator configuration to clear package data after each test, which should improve test stability.

Android test orchestrator is a test runner developed by Google, it is focused on test isolation mainly by running each test in a separate process. Quickstart guide can be found [here](https://developer.android.com/training/testing/instrumented-tests/androidx-test-libraries/runner#use-android)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,6 +119,15 @@ subprojects {
 
                 testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
 
+                // The following argument makes the Android Test Orchestrator run its
+                // "pm clear" command after each test invocation. This command ensures
+                // that the app's state is completely cleared between tests.
+                testInstrumentationRunnerArguments clearPackageData: 'true'
+
+                testOptions {
+                    execution 'ANDROIDX_TEST_ORCHESTRATOR'
+                }
+
                 if (evaluatedSubProjectIsAnApp) {
                     multiDexEnabled true
                 }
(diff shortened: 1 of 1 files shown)


AndyTWF (approved):

ikbalkaya (approved):
Looks good to me. A request for the description of this PR: Can you put a short explanation of what test orchestrator is, why is it used and any other links that would be useful to other people?

QuintinWillison (commented):
My understanding, from the opening commentary for this pull request, is that this change is speculative:

> Added android test orchestrator configuration to clear package data after each test, which should improve test stability.

There is no linked issue indicating particular types of test failure that could be stabilised by making this change, so as a reviewer I remain somewhat under-informed as to why we would want to do this.

I imagine that the desired outcome is to create a more stable runtime environment for tests by running each in more forced isolation. The result, I fear, is that we could end up hiding other kinds of failures - perhaps even SDK bugs - that relate to teardown. One of the things that we are measuring when running integration tests is whether they naturally isolate from one another, as that in itself can reveal subtle issues (it's why I've always promoted us 'properly' closing all resources, and waiting for them to close, before ending a test).

Without further reasoning and justification to allay my fears, I'm unable to approve this pull request in its current form.

davyskiba:
After reading Quintin's response, I'm sure that now is not the right time for this change, might be worth to look into this later

PR #920, opened 2023-01-17
Include API version 33 in tests

AndyTWF:
This change brings the Android versions we test against to:

5, 7, 8.1, 10 and 13.


AndyTWF:
Closing this - was just using it to get an idea of the state of the test suite when running on the latest API level

PR #921, opened 2023-01-17
local ably-java build experiments

ikbalkaya:
This PR is just for experiments of AAT against different version of ably-java and it aims to be used to check CI results. It will be destroyed later


ikbalkaya:
Closing this since this is no longer required here

Issue #918, opened 2023-01-17
Add an integration test for network fault coverage to Subscriber component


jaley:
Many of the faults simulated in #865 and #886 apply equally to the Subscriber component as they do the Publisher. We went the approach of making `NetworkConnectivityTest` a publisher test as the scope for failures there was greater, and having internal access to the `DefaultPublisher` constructors greatly simplified injecting an `AblyRealtime` instance configured to connect to `localhost` without TLS. All of the fault simulation logic and proxy server implementations, however, are in `android-test-common` and so could be used from an integration test in the `Subscriber`, meaning we only need to write a test that uses it and exercises the `Subscriber` API appropriately during faults. Given that this is a separate API to the `Publisher` designed for different customer use cases, there's likely very little wastage in just implementing a whole new test that uses the existing proxy and fault code.

If we adopt a similar approach to that taken in the `Publisher`, the requirements are essentially:

* Add a new integration test to the subscriber component
* Make use of Junit `Paramaterized` test runner to inject the desired fault implementations
* Create some helpers to instantiate `DefaultSubscriber` with an `AblyRealtime` instance that was constructed with `ClientOptions` provided by the proxy implementations (i.e. configured to connect locally without TLS)
* Write integration-level tests that exercise `Subscriber`, activating the injected faults at opportune times and making assertions about the observed state of trackables and/or location updates before, after and during faults.

In terms of what use cases to exercise, an initial cursory look over the `Subscriber` interface suggests:

* Activate fault before starting and stopping the subscriber - should not throw exceptions, as per design change agreed in #879 
* Activate fault during tracking, publish a location update (can do this directly with Ably channel), ensure that the location update arrives when fault resolved by monitoring `subscriber.locations` and/or `subscriber.rawLocations`
* Activate fault during tracking and ensure the expected `Trackable` state transitions happen by watching `subscriber.trackableStates`, and that this recovers when fault resolved
* Activate fault during tracking and ensure `subscriber.publisherPresent` flag updates as expected, and updates when fault resolved

Perhaps lower priority:
* ... same for publisher resolution updates through `subscriber.resolutions`?
* Try updating `resolutionPreferences` during a fault, ensure that this takes affect after fault resolved.

Would be great to get a review on the above from @KacperKluka, @QuintinWillison and perhaps @paddybyers to ensure that the proposed tests are in line with (new) expected behaviours, and any comments on implementation strategy.

jaley:
[...]

PR #924, opened 2023-01-18
Subscriber Connectivity Tests

AndyTWF:
Adds test cases for the main requirements of #918 to test subscriber behaviour when there are network connectivity issues.

Note, some tests are skipped due to the `skipTest` property on the fault - to run all tests you can either set the property to `false` for whichever test you want to run, or remove the `assumeFalse` call in the test classes setup method.

Current test failure status is below, many of the tests that hang may be related to https://github.com/ably/ably-java/issues/474, which we can confirm once the fix is merged.

**Fault before starting subscriber:**

TCP Connection Refused - timeout waiting for Ably to connect, as per publisher
TCP Connection Unresponsive - timeout waiting for Ably to connect, as per publisher
Attach unresponsive - throws Exception re attach timeout
Disconnect with failed resume - exception thrown
Enter failed with non-fatal nack - exception thrown
Disconnect and suspend - timeout waiting for Ably to connect
Enter unresponsive - hang
 
**After starting/before stopping subscriber:**

TCP Connection Refused - hang
TCP Connection Unresponsive - hang
Detach unresponsive - throws exception re detach timeout
Disconnect and suspend - hang
Reenter on resume failed - hang
Enter unresponsive - hang

**Tests checking we get location and publisher updates after a fault:**

TCP Connection Unresponsive - hang
Disconnect and suspend - hang


AndyTWF:
Just a note on testing:

- API Level 21 does not run these tests due to issues with KTOR (see comment in setup method, its the same as publisher tests).
- The tests still sometimes crash the emulator, this is something I've also witnessed on the publisher tests too.

It would be great to get some feedback on things so far whilst I look into emulator-related problems :)

subscribing-sdk/build.gradle
```diff
@@ -17,6 +17,7 @@ android {
 dependencies {
     api project(':core-sdk')
     implementation project(':common')
+    implementation "io.ably:ably-android:$ably_core_version"
```
QuintinWillison, line 20: I think you can use `androidTestImplementation` here, given you're only needing this to be added for the `androidTest` source set.

QuintinWillison (commented):
I've not looked at the tests themselves in depth but a quick, immediate observation. 😁

subscribing-sdk/build.gradle
AndyTWF: Ah I thought I'd changed that! Well, I have now... :D 062db92d

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    /**
+     * Test that Subscriber can handle the given fault occurring before a user
+     * starts the subscriber.
+     *
+     * We expect the subscriber to not throw an exception.
+     */
+    @Test
+    fun faultBeforeStartingSubscriber() {
```
jaley, line 107: There currently aren't any assertions in these first couple of tests. I know that's actually enough to flag the issues with us throwing exceptions on `start()`, but we should probably assert that side effects of starting the subscriber have happened? IIUC, the subscriber's `clientId` should be present on the Ably channel? I'm currently looking at finishing off #928 , which should give us some tools to check that

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
```diff
@@ -0,0 +1,509 @@
...
+     * Check that after resolution, trackable states (via channel presence)
+     * are updated.
+     */
+    @Test
+    fun faultWhilstTrackingTrackableStatesArriveAfterResolution() {
+        withResources { resources ->
+            val subscriber = resources.getSubscriber()
+            resources.fault.enable()
+
+            // Begin listening for trackable state changes - the initial state is offline, so we have to ignore the first one
```
jaley, line 200: It's not quite true that all faults should cause `Trackable`s to go `Offline` while active. Some of them are deliberately non-fatal errors, which should sort of "self resolve" after a couple of retries. The Publisher tests have the `TrackableStateReceiver` provided by each `FaulySimulation` implementation to help tell you what state the fault thinks a Trackable should be in during a given testing stage. Not sure how well that translates to this test, but I wouldn't be surprised if we run into trouble here...

jaley (commented):
Couple of comments to think about. Haven't quite convinced myself whether they're a problem or not for now!

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
AndyTWF: Agreed - I'll keep an eye out for #928 and add in some suitable side-effect changes once that's ready :)

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
AndyTWF: Just adding as a decision log:

@jaley is looking at how we can simplify the state transitions code, so will hold off actioning this until he's had time to do a proof of concept

AndyTWF: @jaley This is now done. Would you be able to check it over at some point and make sure I've gotten the right idea? :)

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
```diff
@@ -0,0 +1,1115 @@
...
+                faultType = resources.fault.type,
+                locationUpdate = null,
+                publisherResolution = null,
+                subscriberResolution = subscriberResolution,
+                subscriberResolutionPreferenceFlow = resources.subscriberResolutions
+            ).waitForStateTransition {
+                // Connect up a publisher to do publisher things
+                defaultAbly.updatePresenceData(resources.trackableId, PresenceData(ClientTypes.PUBLISHER, publisherResolution, false))
+
+                val locationSent = BooleanExpectation("Location sent successfully on Ably channel")
```
davyskiba, line 145: Maybe this is a right time to introduce a suspending version of `sendEnhancedLocation` instead of using expectation?

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
```diff
@@ -0,0 +1,1115 @@
...
+                ) = DefaultAblySdkChannelStateListener(underlyingListener)
+            }
+            val connectionConfiguration = ConnectionConfiguration(
+                Authentication.basic(
+                    SUBSCRIBER_CLIENT_ID,
+                    fault.proxy.clientOptions().key
+                )
+            )
+
+            runBlocking {
```
davyskiba, line 500: Could you update the tests to match the changes regarding coroutines introduced in #994?

davyskiba (commented):
Overall lookgs good :)

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
AndyTWF: Done!

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
AndyTWF: Agreed - done!

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
davyskiba: Now, most of the `runBlocking` calls can be replaced by adding `suspend` to the function declaration. Should have been more specific, sorry :(

AndyTWF: Done :)

davyskiba (approved):

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
davyskiba: awesome <3

Issue #925, opened 2023-01-18
Make `NetworkConnectivityTests` verify expected side-effects of operations publisher SDK claims was successful


lawrence-forooghian:
We noticed that #883 was incorrectly causing a trackable to enter an `online` state even though the enter presence operation had failed. The test `faultBeforeAddingTrackable[EnterFailedWithNonfatalNack]`, which was previously failing due to the trackable not transitioning to `online`, started to pass.

We want to verify that, when the publisher claims to have successfully completed an operation, it has actually performed the side-effects that we'd expect from that operation.

Namely:

- after an added trackable transitions to the `online` state, a subscriber to the trackable’s channel should see the publisher as being present;
- after the publisher receives a location update, the subscriber should receive a location update message on the channel (not sure exactly what the publisher API trigger would be here though, perhaps it needs to be a separate issue not related to publisher API interactions)
- after a successful call to the publisher’s `remove()` method, a subscriber to the trackable’s channel should not see the publisher as present

jaley:
Resolved by #933 

We needed to simplify the `TrackableStateReciever` design to enable tests to make their own assertions about what state trackables should be in based on whether the fault active is intended to be fatal or not. With that information, it's also possible to have the test code verify side-effects directly.

PR #928, opened 2023-01-19
Verify expected side effects when trackable enters online state

lawrence-forooghian:
There’s a bunch of TODOs to address but I think the broad strokes are there.

TODO `faultBeforeAddingTrackable[DetachUnresponsive]` is now failing because it’s not seeing the expected side effects; not had a chance to look into it.

Part of #925 (there are still other states to handle).


lawrence-forooghian:
Replaced by #933.

Issue #927, opened 2023-01-19
Retry behaviour improvements for Ably API calls


jaley:
`Publisher` does not currently retry in all circumstances that we need it to retry. We should design and implement a retry mechanism that causes the following behaviour:

### Connections

* Connection retry attempts are handled indefinitely by `ably-java` - there's no need to retry these in AAT
* Simply allow `ably-java` to continue to attempt connection until it succeeds, as per its own retry policies
* Use the state change listener to check for a fatal error (e.g. invalid credentials). There will be a `4xx` status code for fatal errors.

### Channel Attachments

* Similarly to connections, `ably-java` will perform retries automatically
* Just listen for fatal errors so that we correctly move to `Failed` state when these happens

### Presence

* `enter()`
  * Check callbacks, if you get a `5xx` error, schedule retry for 15 seconds later.
  * If you get a `4xx`, move to failed state, detach from channel and, if this is the last trackable, close connection 
  * Continue retrying on `5xx` errors indefinitely
  * Note that the presence data (resolution preference) may change between retries
* `leave()`
  * Give leave a total 30s time limit to complete successfully
  * If you get a, `onError` callback within that 30 seconds, check the error code:
    * If it's a `5xx`, schedule a retry in 15 seconds (this doesn't restart overall timeout)
    * If it's a`4xx`, give up, detach and disconnect if final trackable
  * If 30s expire, give up, detach and disconnect if final trackable

To relate these back to the `Publisher` APIs:

* `start()`
  * Triggers connection to Ably (TBC?)
  * Should return immediately and follow normal connection retry policy above in background, as connectivity state is expected to vary during normal operation
* `Publisher.track()` and `Publisher.add()`:
  * Trigger connection (if not connected), channel attachment, and presence enter
  * Should return immediately and handle all of this in the background, as per retry policies above
  * `TrackableState` used to check current status
* `Publisher.remove()`
  * Triggers presence leave, channel detach, disconnect if final trackable
  * Should return immediately and let the channel/presence cleanup and retry behaviours above happen in background
  * `StateFlow<TrackableState>` change should also be immediate
* `Publisher.stop()`
  * Similar to `remove()` (?)

@paddybyers , @KacperKluka  -- initial retry behaviour spec to review

jaley:
> we would need to manually set the trackable state to Offline

[...]

PR #929, opened 2023-01-20
Add methods for detecting retriable and fatal Ably exceptions

KacperKluka:
Added helper methods that will enable us to detect fatal and retriable errors. This should allow us to implement appropriate retry mechanisms for certain Ably operations (like presence enter and leave).

Open questions:
- should we only look at the `code` property or maybe also at the `statusCode`?
- how should we classify exceptions that have codes that doesn't start with either `4` or `5`?

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
+++ b/common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
@@ -227,7 +227,7 @@ private fun TokenAuthException.toAblyException(): AblyException =
  * Fatal errors have status codes like 4xx (e.g. 400).
  */
 fun ConnectionException.isFatal(): Boolean {
-    return (400 .. 499).contains(errorInformation.statusCode)
+    return (400..499).contains(errorInformation.statusCode)
 }
 
 /**
@@ -235,5 +235,5 @@ fun ConnectionException.isFatal(): Boolean {
  * Non-fatal errors have status codes in range 500-504.
  */
 fun ConnectionException.isRetriable(): Boolean {
-    return (500 .. 504).contains(errorInformation.statusCode)
+    return (500..504).contains(errorInformation.statusCode)
 }
--- a/common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
+++ b/common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
@@ -224,18 +224,16 @@ private fun TokenAuthException.toAblyException(): AblyException =
 
 /**
  * Indicates whether the exception from Ably is fatal and we should not attempt to retry it.
- * Fatal errors have codes like 4xx (e.g. 400).
+ * Fatal errors have status codes like 4xx (e.g. 400).
  */
 fun ConnectionException.isFatal(): Boolean {
-    val firstErrorCodeDigit = errorInformation.code.toString().first()
-    return firstErrorCodeDigit == '4'
+    return (400 .. 499).contains(errorInformation.statusCode)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
```diff
@@ -221,3 +221,21 @@ private fun TokenAuthException.toAblyException(): AblyException =
         is CouldNotFetchTokenException ->
             AblyException.fromErrorInfo(ErrorInfo(message, 401, 100_002))
     }
+
+/**
+ * Indicates whether the exception from Ably is fatal and we should not attempt to retry it.
+ * Fatal errors have codes like 4xx (e.g. 400).
+ */
+fun ConnectionException.isFatal(): Boolean {
+    val firstErrorCodeDigit = errorInformation.code.toString().first()
```
paddybyers, line 230: This test needs to be on the `statusCode` and not the `code`. There are multiple code ranges that don't start with a 4 or a 5.

Personally I would be happier with a numerical comparison, ie `isRetriable(e) { return e.statusCode >= 500 }`

paddybyers (changes requested):
See inline comment

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
KacperKluka: I used the `code` because as I [mentioned in the issue](https://github.com/ably/ably-asset-tracking-android/issues/927#issuecomment-1398455100) I noticed that the `statusCode` is an optional field. So if it is not set its value will be defaulted to `0`. But I guess we should not worry about it because all important errors will have the `statusCode` correctly set, right? I'll make the changes suggested in your comment 👍

paddybyers: According to our spec, errors from REST requests that require a retry [are those with a `statusCode` of 500..504](https://sdk.ably.com/builds/ably/specification/main/features/#RSC15l3)

paddybyers: These are all the currently defined error codes from the realtime system: https://github.com/ably/ably-common/blob/main/protocol/errors.json

You can't know whether it's retriable or not from the `code` .

KacperKluka: I made the changes https://github.com/ably/ably-asset-tracking-android/pull/929/commits/3bda58378b8ca315839daeb5d4761de54ea85f57:
- using `statusCode` to check if an error is fatal or retriable
- errors with status code in the range 400-499 are considered as fatal
- errors with status code in the range 500-504 are considered as retriable
- all other codes are considered as non-fatal and non-retriable

paddybyers:
I don't think there's any case that's not a 4xx or a 5xx. I think I'd prefer that there's an error logged in that case - even if `statusCode == 0`.

davyskiba (approved):

KacperKluka:
> I don't think there's any case that's not a 4xx or a 5xx. I think I'd prefer that there's an error logged in that case - even if statusCode == 0.

@paddybyers I've created an issue for this (https://github.com/ably/ably-asset-tracking-android/issues/931) as this is a bit tricky to do right and seems a bit out of scope of this PR 😉

common/src/main/java/com/ably/tracking/common/AblyHelpers.kt
paddybyers: I don't think there's any case that's not a 4xx or a 5xx. I think I'd prefer that there's an error logged in that case - even if statusCode == 0.

PR #930, opened 2023-01-20
Fix swapped error codes when creating the malformed message exception

KacperKluka:
When looking through the codebase I've noticed that we have swapped the `code` and `statusCode` values when creating a malformed message exception. The `statusCode` should correspond to the HTTP code and is specified as the first integer parameter. The `code` is AAT specific in this case and is the last parameter of the `ErrorInfo` constructor.


davyskiba (approved):

QuintinWillison (approved):
Well spotted. Nice work! 😁

PR #932, opened 2023-01-23
Additional subscriber tests

AndyTWF:
Adds two more tests from the "low priority" requirement of #918.

- Testing that publisher resolution updates sent during a fault are received by the subscriber when the fault is cleared.
- Testing that subscriber resolution preference updates sent during a fault are received by the publisher when the fault is cleared.

### DIFF
--- a/subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
+++ b/subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
@@ -306,7 +306,7 @@ class NetworkConnectivityTests(private val testFault: FaultSimulation) {
     /**
      * Test that Subscriber can handle the given fault occurring whilst tracking,
      *
-     * Check that after resolution, changes to the channels publisher presence
+     * Check that after resolution, changes to the channels publisher resolutions
      * are received by the subscriber.
      */
     @OptIn(Experimental::class)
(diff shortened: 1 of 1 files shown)


subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
```diff
@@ -66,15 +66,15 @@ class NetworkConnectivityTests(private val testFault: FaultSimulation) {
         fun data() = listOf(
             arrayOf(NullTransportFault(BuildConfig.ABLY_API_KEY)),
             arrayOf(NullApplicationLayerFault(BuildConfig.ABLY_API_KEY)),
-            arrayOf(TcpConnectionRefused(BuildConfig.ABLY_API_KEY)),
+/*            arrayOf(TcpConnectionRefused(BuildConfig.ABLY_API_KEY)),
```
QuintinWillison, line 69: Do you want to commit this change? We have a 'skip test' property for this. 🤔

AndyTWF: Some of the tests were flakey for reasons different to the pub so had done this to allow a clean CI run - have now swapped this for a dedicated property on the fault for sub tests in https://github.com/ably/ably-asset-tracking-android/pull/932/commits/9b19ac85b2a20040ba35f806334e87b82c696d84 :)

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
```diff
@@ -81,7 +81,7 @@ class NetworkConnectivityTests(private val testFault: FaultSimulation) {
 
     @Before
     fun setUp() {
-        Assume.assumeFalse(testFault.skipTest)
```
QuintinWillison, line 84: I think we need to keep this line, as now we need to make sure that _both_ `skipTest` and `skipSubscriberTest` are `false`.

QuintinWillison (changes requested):

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
```diff
@@ -32,7 +32,10 @@ abstract class FaultSimulation {
      * The advantage of this approach is that the test code remains active and continually compiled as
      * a first class citizen of the codebase, while we work on other things to get it passing.
      */
-    open val skipTest: Boolean = false
+    open val skipPublisherTest: Boolean = false
+
+    // As per skipTest, but only skips on subscriber tests.
```
QuintinWillison, line 37: This comment is slightly incorrect, but actually perhaps redundant now that `skipTest` has been renamed to `skipPublisherTest`.

QuintinWillison (approved):
Much easier to understand now, thanks! Just one minor observation.

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
AndyTWF: Updated in https://github.com/ably/ably-asset-tracking-android/pull/932/commits/fd24b0025f13da92fa6eef24d74fd4c4e44e8974

PR #933, opened 2023-01-23
Improve `NetworkingConnectivityTest` assertions

jaley:
I realised when looking at #928 for @lawrence-forooghian and #932 for @AndyTWF that the existing `TrackableStateReceiver` design in this test makes it quite complicated to make any assertions relating to expected behaviours during faults, because faults didn't have a way to communicate whether they're fatal, nonfatal (always) or nonfatal once you call the `resolve()` method. This PR replaces the `Fault.stateReceiverForStage()` interface with `Fault.type()`, where faults simply communicate what kind of fault they are. This also gives us an opportunity for the fault to provide sensible timeouts for the nature of the fault implemented.

The existing `TrackableStateReceiver` code has been moved into the test, as it's now just a helper that `NetworkConnectivityTest` uses to monitor for side-effects generally. As such, it's renamed to `PublisherMonitor`, to reflect what it does, but this could possibly be generalised if other tests need a way to monitor for similar side-effects.

I think this change will simplify implementation of the Subscriber test, so putting this in a draft PR and handing the baton to @AndyTWF so that I'm not blocking the team.

Some assumptions I'm slightly unsure of relating to using REST APIs rather than listening for WebSocket events to validate side-effects. This does simplify things, but worth an extra set of eyes to ensure these assumptions are valid:

* I'm using the Ably Presence API to just get a presence snapshot "now" after state transitions, as I believe this is reasonable -- state transitions to `Online` or `Failed` should only happen when we've confirmed our presence status, I think. The trickier case is `Offline`, when there's an active fault, as the Publisher won't necessarily have been able to communicate with Ably. I therefore don't bother asserting presence status for the `Offline` during `NonfatalWhenResolved` fault, as I figure this is basically just testing Ably server-side timeouts, not AAT.
* I'm also using the History API to look up the most-recently published message. Not so sure about this one... It seems OK for the initial call to `track()`, but doesn't work for the secondary `add()` call. I actually think this is because of the client implementation, not specifically a race with the Ably API state. So, I'm not sure it was actually valid to be asserting that the location update sent while adding second trackable be immediately broadcast anyway?

A few other changes to be aware of:
* Cleaned up `withResources` to pass itself as `this` to the test body, cuts down on boiler-plate.
* Moved `waitForStateTransition()` into `PublisherMonitor` (previously `TrackableStateReceiver`) as all the resources it uses are now members of that class.
* Added a `Fault.cleanUp` function with default implementation to just shutdown the proxy. `DisconnectAndSuspend` overrides this to also cancel its async timer.

State at the time of opening:
* Should be somewhat 'green' as-is, all the problematic tests are still disabled.
* I'm seeing `DisconnectAndSuspend` hang on `faultBeforeAddingTrackable`, seemingly during the `setUp()` function. Not sure what's going on here, it may be related to the new cleanup stuff in the last commit.
* Side-effects are being successfully verified. I didn't see `DetachUnresponsive` move to failing as described by @lawrence-forooghian in #928. And initial glance through the logs suggests that the only `DETACH` operation in that test occurs *after* the fault is resolved anyway. I can only assume this is because the `remove()` call completes before the channel is detached? I'm not sure this is a problem, the end result appears to be that all use cases exercised complete successfully, with the expected side-effects, despite the fault?

### DIFF
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -1,14 +1,9 @@
 package com.ably.tracking.test.android.common
 
-import com.ably.tracking.TrackableState
 import io.ktor.websocket.Frame
 import io.ktor.websocket.FrameType
 import java.util.Timer
 import kotlin.concurrent.timerTask
-import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.flow.first
-import kotlinx.coroutines.flow.mapNotNull
-import kotlin.reflect.KClass
 
 /**
  * Abstract interface definition for specific instances of connectivity
@@ -24,6 +19,12 @@ abstract class FaultSimulation {
      * A human-readable name for this type of fault
      */
     abstract val name: String
+
+    /**
+     * The type of fault this simulates - used to validate the state of trackables
+     * and channel activity during and after the fault.
+     */
+    abstract val type: FaultType
 
     /**
      * Subclasses can override this value to `true` in order to skip test that use this fault.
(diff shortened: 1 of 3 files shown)


AndyTWF:
Tried reproducing DisconnectAndSuspend, but no dice :(

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
```diff
@@ -235,12 +267,18 @@ abstract class DropAction(
     private val action: Message.Action
 ) : ApplicationLayerFault(apiKey) {
 
+    // TODO: This fault needs a limit!
```
AndyTWF, line 270: Does it still need a limit?

QuintinWillison: Deferred: https://github.com/ably/ably-asset-tracking-android/issues/934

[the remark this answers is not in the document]
jaley: (Sorry, just realised this was pending)

PR #935, opened 2023-01-23
setting log level for test location source AblyRealtime instance

davyskiba:
Setting `log level` in `LocationSource` ably client to prevent overwriting the `VERBOSE` log level specified in the client under test options, and some messages were not logged. Can be reverted once [underlying ably-java issue](https://github.com/ably/ably-java/issues/901) is fixed.


KacperKluka (approved):
LGTM 👍

PR #936, opened 2023-01-24
Improve testing in respect of JVM version underlying Gradle

QuintinWillison:


QuintinWillison:
Before https://github.com/ably/ably-asset-tracking-android/pull/936/commits/1b20a6b0f5801dd86d7c43de29a66c36b81c64c5, for Gradle running on Java 19, we were [seeing](https://github.com/ably/ably-asset-tracking-android/actions/runs/3998298318/jobs/6860751851#step:4:473):

```
> Task :core-sdk:processDebugUnitTestJavaRes NO-SOURCE
java.lang.instrument.IllegalClassFormatException: Error while instrumenting jdk/proxy2/$Proxy3.

	at org.jacoco.agent.rt.internal_3570298.CoverageTransformer.transform(CoverageTransformer.java:94)
	at java.instrument/java.lang.instrument.ClassFileTransformer.transform(ClassFileTransformer.java:244)
```

and other related problems. That ☝️ Jacoco upgrade appears to have fixed that.

QuintinWillison:
The final fix required to get this PR green was updating the MapBox secrets for this repository, given the ones we were using had expired.

jaley:
@QuintinWillison - do we actually need the JVM bump and TLS workaround? If not, maybe it's worth reverting those to reduce risk as we're approaching release time and don't want other things mysteriously failing at this point?

KacperKluka (approved):
LGTM 👍

QuintinWillison:
@jaley I agree regarding the TLS workaround and I will push a commit for that shortly, but I would rather leave the JVM "bump" in there. It relates purely to the JVM that underlies the Gradle build tools, so I don't expect it to have any effect on the built product that is published to Maven Central. It also ensures that we support developers who are using different local Java environments via the check workflow.

jaley (approved):
Sounds reasonable, thanks @QuintinWillison

davyskiba (approved):

PR #937, opened 2023-01-24
Switch to CIO ktor engine

jaley:
We're seeing random crashes from the Netty backend currently used as a websocket server in the Layer 7 proxy implementation. These seem to be related to [KTOR-2762](https://youtrack.jetbrains.com/issue/KTOR-2762) and [KTOR-3928](https://youtrack.jetbrains.com/issue/KTOR-3928), neither of which seem to have a quick resolution in sight.

Ktor has different backend options, however. CIO is a pure Kotlin, coroutine-based alternative, which aims to be free of heavyweight JVM dependencies. This seems to be more stable for our purposes, testing on Android. Let's see if CI agrees...

### DIFF
--- a/android-test-common/build.gradle
+++ b/android-test-common/build.gradle
@@ -14,7 +14,7 @@ dependencies {
 
     // Ktor dependencies for Layer 7 proxy in NetworkConnectivityTest
     implementation "io.ktor:ktor-server-websockets:$ktor_version"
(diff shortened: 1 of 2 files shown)


AndyTWF (approved):

jaley:
This is green now that the secrets issue is fixed. Probably worth getting merged to save others hassle!

QuintinWillison (approved):

Issue #907, opened 2023-01-13
Non-fatal errors responses for `presence.enter()` cause Publisher to throw exceptions


jaley:
The publisher treats non-fatal errors from `presence.enter()` (which can be retried) as fatal during calls to `track()`. This is reproduced in `faultBeforeAddingTrackable[EnterFailedWithNonfatalNack]` on #886. Initial analysis below:

## faultBeforeAddingTrackable[EnterFailedWithNonfatalNack]

Publisher `track()` call fails because the `presence.enter()` call fails when ably-java calls the
`onError` handler in response to the injected (non-fatal) `NACK`. This results in an exception and
the operation is aborted with a `ConnectionException` thrown to the caller. This is a non-fatal error,
which should result in transparent retries in the background, with the trackable in the Offline state
until it succeeds.

```
01-13 15:15:39.073 19291 20859 D PUBLISHING SDK IT: 19729:  13-01-23 15:15:39.072: [DefaultAbly] Failed to suspend enter presence for channel com.ably.tracking.common.DefaultAblySdkRealtime$Channel@23e3f6d.name com.ably.tracking.ConnectionException: injected by proxy
01-13 15:15:39.075 19291 20048 D PUBLISHING SDK IT: 19318:  13-01-23 15:15:39.074: [DefaultAbly] Failed to enter the presence of channel tracking:58383c6a-0535-48b0-a19f-097b81707b47 com.ably.tracking.ConnectionException: injected by proxy
01-13 15:15:39.076 19291 20048 D PUBLISHING SDK IT: 19318:  13-01-23 15:15:39.075: [DefaultAbly] Failed to connect for channel tracking:58383c6a-0535-48b0-a19f-097b81707b47 com.ably.tracking.ConnectionException: injected by proxy
01-13 15:15:39.079 19291 20048 D PUBLISHING SDK IT: 19318:  13-01-23 15:15:39.078: [DefaultAbly] Failed to connect for trackable 58383c6a-0535-48b0-a19f-097b81707b47 com.ably.tracking.ConnectionException: injected by proxy
01-13 15:15:39.082 19291 19398 D PUBLISHING SDK IT: 19040:  13-01-23 15:15:39.081: [DefaultPublisher] Publisher track operation failed com.ably.tracking.ConnectionException: injected by proxy
[...]

PR #938, opened 2023-01-25
871 add exceptions

### LINKED ISSUE #871: Exceptions thrown for new publishing requests when offline
Currently, the main public Publisher interfaces for tracking new Trackables (`Publisher.track()`, `Publisher.add()`, also `Publisher.remove()`) all throw `ConnectionException` in the event that we're unable to connect to Ably at the instant those functions are called. We discussed this internally and agreed that this isn't the most helpful design. Summary from internal discussion:

* Generally, we handle being offline for long periods. Trackables simply move to the `Offline` state if we're unable to publish updates.
* It seems arbitrary that, if connectivity failed 1ms after a call to `track()`, there'd be no errors for the client to handle - the Trackable moves to the `Offline` state as designed, but if the call arrives that tiny bit later, we require the client to set up their own background retry loop, effectively duplicating AAT functionality.

Consequently, agreed interpretations of trackable states are:

* `Online`: connected and present;
* `Offline`: not online, but for a non-fatal reason (eg not yet connected, or retriable error attempting to attach or enter);
* `Failed`: the connection or channel are in the failed state, or presence enter failed with a non-retriable error

If any of the operations required to add and track failed for a non-retriable reason, then it's `Failed`. If all of the operations required are in one of the following states:

* still in progress; or
* succeeded; or
* failed for a retriable reason

then it's `Offline`.

davyskiba:
This PR resolves connectivity issues that can occur during add/track calls. When `channel.attach()` or `presence.enter()` fails with a non-fatal exception, adding the trackable will continue. Ably-java will restore the channel and connection, and entering presence will be retried by AAT. If a retry attempt fails with a fatal exception, the trackable will permanently enter the failed state, and the corresponding channel will be removed. 
Additionally, updated trackable state logic to go to `Offline` state if presence has not been entered. 

Disabling publishing location updates for trackables until presence enter succeeds is tracked separately [under this issue.](https://github.com/ably/ably-asset-tracking-android/issues/953).

Resolves #871 and #907


jaley (changes requested):
I've checked this out and run tests locally. I'm getting the same failures as CI:

```
com.ably.tracking.publisher.NetworkConnectivityTests > faultBeforeAddingTrackable[TcpConnectionRefused][test(AVD) - 8.1.0] FAILED 
	java.lang.AssertionError: [fault active] publisher.track() timed out.
	at com.ably.tracking.publisher.PublisherMonitor$waitForStateTransition$1.invokeSuspend(NetworkConnectivityTests.kt:649)

com.ably.tracking.publisher.NetworkConnectivityTests > faultDuringTracking[TcpConnectionUnresponsive][test(AVD) - 8.1.0] FAILED 
	java.lang.AssertionError: Expected publisherPresent: true but got false
	at com.ably.tracking.publisher.PublisherMonitor.assertPresence(NetworkConnectivityTests.kt:701)

com.ably.tracking.publisher.NetworkConnectivityTests > faultBeforeAddingTrackable[TcpConnectionUnresponsive][test(AVD) - 8.1.0] FAILED 
	java.lang.AssertionError: [fault active] publisher.track() did not result in success.
	at com.ably.tracking.publisher.PublisherMonitor$waitForStateTransition$1.invokeSuspend(NetworkConnectivityTests.kt:652)

com.ably.tracking.publisher.NetworkConnectivityTests > faultBeforeAddingTrackable[AttachUnresponsive][test(AVD) - 8.1.0] FAILED 
	java.lang.AssertionError: Expectation '[fault active] publisher.track()' did not result in success.
	at com.ably.tracking.publisher.PublisherMonitor.assertStateTransition(NetworkConnectivityTests.kt:665)
```

jaley (changes requested):
The `track()` and `add` operations are still blocking in this PR. The failures we're seeing from `faultBeforeAddingTrackable()` are because `track()` doesn't return, with the `Trackable` in the `Offline` state as expected. There's no-longer a timeout at the top-level, so it blocks indefinitely while the connection is interrupted, until the test times it out.

This needs a change similar to what @KacperKluka did with `remove()`, making us quickly return with the `StateFlow` in `Offline` state, then updating to `Online` once we successfully connect, attach and enter presence.

I was also surprised to see that we're not using the `retryChannelOperationIfConnectionResumeFails()` for the `presence.enter()` calls, was this intentional? That function seems to be the only thing that checks if the channel is in `suspended` state, and calls to `presence.enter()` will immediately fail (with a fatal error) if the channel is suspended.

KacperKluka:
@jaley you wrote that
> I was also surprised to see that we're not using the `retryChannelOperationIfConnectionResumeFails()` for the `presence.enter()` calls, was this intentional? That function seems to be the only thing that checks if the channel is in `suspended` state, and calls to `presence.enter()` will immediately fail (with a fatal error) if the channel is suspended.

In the meantime, one of things I'm aiming to do in https://github.com/ably/ably-asset-tracking-android/issues/948 is [this](https://github.com/ably/ably-asset-tracking-android/issues/948#issuecomment-1406670324):
> Remove `DefaultAbly.retryChannelOperationIfConnectionResumeFails()` as now we should be retrying all retriable errors but outside of the Ably wrapper class

Does it mean that the `retryChannelOperationIfConnectionResumeFails` method should stay and be used after all? 🤔

jaley:
> Does it mean that the retryChannelOperationIfConnectionResumeFails method should stay

@KacperKluka sorry no, not necessarily. So long as there's some other way we do the appropriate retries. I mentioned it here because I was seeing errors in the logs when testing this branch about us trying to enter presence on a suspended channel. If your change has a different way to handle that, that's great.

KacperKluka (approved):
I've skimmed over some parts (mostly tests) but overall it looks good, especially that the race condition fix PR https://github.com/ably/ably-asset-tracking-android/pull/963 will be merged into this 👍

jaley (approved):
LGTM - I'm assuming the plan is to merge #963 into this, then merge to `main`

AndyTWF (approved):

Issue #905, opened 2023-01-13
`Publisher.remove()` fails during several connectivity faults


jaley:
Relates to #871, being largely addressed on #883, but adding a new ticket as there are a couple of separate underlying causes that can manifest as the same `ConnectionException` throwing from `remove()`:

## faultDuringTracking[TcpConnectionRefused]

Connection gets interrupted with two trackables in Online state. Both trackables move to the Offline state as expected.
`faultDuringTracking[TcpConnectionRefused]` then attempts to remove that second trackable. This eventually fails because
`DefaultAbly.leavePresence$onError` is called when the presence leave operation times out.

```
01-13 15:10:14.550 19291 19398 D PUBLISHING SDK IT: 19040:  13-01-23 15:10:14.548: [AblyConnectionStateChangeWorker] Ably connection state changed OFFLINE
01-13 15:10:14.552 19291 19319 D PUBLISHING SDK IT: 19007:  13-01-23 15:10:14.551: [DefaultPublisher] Publisher remove operation started
01-13 15:10:14.554 19291 19766 D PUBLISHING SDK IT: THREAD 19245 is 'WebSocketConnectReadThread-19245' [no Looper]
01-13 15:10:14.554 19291 19766 D PUBLISHING SDK IT: 19245:  Connection error  java.net.ConnectException: failed to connect to localhost/127.0.0.1 (port 13579) from /:: (port 49834): connect failed: ECONNREFUSED (Connection refused)
01-13 15:10:14.667 19291 19398 D PUBLISHING SDK IT: 19040:  13-01-23 15:10:14.665: [AblyConnectionStateChangeWorker] Ably connection state changed OFFLINE

01-13 15:10:44.828 19291 19798 D PUBLISHING SDK IT: 19249:  Connection error  java.net.ConnectException: failed to connect to localhost/127.0.0.1 (port 13579) from /:: (port 43456): connect failed: ECONNREFUSED (Connection refused)

(for 2 mins)

01-13 15:12:15.896 19291 19723 D PUBLISHING SDK IT: 19219:  13-01-23 15:12:15.894: [DefaultAbly] Failed to leave presence for channel tracking:6950bb58-3d79-4cf9-bf9e-f70649923b86 com.ably.tracking.ConnectionException: failed to connect to localhost/127.0.0.1 (port 13579) from /:: (port 44444): connect failed: ECONNREFUSED (Connection refused)
[...]

PR #942, opened 2023-01-26
Change publisher's remove operation to always succeed

KacperKluka:
Changed the `ably.disconnect()` method:
- it takes maximum 30 seconds
- if the `presence.leave()` fails for a retriable reason it is being retried in 15 seconds
- if the `presence.leave()` fails for a non-retriable reason it is not retried
- after the `presence.leave()` the channel is removed

Thanks to this the `disconnect()` method shouldn't fail, therefore, the whole `Publisher.remove()` operation shouldn't have a reason to fail as well. Additionally, the `remove()` now succeeds immediately and switches trackable state to `Offline` and triggers the `disconnect()` operation afterwards.

**Note:**
This fixes the `faultDuringTracking[TcpConnectionRefused]` and `faultDuringTracking[TcpConnectionUnresponsive]` mentioned in the https://github.com/ably/ably-asset-tracking-android/issues/905 issue. However, the `faultBeforeAddingTrackable` test still fails for those faults. Because of that, I'm not able to enable those two faults in `NetworkConnectivityTests`.

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -497,14 +497,13 @@ constructor(
         try {
             leavePresence(channel, presenceData)
         } catch (connectionException: ConnectionException) {
-            if (connectionException.isRetriable()) {
-                logHandler?.w("$TAG Failed to leave presence for channel ${channel.name} due to a retriable exception, the operation will be retried in $PRESENCE_LEAVE_RETRY_DELAY_IN_MILLISECONDS ms", connectionException)
-                delay(PRESENCE_LEAVE_RETRY_DELAY_IN_MILLISECONDS)
-                leavePresenceRepeating(channel, presenceData)
-            } else {
+            if (!connectionException.isRetriable()) {
                 logHandler?.w("$TAG Failed to leave presence for channel ${channel.name} due to a non-retriable exception", connectionException)
                 throw connectionException
             }
+            logHandler?.w("$TAG Failed to leave presence for channel ${channel.name} due to a retriable exception, the operation will be retried in $PRESENCE_LEAVE_RETRY_DELAY_IN_MILLISECONDS ms", connectionException)
+            delay(PRESENCE_LEAVE_RETRY_DELAY_IN_MILLISECONDS)
+            leavePresenceRepeating(channel, presenceData)
         }
     }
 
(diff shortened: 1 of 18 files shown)


jaley:
> However, the faultBeforeAddingTrackable test still fails for those faults. Because of that, I'm not able to enable those two faults in NetworkConnectivityTests.

Would that be because the `publisher.track()` fix is on @davyskiba's PR? If so, I guess we just need to make sure we merge on of these PRs, then catch up the other to enable the tests.

KacperKluka:
@jaley 
> Would that be because the publisher.track() fix is on @davyskiba's PR?

Yes, they fail due to errors during the `add()`/`track()` operation and they should be gone after Wojtek's PR is ready 😉

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -493,19 +464,46 @@ constructor(
...
-                }
+    private suspend fun leavePresenceRepeating(channel: AblySdkRealtime.Channel<ChannelStateListenerType>, presenceData: PresenceData) {
+        try {
+            leavePresence(channel, presenceData)
+        } catch (connectionException: ConnectionException) {
+            if (connectionException.isRetriable()) {
+                logHandler?.w("$TAG Failed to leave presence for channel ${channel.name} due to a retriable exception, the operation will be retried in $PRESENCE_LEAVE_RETRY_DELAY_IN_MILLISECONDS ms", connectionException)
+                delay(PRESENCE_LEAVE_RETRY_DELAY_IN_MILLISECONDS)
+                leavePresenceRepeating(channel, presenceData)
+            } else {
```
AndyTWF, line 504: Just a stylistic thought, not super important - I'd probably use a `return` statement here and remove the `else`. Avoids code indentation and better indicates which path is the "happy" path :)

AndyTWF (approved):

common/src/main/java/com/ably/tracking/common/Ably.kt
KacperKluka: That's a very good suggestion 👍 Though, I inverted your request a bit and put the error path inside an `if` (since it already `throw`s and returns from the method) while keeping the happy path on the "main lane" 😅 WDYT? 😉 https://github.com/ably/ably-asset-tracking-android/pull/942/commits/ae2b7f587d02e9f39407c0bf760eb656e62d2ef8

jaley (approved):
Looks good - thanks @KacperKluka

common/src/main/java/com/ably/tracking/common/Ably.kt
AndyTWF: Beautiful :D

AndyTWF (approved):
Approving after the stylistic change

PR #944, opened 2023-01-26
Fix flakey publisher test

### LINKED ISSUE #943: Flakey test: com.ably.tracking.publisher.NetworkConnectivityTests > faultDuringTracking[NullTransportFault]
```
com.ably.tracking.publisher.NetworkConnectivityTests > faultDuringTracking[NullTransportFault][test(AVD) - 10] FAILED 
	java.lang.AssertionError: Expected location update LocationMessage(type=Feature, geometry=LocationGeometry(type=Point, coordinates=[102.0, 82.0, 0.0]), properties=LocationProperties(accuracyHorizontal=5.0, bearing=0.0, speed=5.0, time=1.674684137613E9)) but last was LocationMessage(type=Feature, geometry=LocationGeometry(type=Point, coordinates=[103.0, 83.0, 0.0]), properties=LocationProperties(accuracyHorizontal=5.0, bearing=0.0, speed=5.0, time=1.674684139815E9))
	at com.ably.tracking.publisher.PublisherMonitor.assertLocationUpdated(NetworkConnectivityTests.kt:734)
```

First seen on https://github.com/ably/ably-asset-tracking-android/pull/940

### LINKED ISSUE #946: Reduce publisher location check polling interval
After https://github.com/ably/ably-asset-tracking-android/pull/944 is merged, reduce the polling delay for location updates to 50ms

AndyTWF:
This addresses two aspects of flakeyness in `com.ably.tracking.publisher.NetworkConnectivityTests > faultDuringTracking`

1. Locations in this test are fed to mapbox via an Ably channel and there is almost certainly a delay between the update landing on the Ably channel, and mapbox turning it into a location update that is published on the trackables channel by the publisher. This occasionally leads to issues where we receive a stale update when querying channel history for location updates. The fix here is to to allow a grace and back-off period when checking the trackables channel for location updates, to allow time for the update to come through.

2. Occasionally, we would see location updates coming through from a different test. I believe this is because the tests share the same Ably channel for their mapbox location source. This is fixed by ensuring that every test uses a unique channel name.

Fixes #943
Fixes #946

### DIFF
--- a/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
+++ b/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
@@ -457,7 +457,7 @@ class LocationHelper {
     private val opts = CLIENT_OPTS_NO_PROXY
     private val ably = AblyRealtime(opts)
 
-    val channelName = "testLocations"
+    val channelName = "publisherIntergrationTestLocations-" + UUID.randomUUID()
     private val channel = ably.channels.get(channelName)
 
     private val gson = Gson()
--- a/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
+++ b/publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
@@ -57,6 +57,7 @@ import io.ably.lib.util.Log
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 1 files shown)


jaley (approved):
Thanks, Andy! This should make the tests a bit more stable.

publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
```diff
@@ -729,36 +731,39 @@ class PublisherMonitor(
...
+                                EnhancedLocationUpdateMessage::class.java
+                            ).location
+
+                            if (latestLocation.equalGeometry(expectedLocationUpdate)) {
+                                testLogD("PublisherMonitor: $label - (PASS) lastPublishedLocation = $latestLocation")
+                                return@withTimeout
+                            }
+                        }
+
+                        delay(500)
```
QuintinWillison, line 758: I'm not sure how often `assertLocationUpdate()` is called, but I am assuming this polling interval is arbitrary. Perhaps `500` (presumably milliseconds?) could be shorter, as these could add up quickly. Still arbitrary, but even 20Hz wouldn't feel too high, so perhaps `50` ms?

QuintinWillison (approved):
I recognise that we want and need to get this PR merged ASAP, so I would not address this within current scope, but if you agree then perhaps consider creating an issue to look at it later.

publishing-sdk/src/androidTest/java/com/ably/tracking/publisher/NetworkConnectivityTests.kt
AndyTWF: As a check was going to fail for an unrelated flakey test, have done in [a84fc67](https://github.com/ably/ably-asset-tracking-android/pull/944/commits/a84fc675f0d26dbaa060e7ea16bce88079a4250d)

PR #945, opened 2023-01-26
Change publisher and subscriber stop operation to always succeed

KacperKluka:
Previously, publisher and subscriber `stop()` method could throw an `ConnectionException`. By a lucky accident it was changed when we introduced the `Ably.stopConnection()` method (a few moths ago) which caught the exception and wrapped it within a `Result`. Now we made it clear that the `stop()` method won't throw by removing the `@throws` information from its code docs.

Additionally, the `stopConnection()` method will now wait for the connection to close only for 30 seconds. After that time passes it will log that a timeout happened and carry on.


jaley (approved):
LGTM!

AndyTWF (approved):

Issue #896, opened 2023-01-12
Adding a trackable stalls forever if presence.enter() is interrupted by a disconnection


paddybyers:
This behaviour has been observed during analysis of integration test failures (see https://github.com/ably/ably-asset-tracking-android/issues/865).

The underlying cause is https://github.com/ably/ably-java/issues/474.

Related: https://github.com/ably/ably-asset-tracking-android/issues/862

AndyTWF:
This possibly may be the underlying cause of the timeout issues in the flakey test at https://github.com/ably/ably-asset-tracking-android/issues/846 ?

jaley:
@ikbalkaya - I believe your incoming PR to update `ably-java` will resolve this issue

ikbalkaya:
@jaley I just had a run with tests against new changes. I'd assume that test for this case would be faultDuringTracking[DisconnectWithFailedResume] If that's the case it fixes the issue, I'll reeanble the test and create a PR on this

PR #947, opened 2023-01-26
Reenable disconnect failed resume faults

ikbalkaya:
This PR does three things 
* Increases ably-android version to 1.2.23
* Reenables skipped tests for `DisconnectFailedWithResume`
* Reenables skipped tests for `EnterUnresponsive`
Closes https://github.com/ably/ably-asset-tracking-android/issues/896

### DIFF
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -499,11 +499,6 @@ class DisconnectWithFailedResume(apiKey: String) : ApplicationLayerFault(apiKey)
         }
     }
 
-    /*
-        Currently failing due to ably-java#474 presence bug
-     */
-    override val skipTest = false
-
     /**
      * State of the fault, used to control whether we're intercepting
      * the connection or looking to inject a WebSocket CLOSE at an appropriate time
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -474,12 +474,6 @@ class EnterUnresponsive(apiKey: String) : UnresponsiveAfterAction(
         }
     }
 
-    /*
-        This test currently fails because the ably-java hangs the client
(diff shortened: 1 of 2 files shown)


android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
```diff
@@ -502,7 +502,7 @@ class DisconnectWithFailedResume(apiKey: String) : ApplicationLayerFault(apiKey)
     /*
         Currently failing due to ably-java#474 presence bug
      */
-    override val skipTest = true
+    override val skipTest = false
```
AndyTWF, line 505: The default value of `skipTest` is false, so we should just be able to remove this completely :)

AndyTWF (changes requested):

jaley:
From my testing @ikbalkaya, it should also be safe to reenable `EnterUnresponsive`?

ikbalkaya:
> From my testing @ikbalkaya, it should also be safe to reenable `EnterUnresponsive`?

@AndyTWF In my local, only one passes unfortunately. The one with `faultBeforeAddingTrackable` seems to fail

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
ikbalkaya: `skipTest` now should have ceased to exist in this fault d1badac02ee6c847310b9e89c223cddec7b27387

AndyTWF (approved):

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
AndyTWF: 👍🏻

jaley:
OK, no worries. Let's leave it disabled if you're seeing problems and just prioritise getting the ably-java bump landed

ikbalkaya:
@jaley I realized I had this test run without changes also from main - Now I can confirm that both test pass and I reenabled the EnterUnresponsive fault as well now

jaley (approved):
LGTM, thanks!

QuintinWillison (approved):

PR #950, opened 2023-01-26
Get better information about flakey test

AndyTWF:
Adding some extra logging to try and work out why a test is failing.


AndyTWF:
I have the information now.

PR #952, opened 2023-01-26
Fix `publish.stop()` hang when no trackables added

jaley:
In `main`, the following works:

```
val publisher = createPublisher()
publisher.track(Trackable("foo"))
publisher.stop()
```

However, the following will hang indefinitely when `stop()` is called:

```
val publisher = createPublisher()
publisher.stop()
```

This is because `ably-java` will not deliver the state change callbacks expected to complete the `stop()` call until we've called `connect()`, which only happens during a `track()` operation. We have special cases to handle `failed` and `closed` states, but were not checking for the `initialized` state.

This was find by adding a `NetworkConnectivityTest` case that stops the publisher during a fault and restarts it. The fault was actually not necessary, this problem happens regardless of any active faults. The new test case passes for all existing fault implementations, with the fix applied.

(Expecting some issues with CI due to ongoing flaky test issues... let's see!)


QuintinWillison (approved):
Nice find. 💪

AndyTWF (changes requested):
Changes look good - just a few linting issues

KacperKluka (approved):
LGTM, great that you've spotted this problem 👍

AndyTWF (approved):

PR #960, opened 2023-01-27
`NetworkingConnectivityTest` fixes

### LINKED ISSUE #934: Investigate TODO comment in `DropAction` fault regarding "limit"
As commented on [here](https://github.com/ably/ably-asset-tracking-android/pull/933/files#r1083957234) by @AndyTWF.

The comment might be inert or otherwise unimportant, but in the interests of getting the aforementioned pull request merged adding this issue to defer looking at this until later.

jaley:
Two bugs in this test:

(1) We never resolve the fault in `faultBeforeAddingTrackable()`! This wasn't obvious previously because the failures we've seen here come from the earlier call to `track()` throwing an exception while the fault is still active, but this fix will be needed for #938 

(2) We're asserting that a location update should have somehow happened through all active faults. This isn't possible, because the proxy will be blocking the outgoing messages, so we should only assert that the location update has happened during a non-fatal (auto-resolving) fault, or after the fault is resolved (already implemented)

Edit: added one more:

(3) The `TODO` in `DropAction` was also problematic, as the `FaultType` was set to `Nonfatal`, which is correct, but only if the issue self-heals. The fault simulation is unrealistic if permanent, because we can't really have a situation where certain message types are being delivered over WS but not others. Included a fix for this. Resolves #934

### DIFF
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -485,6 +485,11 @@ class EnterUnresponsive(apiKey: String) : UnresponsiveAfterAction(
  */
 class DisconnectWithFailedResume(apiKey: String) : ApplicationLayerFault(apiKey) {
 
+    /*
+     * This test fails intermittently. Investigation ongoing in #951
+     */
+    override val skipTest = true
+
     companion object {
         val fault = object : Fault() {
             override fun simulate(apiKey: String) = DisconnectWithFailedResume(apiKey)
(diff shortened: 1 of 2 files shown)


AndyTWF (approved):

davyskiba (approved):

jaley:
I think we optimistically may have enabled `DisconnectWithFailedResume` prematurely when we merged ably/ably-java#900 as that flakes a lot right now, making it impossible to get a green build. I've pushed an update to disable until @AndyTWF and I are done investigating that, hopefully it'll unblock unrelated changes.

PR #963, opened 2023-01-28
Fix race conditions in presence retry

AndyTWF:
This change fixes three issues.

1. A race condition on retrying presence enter.

At the moment the work to RetryEnterPresence is queued by AddTrackableWorker. RetryEnterPresenceWorker relies on the trackable being in the publishers trackables list. The trackable is added to this list by ConnectionReadyWorker, which is itself queued a few workers down the connection chan from AddTrackableWorker. The race condition is that in some conditions, especially where flakey connections are involved, the RetryEnterPresenceWorker can run before ConnectionReadyWorker, meaning that the trackable will never enter presence, and therefore never come online (RetryEnterPresenceWorker does not queue a re-attempt if the trackable isn't known).

This change fixes the issue by moving the queuing of the RetryEnterPresence work into ConnectionReadyWorker, which ensures that the trackable is in the publisher properties prior to execution.

2. Trackables entering a failed state

RetryEnterPresenceWorker transitions trackables into the failed state if a fatal error occurs whilst entering presence.
This can happen if the connection re-connects but the channel is not yet attached.

This change fixes the issue by introducing a listener on the ably channel state, which suspends the coroutine until
the channel notifies us of an online state. Thus ensuring that we don't try to enter presence until it is possible to
do so.

3. Do not retry enter presence when trackable marked for removal

Add a check that prevents RetryEnterPresenceWorker from being run if
the trackable has been marked for removal.

There is a test case for this already, but it was a false positive pass - as the trackable
had not been added to the publisher properties, so the worker didn't run anyway. The test case has
been amended to check the correct behaviour.

### DIFF
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -158,7 +158,7 @@ class TcpConnectionRefused(apiKey: String) : TransportLayerFault(apiKey) {
     }
 
     override val type = FaultType.NonfatalWhenResolved(
-        offlineWithinMillis = 30_000,
+        offlineWithinMillis = 60_000,
         onlineWithinMillis = 60_000
     )
 
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -165,8 +165,7 @@ class TcpConnectionRefused(apiKey: String) : TransportLayerFault(apiKey) {
     /**
      * This fault type is temporarily disabled at runtime. It can be re-enabled by removing this override.
      * We will re-enable this test when the following have been addressed:
-     * - https://github.com/ably/ably-asset-tracking-android/issues/859
-     * - https://github.com/ably/ably-asset-tracking-android/issues/871
+     * - https://github.com/ably/ably-asset-tracking-android/issues/948
      */
     override val skipTest = true
 
@@ -200,8 +199,7 @@ class TcpConnectionUnresponsive(apiKey: String) : TransportLayerFault(apiKey) {
     /**
      * This fault type is temporarily disabled at runtime. It can be re-enabled by removing this override.
      * We will re-enable this test when the following have been addressed:
-     * - https://github.com/ably/ably-asset-tracking-android/issues/859
-     * - https://github.com/ably/ably-asset-tracking-android/issues/871
+     * - https://github.com/ably/ably-asset-tracking-android/issues/948
      */
     override val skipTest = true
 
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -162,13 +162,6 @@ class TcpConnectionRefused(apiKey: String) : TransportLayerFault(apiKey) {
         onlineWithinMillis = 60_000
     )
 
-    /**
-     * This fault type is temporarily disabled at runtime. It can be re-enabled by removing this override.
-     * We will re-enable this test when the following have been addressed:
-     * - https://github.com/ably/ably-asset-tracking-android/issues/948
-     */
-    override val skipTest = true
-
     override fun enable() {
         tcpProxy.stop()
     }
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -196,13 +196,6 @@ class TcpConnectionUnresponsive(apiKey: String) : TransportLayerFault(apiKey) {
         onlineWithinMillis = 60_000
     )
 
-    /**
-     * This fault type is temporarily disabled at runtime. It can be re-enabled by removing this override.
-     * We will re-enable this test when the following have been addressed:
-     * - https://github.com/ably/ably-asset-tracking-android/issues/948
-     */
-    override val skipTest = true
-
     override fun enable() {
         tcpProxy.isForwarding = false
     }
--- a/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
+++ b/android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
@@ -158,7 +158,7 @@ class TcpConnectionRefused(apiKey: String) : TransportLayerFault(apiKey) {
     }
 
     override val type = FaultType.NonfatalWhenResolved(
-        offlineWithinMillis = 60_000,
+        offlineWithinMillis = 30_000,
         onlineWithinMillis = 60_000
     )
 
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -904,9 +904,11 @@ constructor(
      *
      * TODO: At the moment, due to how listeners are wrapped in AAT we're not able to do this the elegant way
      * without rewriting a lot of code. Not ignoring subsequent updates breaks continuations. To get around this for
-     * now - ignore any future event updates.
+     * now - ignore any future event updates. We should change this in the future. There shouldn't be a risk of overflows
+     * as this is at a channel level - and the channels are retired after a trackable is done with.
      */
     override suspend fun waitForChannelToAttach(trackableId: String): Result<Unit> {
+
         return suspendCoroutine { continuation ->
             var resumed = false
             subscribeForChannelStateChange(trackableId) {
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -909,7 +909,7 @@ constructor(
      */
     override suspend fun waitForChannelToAttach(trackableId: String): Result<Unit> {
 
-        return suspendCoroutine { continuation ->
+        return suspendCancellableCoroutine { continuation ->
             var resumed = false
             subscribeForChannelStateChange(trackableId) {
                 if (resumed) {
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -922,7 +922,9 @@ constructor(
                 }
                 if (it.state == com.ably.tracking.common.ConnectionState.FAILED) {
                     resumed = true
-                    continuation.resume(Result.failure(Exception("Channel failed")))
+                    val errorMessage = "Channel failed whilst waiting for attach: ${it.errorInformation?.message}"
+                    logHandler?.w(errorMessage)
+                    continuation.resume(Result.failure(ConnectionException(ErrorInformation(errorMessage))))
                 }
             }
         }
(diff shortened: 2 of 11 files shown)


publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
```diff
@@ -68,16 +70,25 @@ internal class ConnectionCreatedWorker(
                         isSubscribedToPresence = false
                     )
                 )
+            } catch (exception: TimeoutException) {
```
KacperKluka, line 73: AFAIK the `withTimeout` throws a `TimeoutCancellationException`, so this catch block might not work 🤔

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionReadyWorker.kt
```diff
@@ -53,6 +53,10 @@ internal class ConnectionReadyWorker(
             postWork(WorkerSpecification.RetrySubscribeToPresence(trackable, presenceUpdateListener))
         }
 
+        if (properties.trackableEnteredPresenceFlags[trackable.id] == null) {
```
KacperKluka, line 56: What if it is equal to `false`? Shouldn't it also trigger the retry mechanism? 🤔

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/RetryEnterPresenceWorker.kt
```diff
@@ -59,6 +75,19 @@ internal class RetryEnterPresenceWorker(
...
+            ably.subscribeForChannelStateChange(trackable.id) {
+                if (it.state == ConnectionState.ONLINE) {
+                    continuation.resume(Result.success(Unit))
+                }
+                if (it.state == ConnectionState.FAILED) {
+                    continuation.resume(Result.failure(ConnectionException(it.errorInformation!!)))
+                }
+            }
+        }
+    }
```
KacperKluka, line 89: If the `continuation.resume()` is called twice it will throw an exception, so we should unregister the listener once we call the `resume()` 😉

paddybyers:
> RetryEnterPresenceWorker transitions trackables into the failed state if a fatal error occurs whilst entering presence.
This can happen if the connection re-connects but the channel is not yet attached.

Currently ably-java responds with a 400 error when various operations are attempted when a connection of channel is in the `suspended` state - I don't think this is valid (although the spec doesn't explicitly specify which error should be returned). This means that you're getting this response that looks like a nonretriable error - it's not immediately retriable (all the time that the connection/channel is `suspended`, you're going to get the same response) - but OTOH it's not fatal, because we will continue to retry the connection, and after connection establishment the operation can be reattempted.

So the solution is for AAT to be more aware of connection and channel state, and only schedule retry operations once it knows that the connection/channel state makes success possible.

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionReadyWorker.kt
AndyTWF: Agreed, I had done this but didn't push it - it's pushed now!

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
AndyTWF: Changed!

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
```diff
@@ -158,17 +158,10 @@ class TcpConnectionRefused(apiKey: String) : TransportLayerFault(apiKey) {
     }
 
     override val type = FaultType.NonfatalWhenResolved(
-        offlineWithinMillis = 30_000,
+        offlineWithinMillis = 60_000,
```
jaley, line 161: If we need a lengthy time-out here to get things stable today, that's fine, but IMHO we should consider what this actually means and perhaps raise a ticket to improve later. This timeout represents the maximum amount of time that we think a caller of `publisher.track()` should reasonably have to wait during connectivity issues, before we return control to the app and give them a `StateFlow<TrackableState>` (initially `Offline`) to monitor.

How do we think this will translate in terms of user experience? As an app developer, using this API, one might see that it's a async suspending function returning a `StateFlow` to monitor ongoing transitions and directly connect the UI updates to that `StateFlow`, no? I'm concerned that basically the app will appear unresponsive, or at least the most recent request to add a trackable will appear ignored if it can really take a minute before `track()` gets its `StateFlow` back?

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/RetryEnterPresenceWorker.kt
AndyTWF: Discussed this with Kacper this morning:

TLDR, it's not easy to do "properly" using ably.channel.off() due to AAT wrappers around the listener and the nature of the listener calls (we send events to the listener before it's .on() the channel). In the interest of time and deadlines, we agreed to introduce a local way of preventing duplicate events and raise a separate issue (https://github.com/ably/ably-asset-tracking-android/issues/964) to solve the wider issue.

jaley (approved):
Nice work, thanks @AndyTWF!

I'm a bit concerned about the time to go offline timeout needing to be so high -- see comment below. Something we can resolve in a subsequent ticket if need though.

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
AndyTWF: I agree - now I've made further changes, I will revert this back to 30s again and see if that's any more stable :)

davyskiba (commented):
Awesome that you've found that race <3

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
```diff
@@ -118,14 +118,11 @@ internal class AddTrackableWorker(
                     isConnectedToAbly = true
                 )
             )
-        } else {
-            val enteredPresence = connectResult.isSuccess
-            postWork(createConnectionCreatedWorker(enteredPresence))
-            if (!enteredPresence) {
-                delay(PRESENCE_ENTER_DELAY_IN_MILLISECONDS)
```
davyskiba, line 125: Please remove now unused `PRESENCE_ENTER_DELAY_IN_MILLISECONDS` constant

davyskiba (approved):

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/AddTrackableWorker.kt
AndyTWF: Done in [9a72a87](https://github.com/ably/ably-asset-tracking-android/pull/963/commits/9a72a8729ba9e774f8d31b4f0c358ff749bef901)

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ConnectionCreatedWorker.kt
```diff
@@ -68,16 +70,25 @@ internal class ConnectionCreatedWorker(
...
         }
 
         return properties
     }
 
     private suspend fun subscribeToPresenceMessages(): Result<Unit> {
-        return suspendCoroutine { continuation ->
-            ably.subscribeForPresenceMessages(trackable.id, presenceUpdateListener) { result ->
-                continuation.resume(result)
+        return withTimeout(5000) {
```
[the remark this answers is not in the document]
AndyTWF: Done in [e6719fd](https://github.com/ably/ably-asset-tracking-android/pull/963/commits/e6719fd8545ae92d5f0684b787869effcad1a644)

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/RetryEnterPresenceWorker.kt
```diff
@@ -59,6 +75,27 @@ internal class RetryEnterPresenceWorker(
         }
     }
 
+    private suspend fun waitForChannelToBeConnected(trackable: Trackable): Result<Unit> {
```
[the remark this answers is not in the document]
AndyTWF: Done in [c38d6d2](https://github.com/ably/ably-asset-tracking-android/pull/963/commits/c38d6d245d23c5e4cbbfb8a17302d6aead54750f)

android-test-common/src/main/java/com/ably/tracking/test/android/common/Faults.kt
AndyTWF: Investigating making this instant in https://github.com/ably/ably-asset-tracking-android/issues/966

KacperKluka (changes requested):
Looks great, just a small suggestion 😉

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -892,6 +898,36 @@ constructor(
...
+     * Ably channel-terms means ATTACHED).
+     *
+     * TODO: At the moment, due to how listeners are wrapped in AAT we're not able to do this the elegant way
+     * without rewriting a lot of code. Not ignoring subsequent updates breaks continuations. To get around this for
+     * now - ignore any future event updates. We should change this in the future. There shouldn't be a risk of overflows
+     * as this is at a channel level - and the channels are retired after a trackable is done with.
+     */
+    override suspend fun waitForChannelToAttach(trackableId: String): Result<Unit> {
+
+        return suspendCoroutine { continuation ->
```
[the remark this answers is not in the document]
AndyTWF: 01298df4

KacperKluka (approved):
LGTM 👍

common/src/main/java/com/ably/tracking/common/Ably.kt
```diff
@@ -892,6 +898,36 @@ constructor(
...
+            var resumed = false
+            subscribeForChannelStateChange(trackableId) {
+                if (resumed) {
+                    return@subscribeForChannelStateChange
+                }
```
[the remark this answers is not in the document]
AndyTWF: I'm not sure that this will work - in theory, we could get another connection change event before the continuation actually resumes and thus have duplicates? That's what people suggested on https://stackoverflow.com/questions/48227346/kotlin-coroutine-throws-java-lang-illegalstateexception-already-resumed-but-go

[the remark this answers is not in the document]
KacperKluka: Andy mentioned that the `isActive` might be dangerous in use so IMO we can stick to the local variable if that feels safer 😁 @AndyTWF feel free to resolve this conversation and either leave the local variable or use the `isActive` 👍

AndyTWF:
We're having some CI weirdness at the moment, but this branch is supposed to be going out with its target, so will merge this down and continue the quest in base.

Issue #948, opened 2023-01-26
Remove `connect()` and `attach()` timeouts from `DefaultAbly`


jaley:
Currently, AAT proactively times out `ably-java` `connect()` and `attach()` operations. This is counterproductive, as `ably-java` has an internal retry mechanism for these functions, which we don't benefit from if we time them out early.

The fix should be to remove the following timeouts:
* [attach](https://github.com/ably/ably-asset-tracking-android/blob/main/common/src/main/java/com/ably/tracking/common/Ably.kt#L880)
* [connect](https://github.com/ably/ably-asset-tracking-android/blob/main/common/src/main/java/com/ably/tracking/common/Ably.kt#L1010)

... and then ensure that nothing else is broken by this change.

We must still observe the callback listener to ensure that we transition to the `Failed` state on terminal errors, however.

This should fix the following failing `NetworkConnectivityTest` cases in `main`:
* `faulBeforeAddingTrackable[AttachUnresponsive]`
* `faulBeforeAddingTrackable[DisconnectAndSuspend]`
* `faulBeforeAddingTrackable[TcpConnectionRefused]`
* `faulBeforeAddingTrackable[TcpConnectionUnresponsive]`
* `faulDuringTracking[TcpConnectionUnresponsive]`

jaley:
Note this should fix #859 too.

KacperKluka:
> The fix should be to **remove the following timeouts**
> - attach
> - connect
> 
> ... and then ensure that nothing else is broken by this change.
> **We must still observe the callback listener** to ensure that we transition to the Failed state on terminal errors, however.

If we do that then we can potentially wait indefinitely for those callback listeners to be called (unless there is a build-in timeout in the `ably-java` implementation), right? Are we okay with this @jaley? Operations affected by those methods are:
- attach
  - `Subscriber.resolutionPreference()` which is used by subscribers to request different resolutions from a publisher
  - `sendMessage()` which is used to send enhanced and raw location updates from the publisher SDK
- connect
  - `SubscriberBuilder.start()` - used by the `DefaultPublisher` when starting and connecting to Ably
  - `Publisher.add()` / `Publisher.track()` - used to start Ably connection when adding the first trackable

Regarding attach:
I've noticed that we are not triggering the `channel.attach()` but simply waiting for the channel to autonomously switch to the `attached` state. Is this correct or should we manually call `attach()` as well?

deanna-lad:
[...]

PR #965, opened 2023-01-30
Remove timeouts from attach and connect

KacperKluka:
Removed the `withTimeout` calls related to `attach` (actually it was waiting on an automatic re-attach) and `connect` operations. We should rely on internal Ably mechanisms instead of introducing artificial timeouts.

The details about the changes can be found [here](https://github.com/ably/ably-asset-tracking-android/issues/948#issuecomment-1406670324).


davyskiba (dismissed):

jaley:
Hmmm... So, I've tried testing this branch locally with the fixes on #938 merged too, and `faultBeforeAddingTrackable[TcpConnectionRefused]` hangs indefinitely. I think this is also what's happening on CI. 

I have a feeling the changes now on #938 actually rely on these timeouts for `connect()` and `attach()` to ensure that the initial call to `track()` does actually return to the caller with the `Offline` status if connectivity is down. Therefore, if we want to remove these timeouts completely, which I think would be ideal, we also need #966 ?

I could be wrong, but that's what it looks like may be happening to me. Maybe @AndyTWF and @davyskiba have some ideas?

KacperKluka:
While this PR removes the timeouts it introduces a potential bug at the same time. The problem is that the `ably.startConnection()` can now never complete and suspend forever if the connection state does not enter either `connected` or `failed` state.

lawrence-forooghian:
As discussed in standup, I’ve merged @AndyTWF’s #970 into this branch. I have the following `NetworkConnectivityTests` issues:

- `faultBeforeAddingTrackable[TcpConnectionRefused]` – still hanging indefinitely (after 3 mins all it's doing is printing the occasional `onWebsocketPing()` to the logs); this is the hang that [@jaley believed would be fixed by #970](https://github.com/ably/ably-asset-tracking-android/pull/965#issuecomment-1409367150) but it doesn't seem to have been
- `faultBeforeAddingTrackable[AttachUnresponsive]` – fails by being timed out by the tests after 2 minutes. This was previously passing before merging #970 here.

AndyTWF:
Could it possible be related to the issue I'm fixing on https://github.com/ably/ably-java/issues/908?

lawrence-forooghian:
`faultBeforeAddingTrackable[AttachUnresponsive]` is now failing because the publisher never enters the `online` state. This is happening because the `online` state [requires that the publisher have sent at least one location](https://github.com/ably/ably-asset-tracking-android/blob/2c4c4dc40f2b8085d0c0f9bca03194cd6edb245a/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt#L466).

We no longer succeed in sending a location in this test because of the following two changes in this PR:

1. We no longer check whether a channel is `SUSPENDED` before attempting to publish to it (this check was previously made by `DefaultAbly.sendRawLocation`).
   
   (This seems related to #973, which @davyskiba is working on.)
2. The [error thrown by ably-java when we try to publish to a `SUSPENDED` channel](https://github.com/ably/ably-java/blob/e57183f8801615a2a9ca916f745928f49f192cba/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L1016) is now considered by `SendEnhancedLocationFailureWorker` to be a fatal error and hence we no longer retry location publishes that have failed due to the channel being `SUSPENDED`.

   (The change suggested by #972 would fix this, if we considered the scope of that issue to include publish attempts and not just presence enter, which [Paddy’s comment](https://github.com/ably/ably-asset-tracking-android/issues/972#issuecomment-1410372130) suggests we should.)

So, what happens now in the test is that even though the channel is `SUSPENDED` (due to the active fault), we still attempt to publish a location update (due to point 1 above) and is subsequently not retried (due to point 2 above).

When exploring solutions for this, I tried the following (ignoring the possible issues this would introduce):

- considering the `SUSPENDED` channel publish error to be non-fatal (i.e. reverting change 2 above) — this caused us to retry the publish, but, since the channel was still suspended, that second publish still failed. And then, since we’d hit the [publish retry count limit](https://github.com/ably/ably-asset-tracking-android/blob/2c4c4dc40f2b8085d0c0f9bca03194cd6edb245a/publishing-sdk/src/main/java/com/ably/tracking/publisher/LocationsPublishingState.kt#L7-L10) (that is, one retry) we did not try to publish again, so same end result. That is, reverting point 2 alone is not enough.
- removing the publish retry limit – this made the test pass since it now kept retrying the location publish until the channel became `ATTACHED` again, at which point the publish succeeded.

I think that to fix this properly we need to do one of the following:

1. Re-introduce a check for whether the channel is `SUSPENDED` before attempting to publish to it (possibly @davyskiba is doing this in #973?)
2. Do not introduce a check for whether the channel is `SUSPENDED` before attempting to publish to it, but make sure that errors resulting from a publish attempt on a `SUSPENDED` channel do not have a retry limit applied to them – I don't think we can do this given that the [error](https://github.com/ably/ably-java/blob/e57183f8801615a2a9ca916f745928f49f192cba/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L1016) makes no distinction between `SUSPENDED` and `FAILED` channels.
3. Completely remove the retry limit for publishes – I assume there’s a good reason this retry limit exists and so it shouldn’t be removed?

So it seems like really we need to do 1.

lawrence-forooghian:
> While this PR removes the timeouts it introduces a potential bug at the same time. The problem is that the `ably.startConnection()` can now never complete and suspend forever if the connection state does not enter either `connected` or `failed` state. ([link to an internal conversation](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1675168574559329?thread_ts=1675152822.165529&cid=C01EPJENRD0))

I've addressed @KacperKluka’s concern by handling the `CLOSED` state in 05f237494958b54b6b56fdd16cb1fcc7265ca3cc. I don't think there are any other states that we need to explicitly handle, since all states other than `FAILED` or `CLOSED` indicate that the realtime instance is still trying to progress towards a `CONNECTED` state.

jaley (approved):
I've looked over the production code changes and `NetworkConnectivityTest`, and that looks all good to me.

AndyTWF:
The failing test is fixed in https://github.com/ably/ably-asset-tracking-android/pull/981, so landing this into the integration branch now and we'll resolve any issues there.

PR #910, opened 2023-01-15
869 continue writing default ably tests

lawrence-forooghian:

PR #977, opened 2023-02-01
Make some fixes and improvements to recent `DefaultAbly` test changes

lawrence-forooghian:
There were some changes to the `DefaultAbly` tests whilst I was away. I took a look at them and there were some things to fix (some errors that made the human-readable descriptions incorrect, some stylistic things). This fixes them. More details in commit messages.


davyskiba (approved):

jaley (commented):
Hey @lawrence-forooghian,

Thanks for making these improvements. I just wanted to get your thoughts on some more potential changes, which you could either do in this PR or another ticket at a later date, whichever you think is right.

In [this change](https://github.com/ably/ably-asset-tracking-android/pull/952), I found myself basically humouring the mocks in this test. We had a bug whereby we weren't checking if the Ably connection is in the `initializing` state, where it won't give you any callbacks, and so we were waiting on a callback listener that would never happen, causing `stop()` to hang indefinitely.

The fix was simple enough - there was already a case to handle other states where callbacks don't happen, but not this one. Adding another call to retrieve the current channel state to that conditional, however, caused about 7 tests to fail here, because they bake in expectations around how many times we should be reading the channel state.

To me, that check doesn't really add any value. It's a stateless property accessor and kind of incidental the important side-effects we need to verify from this class. I'd be happy with just removing all of those checks about reading channel state, as I think what matters is what `DefaultAbly` actually does with that information.

Performing these checks really just makes it impossible to change the code at all without adding much information as to whether the change is correct and makes this test very brittle. In fact, in this case, we were enforcing the wrong behaviour -- *not* checking for the `initialized` state.

I'd encourage focusing on the important side-effects expected of `DefaultAbly` - such as whether we attempt to publish messages as a consequence of `sendEnhancedLocation()`, or whether we schedule the appropriate retries in the `disconnected` or `suspended` states. That's likely too much change for this PR, but perhaps loosening off the checks on stateless property accessors will save others some time in the short term?

lawrence-forooghian:
Thanks for your thoughts @jaley! @davyskiba actually brought up this issue before, in [this internal Slack conversation](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1673867947399909).

Summarising my response, I thought it valuable to know which states a given method is checking, since that tells us which inputs we need to vary when writing test cases. (Bear in mind that at the time I was coming at this from the point of view of writing black box tests with no knowledge of the implementation, using the tests as a way of exploring the behaviour of the class.)

I think that perhaps as part of the remaining work on #910, specifically "augment the black-box tests with any further information found by looking at the code of `DefaultAbly`" we could get rid of this exploratory code.

I also wonder whether this problem will go away by itself anyway; it seems like over time we’re removing a lot of logic from `DefaultAbly` and spreading it out to the workers. For example https://github.com/ably/ably-asset-tracking-android/pull/965/files seems to remove a bunch of channel state checking logic.

PR #990, opened 2023-02-03
Fix issues re-adding trackable

### LINKED ISSUE #984: Re-adding trackable immediately after removing it causes NullPointerException
When RemoveTrackableWorker is called, the trackables stateflow is removed, but the trackable itself isn't removed until DisconnectSuccessWorker - which happens after various async work. Control returns to the caller of publisher.remove() during RemoveTrackableWorker.

In the unlikely event that the trackable call is re-added after RemoveTrackableWorker but before the rest, a NPE is thrown in AddTracakbleWorker because the trackable state flow no longer exists.

### LINKED ISSUE #988: Re-added trackable does not enter or subscribe to presenc
If a trackable is removed from the publisher and then re-added, that trackable will never enter presence or subscribe.

This is because DisconnectSuccessWorker does not remove the trackable from the trackable removal guard, and thus it is permanently "marked for removal". The EnterPresenceWorker and SubscribeToPresenceWorker check if a trackable is being removed and will not run if that is the case.

### LINKED ISSUE #989: Re-adding a trackable results in trackable being offline
Due to the issue in https://github.com/ably/ably-java/issues/913

If you use the publishing example apps to add two trackables, remove one and then re-add it, that trackable will enter a permanently offline state.

Fixing the bug in ably-java would resolve this issue.

AndyTWF:
This change:

- Fixes an issue where re-adding a removed trackable would prevent that trackable from entering presence. This is because DisconnectSuccessWorker does not remove trackables from the removal guard, meaning that once a trackable is removed the first time it is permanently marked for removal. This change makes sure that this un-marking occurs.
- Fixes an issue where a NullPointerException is thrown if a trackable is re-added very quickly after being removed. This is caused by the fact that RemoveTrackableWorker clears the trackables StateFlow but doesn't remove the trackable itself until a later worker (DisconnectSuccessWorker). Therefore, if AddTracakbleWorker is called in the meantime, that will see the trackable is around, and throw an NPE when it tries to access the non-existent StateFlow. This change fixes the issue by adding a check to the trackable removal guard when doing AddTrackableWorker, and delaying the work for a bit if a trackable is being removed.
- Introduces a new version of ably-java that fixes an issue where channels can be brought back spuriously after being released, meaning that trackables cannot be re-added.
 
Fixes #988  
Fixes #984 
Fixes #989 

To test the ably-java fixes, see this draft PR: https://github.com/ably/ably-asset-tracking-android/pull/993

- Go to the publishing example app
- Add two trackables
- Remove one trackable
- Re-add the trackable with the same ID
- The trackable should come online almost immediately

### DIFF
--- a/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/DisconnectSuccessWorker.kt
+++ b/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/DisconnectSuccessWorker.kt
@@ -1,8 +1,7 @@
 package com.ably.tracking.publisher.workerqueue.workers
 
 import com.ably.tracking.common.Ably
-import com.ably.tracking.common.ResultCallbackFunction
-import com.ably.tracking.common.workerqueue.CallbackWorker
+import com.ably.tracking.common.workerqueue.Worker
 import com.ably.tracking.publisher.PublisherInteractor
 import com.ably.tracking.publisher.PublisherProperties
 import com.ably.tracking.publisher.PublisherState
@@ -11,11 +10,10 @@ import com.ably.tracking.publisher.workerqueue.WorkerSpecification
 
 internal class DisconnectSuccessWorker(
     private val trackable: Trackable,
-    callbackFunction: ResultCallbackFunction<Unit>,
     private val publisherInteractor: PublisherInteractor,
     private val shouldRecalculateResolutionCallback: () -> Unit,
     private val ably: Ably,
-) : CallbackWorker<PublisherProperties, WorkerSpecification>(callbackFunction) {
+) : Worker<PublisherProperties, WorkerSpecification> {
     /**
      * Whether the worker is also performing disconnecting.
      * Used to properly handle unexpected exceptions in [onUnexpectedAsyncError].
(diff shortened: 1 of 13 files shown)


AndyTWF:
@lawrence-forooghian @QuintinWillison @davyskiba @ikbalkaya The fixes over at https://github.com/ably/ably-java/pull/914 still require a review before they can be brought into this branch, but I thought it would be worth getting some comments on this whilst those are also being reviewed :) I'll add the new ably-java version later once it's released :)

publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/DisconnectSuccessWorker.kt
```diff
@@ -117,4 +115,12 @@ internal class DisconnectSuccessWorker(
             postWork(WorkerSpecification.StoppingConnectionFinished)
         }
     }
+
+    override fun doWhenStopped(exception: Exception) {
+        // No op
+    }
+
+    override fun onUnexpectedError(exception: Exception, postWork: (WorkerSpecification) -> Unit) {
+        // No op
```
davyskiba, line 124: Maybe we should remove this trackable from `trackableRemovalGuard` anyway?

AndyTWF: Yes - we probably should, done!

davyskiba (approved):
LGTM

lawrence-forooghian (approved):

PR #994, opened 2023-02-07
Test exceptions

### LINKED ISSUE #909: Exceptions thrown inside a block passed to `DefaultAbly`’s `scope.launch()` are suppressed
See 4430c3527b70fa8649cba99501dde2cb84330e13 which demonstrates this issue. The part that worries me is in bold.

> I have modified the code so that the test "connect - when channel attach does not complete" will cause line 374 of Ably.kt to throw an exception.
> 
> When running this test, a message is logged (presumably by the coroutines runtime) about the caught exception, but it does not seem to propagate any further and the test passes.
> 
> This issue only occurs when I throw the exception inside the `scope.launch` block - if I call it outside that block then the exception propagates and the test fails.
> 
> I guess this is related to the fact that `scope` uses `SupervisorJob`, which I believe introduces some sort of error suppression behaviour. I don’t know whether this is intentional, perhaps so that unexpected errors don’t bring the whole process down? **But in unit tests, it has the nasty side effect of causing MockK’s fail-fast mocks – which rely on exceptions to communicate the fact that they have been incorrectly configured (i.e. that a called method has not been mocked) – to not cause a test failure when incorrectly configured. This can lead to incorrectly passing tests.**

davyskiba:
Test exceptions are exposed in the test by passing the coroutine scope to `DefaultAbly` via the constructor. Due to [runTest](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-test.html) limitations, exceptions from `launch` blocks are thrown at the end of the test. I tried to introduce `runTest` to `NetworkConnectivityTests`, but it broke the tests, so I left preparatory cleanup only.

Resolves #909

### DIFF
--- a/common/src/main/java/com/ably/tracking/common/Ably.kt
+++ b/common/src/main/java/com/ably/tracking/common/Ably.kt
@@ -30,18 +30,16 @@ import io.ably.lib.types.ErrorInfo
 import io.ably.lib.types.Message
 import io.ably.lib.types.Param
 import io.ably.lib.util.Log
-import kotlin.coroutines.resume
-import kotlin.coroutines.resumeWithException
-import kotlin.coroutines.suspendCoroutine
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.TimeoutCancellationException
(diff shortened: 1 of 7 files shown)


AndyTWF (approved):
LGTM - just some linting issues :)

PR #996, opened 2023-02-07
Add migration guides for 1.6.0 and 1.6.1

### LINKED ISSUE #974: Improve the version `1.5.1` to `1.6.0` migration guide
Per @KacperKluka's https://github.com/ably/ably-asset-tracking-android/pull/971#discussion_r1091750377.

lawrence-forooghian:
I’ve written these by looking at the corresponding `CHANGELOG` entries.

Closes #974.

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -1,6 +1,24 @@
 # Upgrade / Migration Guide
 
+## Version 1.6.0 to 1.6.1
+
+### `Publisher`’s `add()` and `track()` methods no longer throw `ConnectionException`
+
+These methods now use the state emitted by the returned `StateFlow<TrackableState>` (namely the `Offline` and `Failed` states) to communicate any errors that occur when connecting to Ably.
+
+The documentation for these methods still states that they can throw `ConnectionException` “when something goes wrong with the Ably connection”, but this is no longer true and this exception is no longer thrown. We will correct the documentation in the next release (see https://github.com/ably/ably-asset-tracking-android/issues/995).
+
+We have also introduced retry behaviours that mean that non-fatal errors upon adding a trackable are now handled by the Publisher SDK and will not cause the trackable to enter the `Failed` state.
+
 ## Version 1.5.1 to 1.6.0
+
+### `Publisher.remove()` no longer throws `ConnectionException`
+
+This operation will no longer fail.
+
+### `Publisher` and `Subscriber`’s `stop()` methods no longer throw `ConnectionException`
+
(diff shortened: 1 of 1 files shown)


UPDATING.md
```diff
@@ -1,7 +1,25 @@
...
+
+The documentation for these methods still states that they can throw `ConnectionException` “when something goes wrong with the Ably connection”, but this is no longer true and this exception is no longer thrown. We will correct the documentation in the next release (see https://github.com/ably/ably-asset-tracking-android/issues/995).
+
+We have also introduced retry behaviours that mean that non-fatal errors upon adding a trackable are now handled by the Publisher SDK and will not cause the trackable to enter the `Failed` state.
+
 ## Version 1.5.1 to 1.6.0
 
+### `Publisher.remove()` no longer throws `ConnectionException`
+
+This operation will no longer fail.
```
AndyTWF, line 17: Maybe worth adding that some of the operations (disconnecting from Ably, removing stateflows etc) will happen in the background, but this method will return immediately?

UPDATING.md
```diff
@@ -1,7 +1,25 @@
 # Upgrade / Migration Guide
 
+## Version 1.6.0 to 1.6.1
+
+### `Publisher`’s `add()` and `track()` methods no longer throw `ConnectionException`
+
+These methods now use the state emitted by the returned `StateFlow<TrackableState>` (namely the `Offline` and `Failed` states) to communicate any errors that occur when connecting to Ably.
```
AndyTWF, line 7: Worth adding that these will return immediately with no delay?

davyskiba: There still is a chance of some delay due to queue processing, but it shouldn't be significant

davyskiba, line 7: Is it worth adding that the trackable will most likely be in an `Offline` state when these methods return? 
I don't know if DH is relying on that 🤔

UPDATING.md
lawrence-forooghian: Added a "`Publisher.remove()` returns more quickly" section.

UPDATING.md
lawrence-forooghian: Added a "`Publisher`’s `add()` and `track()` methods return more quickly" section.

lawrence-forooghian: Mentioned this in the new "`Publisher`’s `add()` and `track()` methods return more quickly" section.

davyskiba (approved):

PR #1011, opened 2023-02-20
Unify invalid mapbox values handling

### LINKED ISSUE #1004: Handling invalid values from mapbox
Related to https://github.com/ably/ably-asset-tracking-swift/pull/521

We want to ensure that "invalid" values in location updates are handled consistently across the Android and Swift SDKs. In a previous discussion in January ([internal link](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1672839686628759)), it was agreed that `-1` would be used.

davyskiba:
Unified handling of invalid values coming in MapBox locations with the iOS approach, using -1 as a missing value. 
I've used this opportunity to clean up `MapboxTest` and introduced a `TimeProvider` interface to allow time control in tests.

Resolves #1004


common/src/main/java/com/ably/tracking/common/LocationMappers.kt
```diff
@@ -2,5 +2,34 @@ package com.ably.tracking.common
...
 
-fun android.location.Location.toAssetTracking(timestamp: Long = time): Result<Location> =
-    Location(latitude, longitude, altitude, accuracy, bearing, speed, timestamp).sanitize().validate()
+fun android.location.Location.toAssetTracking(timestamp: Long = time): Result<Location> {
+    // According to Location class documentation, i.e. https://developer.android.com/reference/android/location/Location#getAltitude(),
+    // those fields have a real value only if “has function” returns true. Those checks prevent passing bogus data.
+    val altitude = if (hasAltitude()) {
+        altitude
+    } else {
+        Double.NaN
```
AndyTWF, line 11: Should this be `Location.INVALID_VALUE` also? The docstring on the `Location` class suggests that it should

JakubJankowski: I'm pretty sure that the conclusion from our previous discussions was that only accuracy, bearing and speed should be repairable. That's also how it's implemented Swift AAT. So I'd say that the docstring should be updated.

JakubJankowski (commented):
Not a Kotlin expert, so I won't throw in an approval, but I've looked through the code and it LGTM (apart from the wrong docstring for the `Location.altitude` that Andy's found).

common/src/main/java/com/ably/tracking/common/LocationMappers.kt
davyskiba: Great catch 💪 
I think the implementation is correct: if `hasAltitude()` returns false, we use the invalid value `Double.NaN`, and the location gets skipped.
Updated the docstring in 8edd53a84420b4a17eb645b08a925daee5204a2b

AndyTWF (approved):

PR #1032, opened 2023-03-14
Subscriber extended publisher presence

AndyTWF:


common/src/test/java/com/ably/tracking/common/AblyHelpersTest.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.Assert
+import io.ably.lib.types.PresenceMessage as AblyJavaPresenceMessage
+import org.junit.Test
+
+class AblyHelpersTest {
+    @Test
+    fun itConvertsPresenceMessagesToTracking()
+    {
+        val gson = Gson();
+        val incomingPresenceMessage = "{\"id\": \"6lxmVGvq-4:1:0\", \"clientId\": \"AatNetworkConnectivityTests_Subscriber\", \"connectionId\": \"6lxmVGvq-4\", \"timestamp\": 1678439051717, \"data\": \"{\\\"resolution\\\":{\\\"accuracy\\\":\\\"BALANCED\\\",\\\"desiredInterval\\\":1,\\\"minimumDisplacement\\\":0.0},\\\"type\\\":\\\"SUBSCRIBER\\\"}\", \"action\": 4 }"
```
davyskiba, line 16: With `trimIndent` like [here](https://github.com/ably/ably-asset-tracking-android/blob/8eda44cc2613f969180bc412138066bda8aafe3d/common/src/test/java/com/ably/tracking/common/MessageValidationTests.kt#L165) this json could be more readable

common/src/test/java/com/ably/tracking/common/AblyHelpersTest.kt
```diff
@@ -0,0 +1,27 @@
...
+        val ablyJavaMessage = AblyJavaPresenceMessage.fromEncoded(incomingPresenceMessage, ChannelOptions())
+
+        val trackingMessage: PresenceMessage = ablyJavaMessage.toTracking(gson)!!
+
+        Assert.assertEquals(1678439051717, trackingMessage.timestamp)
```
davyskiba, line 21: It would be great if you could convert those to `Truth` fluent assertions, but that's definitely minor 😉

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Subscriber.kt
```diff
@@ -86,6 +86,13 @@ interface Subscriber {
     val publisherPresence: StateFlow<Boolean>
         @JvmSynthetic get
 
+    /**
+     * The shared flow emitting state changes in publisher presence.
```
lawrence-forooghian, line 90: Now that I see this alongside the `trackableStates` flow, it's a bit inconsistent. There is _always_ a trackable state, but it seems like there is only a publisher presence state after the first change — that is, there doesn't seem to be an initial state.

How do we feel about this inconsistency? If we're okay with it, then we should probably document at which moment subscribers can expect this to emit its first value.

common/src/test/java/com/ably/tracking/common/AblyHelpersTest.kt
AndyTWF: Done in 7188b5a4 :)

common/src/test/java/com/ably/tracking/common/AblyHelpersTest.kt
AndyTWF: Done in 223ab10

core-sdk-java/build.gradle
```diff
@@ -9,6 +9,7 @@ apply from: '../jacoco.gradle'
 
 dependencies {
     api project(':core-sdk')
+    api project(':subscribing-sdk')
```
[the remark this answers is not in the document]
AndyTWF: Have gotten around this by moving the listener in 2f7671e7

All the other listeners live together in the `core` SDK, but seeing as this is a subscriber-only one, I think it makes sense for it to live in the subscriber java SDK

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/Subscriber.kt
lawrence-forooghian: I guess it's maybe a question about the API in that case — why did we decide to go with a flow that emits state _changes_ instead of just emitting _states_? I get that it's what the Ably SDKs do, but it's not very consistent with the rest of the Asset Tracking SDK.

AndyTWF: That's a fair comment - one of the driving factors with the ADR, as you mentioned, was to bring the API of AAT closer to the idioms we have in the other SDKs, which would mean representing stage changes as changes, rather than actual states. Also that in the future, we could add extra functionality around other state flows (e.g. TrackableState) that follow the same pattern.

I guess by doing changes, it can help consumers avoid having to track the current state when a new one comes along, they can simply see "The state went from X to Y".

That said, I can see some potential complications with the implementation of things as state **changes** - namely that StateFlows are perfectly at liberty of dropping intermediate changes, so you could get some weird event flows where the state goes from "Online with metadata X Y Z" to "Went from offline to online", which makes no sense.

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresenceStateChange.kt
```diff
@@ -0,0 +1,64 @@
...
+ * For an individual publisher, this enum represents their last known state.
+ */
+enum class LastKnownPublisherState {
+    PRESENT,
+    ABSENT,
+}
+
+/**
+ * Represents a publisher that AAT knows (via presence history of presence events)
+ * has been present at some point. This class can be used to identify previous publishers
```
lawrence-forooghian, line 29: Should we document the "some point" timeframe here? This might be important for people who make decisions based off whether there is a publisher in `PublisherPresenceStateChange.publisher`.

lawrence-forooghian: Unless you'd prefer to add this in #1036, @davyskiba, once this timeframe is known?

AndyTWF: I think it'd be good to do it once know :)

PR #1034, opened 2023-03-15
Add network connectivity tests assertions on new publisher presence state API

### LINKED ISSUE #1037: Improved Publisher Presence: Network Connectivity Tests
Expand the assertions of the Subscriber network connectivity tests to cover the new StateFlows exposed on the Subscriber as part of the extended Publisher Presence API (as per ADR88).

lawrence-forooghian:
This adds assertions to the subscriber tests for the new `Subscriber.publisherPresenceStateChanges` API added in 603908a.

It only adds assertions for the `current` property of the state change. I thought that adding assertions about the other properties might be quite complicated, and would be easier with unit tests. I’m keen to hear opinions, though.

There are a few things I tidied up along the way — see commit messages for more details.

Resolves #1037.


subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
```diff
@@ -373,7 +373,7 @@ class NetworkConnectivityTests(private val testFault: Fault) {
             trackableId = trackableId,
             faultType = fault.type,
             locationUpdate = thirdLocationUpdate,
-            expectedPublisherPresence = false,
+            publisherDisconnected = true,
```
davyskiba, line 376: I have doubts if this name change is the right direction 🤔 It is more consistent with other parameters but might be a bit misleading. We are asserting publisher presence reported by the subscriber, not the actual publisher state. What do you think?

davyskiba (commented):
Assertions look good 👌

subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt
lawrence-forooghian: I think that the factory method, in this case, is capable of calculating the assertion that needs to be made (i.e. should I expect the subscriber to report the publisher as present?), given the state of the world (i.e. is the fault active, what kind of fault is it, is there _actually_ a publisher present on the channel?). And indeed, that's the way these methods already work in general — they take information about the state of the world (i.e. is the fault active, what kind of fault is it?), and they only take additional instructions about which assertions they need to make when they can't already calculate that from the information they're given about the state of the world.

I'm not sure what you mean about the assumptions comments (e.g. "A publisher is present on the channel if and only if `publisherDisconnected` is false.") — do you think it would help if I made it clearer that these refer to the state of the world, and not the assertions to be made?

davyskiba (approved):

PR #1042, opened 2023-03-22
Publisher presence events

AndyTWF:
Implements the event emission part of the enhanced publisher presence API per ADR88.

- Adds a "newness" check to AATs presence messages to allow for comparison.
- Adds a helper class that keeps track of publisher presence messages and the latest message for each publisher. Can be updated with a set of messages and return those that are "new".
- Adds a helper class that takes presence messages, gets "new" ones and updates an internal publisher presence map. Also keeps a history of known publishers, along with the last times they were seen.


subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresenceMessageProcessor.kt
```diff
@@ -0,0 +1,44 @@
...
+     * Given a list of presence messages, process them and return the last "new" presence message for
+     * any given publisher.
+     */
+    fun processPresenceMessagesAndGetChanges(messages: List<PresenceMessage>): List<PresenceMessage>
+}
+
+class DefaultPublisherPresenceMessageProcessor: PublisherPresenceMessageProcessor {
+
+    // Stores the last applicable presence message associated with this member
+    private val memberMap: MutableMap<String, PresenceMessage> = mutableMapOf()
```
davyskiba, line 20: Could we lift this state to`PublisherPresence` and pass it as a parameter?

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
```diff
@@ -0,0 +1,172 @@
...
+        scope.launch {
+            _stateChanges.emit(
+                PublisherPresenceStateChange(
+              when (hasPresentPublishers()) {
+                        true -> PublisherPresenceState.PRESENT
+                        else -> PublisherPresenceState.ABSENT
+                    },
+                    null,
+                    nowTime,
+                    publisherMapAsList().map {
```
davyskiba, line 100: This map can be simplified to
```
publisherMapAsList().map {
                        it.apply {
                            if(state == LastKnownPublisherState.PRESENT) {
                                lastSeen = nowTime
                            }
                        }
                    }
```

common/src/main/java/com/ably/tracking/common/AblyModels.kt
```diff
@@ -16,7 +16,46 @@ data class PresenceMessage(
...
+            return false
+        }
+
+        val thisMessageSerial = thisIdSplit[1].toIntOrNull()
+        val thisIndex = thisIdSplit[2].toIntOrNull()
+        val compareMessageSerial = compareIdSplit[1].toIntOrNull()
+        val compareMessageIndex = compareIdSplit[2].toIntOrNull()
+
+
+        return thisMessageSerial != null && thisIndex != null && compareMessageSerial != null && compareMessageIndex != null &&
```
davyskiba, line 51: Would it make sense to return `null` if any of compared fields are missing?

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/DefaultPublisherPresenceMessageProcessorTest.kt
```diff
@@ -0,0 +1,277 @@
...
+            data = mockPresenceData,
+            timestamp = 123,
+            memberKey = "abc:def",
+            connectionId = "abc",
+            clientId = "def",
+            id = "abc:0:2"
+        )
+
+        // Not as "new" as 1a, will not be seen in output of first call
+        val message1b = PresenceMessage(
```
davyskiba, line 205: Most of the data in those messages is irrelevant to the test, we could hide it in a factory function to expose only those parameters that are important for the test similar to `createPublisherProperties`. Might also apply to `WorkerTestUtils.DefaultPublisherPresenceTest`

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
```diff
@@ -0,0 +1,172 @@
...
+            return
+        }
+
+        // Emit an event
+        // If the individual publishers are present, then their lastSeen time is now.
+        val nowTime = Date().time
+        scope.launch {
+            _stateChanges.emit(
+                PublisherPresenceStateChange(
+              when (hasPresentPublishers()) {
```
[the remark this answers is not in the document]
AndyTWF: 10f4673a

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
```diff
@@ -0,0 +1,172 @@
...
+                PublisherPresenceStateChange(
+                    PublisherPresenceState.UNKNOWN,
+                    ErrorInformation(
+                        code = PublisherStateUnknownReasons.SUBSCRIBER_NOT_ONLINE.value,
+                        statusCode = 0,
+                        message = "Subscriber is not online",
+                        href = null,
+                        cause = null
+                    ),
+                    Date().time, publisherMapAsList())
```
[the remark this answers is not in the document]
AndyTWF: 3fc4eecc

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresenceMessageProcessor.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    override fun processPresenceMessagesAndGetChanges(messages: List<PresenceMessage>): List<PresenceMessage> {
```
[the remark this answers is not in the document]
AndyTWF: Done the second in 4f2481db

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresenceMessageProcessor.kt
AndyTWF: I'm not sure about passing this parameter in - it's key to the internal workings of the class and isn't going to be used outside, so it makes sense to not expose it. The PublisherPresence also takes this class as an interface, so we'd have to elevate the implementation details to the interface, which I don't think is ideal :)

common/src/main/java/com/ably/tracking/common/AblyModels.kt
AndyTWF: We could, but I think its probably best to remain consistent with `ably-java` on this one: https://github.com/ably/ably-java/blob/deec71a40affe522c59160b9278334768b928d90/lib/src/main/java/io/ably/lib/realtime/Presence.java#L1106

subscribing-sdk/src/test/java/com/ably/tracking/subscriber/DefaultPublisherPresenceMessageProcessorTest.kt
AndyTWF: 2ac99dd9

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
```diff
@@ -119,33 +114,32 @@ class DefaultPublisherPresence (
...
+                    cause = null
+                ),
+                now,
+                publisherMapAsList()
             )
-        }
+        )
     }
 
     override fun hasPresentPublishers(): Boolean = publisherMap.asSequence().firstOrNull { it.value.state == LastKnownPublisherState.PRESENT } != null
```
davyskiba, line 140: I think `asSequence` is not doing anything here 🤔

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
```diff
@@ -168,5 +162,22 @@ class DefaultPublisherPresence (
         else -> LastKnownPublisherState.PRESENT
     }
 
-    private fun publisherMapAsList(): List<KnownPublisher> = publisherMap.toList().map{ it.second }
+    private fun publisherMapAsList(): List<KnownPublisher> = publisherMap.toList().map { it.second }
```
davyskiba, line 165: `publisherMap.values.toList()` would be more efficient here

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresenceMessageProcessor.kt
davyskiba: Alright, you've convinced me :)

davyskiba (approved):

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
AndyTWF: `firstOrNull` doesnt exist on the map apparently :(

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
AndyTWF: eed16b24

subscribing-sdk/src/main/java/com/ably/tracking/subscriber/PublisherPresence.kt
davyskiba: Missed that this is a map 🤦  In that case, I think 
`publisherMap.any { it.value.state == LastKnownPublisherState.PRESENT }` is the way to go

lawrence-forooghian (approved):
I've taken an overall look at this but not in detail. I think that if between you and @davyskiba you're happy with this then that's good enough 😅

PR #1052, opened 2025-09-28
Update README with sunset notice for Ably Asset Tracking

jamienewcomb:
We have made the decision to sunset AAT. This change details that in the readme


AndyTWF (approved):
Happy to approve with failed CI as sunsetting.

